PR #5, opened 2020-11-30
OOP Advanced Task

IvanLuchkin:


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,51 @@
...
+        }
+        return result;
+    }
+
+    private static void printFigures(List<Figure> listOfFigures) {
+        for (Figure figure : listOfFigures) {
+            System.out.println(figure.draw());
+        }
+    }
+
```
Sofasmile, line 50: let's remove empty line

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,51 @@
...
+    public static void main(String...args) {
+        printFigures(generateRandomFigureList());
+    }
+
+    private static List<Figure> generateRandomFigureList() {
+        int initialCapacity = (int) ((Math.random() + 3) * 5);
+        List<Figure> result = new ArrayList<>(initialCapacity);
+        int iterator = 0;
+        while (iterator < initialCapacity) {
+            switch ((int) (Math.random() * 5)) {
```
Sofasmile, line 23: declare 5 as a constant, now it's magic number

src/main/java/core/basesyntax/model/IsoscelesTrapezoid.java
```diff
@@ -0,0 +1,67 @@
...
+    private final int smallerBase;
+
+    public IsoscelesTrapezoid(
+            int sideAngle, int sideLength, int smallerBase, int biggerBase, Color color) {
+        this.sideLength = sideLength;
+        this.sideAngle = sideAngle;
+        this.smallerBase = smallerBase;
+        this.biggerBase = biggerBase;
+        this.color = color;
+        this.area = Math.round(((smallerBase + biggerBase) / 2)
```
Sofasmile, line 17: why do we use Math.round() for calculation area?

src/main/java/core/basesyntax/model/Figure.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax.model;
+
+public abstract class Figure {
+
```
Sofasmile, line 4: let's remove one empty line

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+public class Main {
+
+    public static void main(String...args) {
+        printFigures(generateRandomFigureList());
+    }
+
+    private static List<Figure> generateRandomFigureList() {
```
Sofasmile, line 18: let's move this method to the separate class - FigureGenerator

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,51 @@
...
+        int initialCapacity = (int) ((Math.random() + 3) * 5);
+        List<Figure> result = new ArrayList<>(initialCapacity);
+        int iterator = 0;
+        while (iterator < initialCapacity) {
+            switch ((int) (Math.random() * 5)) {
+                case 0 :
+                    result.add(new Circle());
+                    break;
+                case 1 :
+                    result.add(new IsoscelesTrapezoid());
```
Sofasmile, line 28: it is hard to understand, that an empty constructor will be generated figures with different parameters. 
Let's generate all values in this method and pass to the constructor as arguments.

src/main/java/core/basesyntax/model/Color.java
```diff
@@ -0,0 +1,18 @@
...
+import java.util.Random;
+
+public enum Color {
+    BLACK,
+    WHITE,
+    RED,
+    GREEN,
+    BLUE;
+
+    public static Color randomColor() {
```
Sofasmile, line 12: better move this method to FigureGenerator class too

src/main/java/core/basesyntax/model/Figure.java
```diff
@@ -0,0 +1,10 @@
...
+    public double area;
+    public Color color;
+
+    public abstract String draw();
```
Sofasmile, line 8: this method better to move to the interface:
- interface - behaviour
- abstract class - state

Sofasmile (changes requested):

src/main/java/core/basesyntax/Main.java
IvanLuchkin: Won't it reduce the readability? Some classes require a low of paremeters

IvanLuchkin: My solution involves implementing a separate method for each `Figure` generation in `FigureGenerator`

src/main/java/core/basesyntax/model/Figure.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.model;
+
+public abstract class Figure implements Drawable {
+    public static final int NUMBER_OF_FIGURES = 5;
```
Sofasmile, line 4: let's move this constant to the FigureGenerator class)

src/main/java/core/basesyntax/model/Figure.java
```diff
@@ -0,0 +1,7 @@
...
+    public double area;
+    public Color color;
```
Sofasmile, line 6: let's make these variables private)
remember about encapsulation

Sofasmile (changes requested):
good) left some comments

src/main/java/core/basesyntax/model/Figure.java
IvanLuchkin: In my opinion, since this constant represents the total amount of classes, inherited from `Figure`, it would be better to store it here. If we want to change something in random generation logic, this constant has to stay the same anyway

src/main/java/core/basesyntax/model/Figure.java
IvanLuchkin: I aggree about the mistake here, but i can't make them `private` because inherited classes won't be able to have this fields. Is making them `protected` a valid solution in such case?

src/main/java/core/basesyntax/model/Figure.java
Sofasmile: we can make these fields private and create constructor, getters or setters

src/main/java/core/basesyntax/model/Figure.java
Sofasmile: we can make these fields private and create constructor, getters or setters

Sofasmile (changes requested):
added a comment to the previous review

src/main/java/core/basesyntax/model/Figure.java
IvanLuchkin: But the area is calculated differently for each figure, and if I have a milti-level inheritance, I cant use parent constructors in order to initialize area field. I consider a workaround with protected setArea and setColor, but is it valid?

Sofasmile (approved):
great job))

PR #6, opened 2020-12-01
Add solution

Resci:


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,19 @@
...
+    public static void main(String[] args) {
+        Random randomGenerator = new Random();
+        int random = randomGenerator.nextInt(100);
+        Figures[] figures = new Figures[random];
+
+        for (int i = 0; i < random; i++) {
+            figures[i] = Figures.getRandomFigure();
+        }
+
+        for (Figures some: figures) {
```
RomanDubovskyi, line 15:
```suggestion
        for (Figures figure/element: figures) {
```

src/main/java/core/basesyntax/Figures.java
```diff
@@ -0,0 +1,52 @@
...
+
+    public Color getColor() {
+        return color;
+    }
+
+    public double getArea() {
+        return area;
+    }
+
+    public static Figures getRandomFigure() {
```
RomanDubovskyi, line 22: With this logic your abstract class depends on the impl and it's a bad practice, move this logic to separate class, e.g. "FigureGenerator". This way your code will follow single responsibility principle

src/main/java/core/basesyntax/Figures.java
```diff
@@ -0,0 +1,52 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public abstract class Figures {
+    private static Random random = new Random();
+    double area;
+    Color color;
```
RomanDubovskyi, line 8: what about access modifiers, should it be default?

src/main/java/core/basesyntax/Color.java
```diff
@@ -0,0 +1,60 @@
...
+
+    private static Random random = new Random();
+    private String code;
+
+    Color(String code) {
+        this.code = code;
+    }
+
+    public static Color getRandom() {
+        switch (random.nextInt(8)) {
```
RomanDubovskyi, line 23: 8 is a magic number, why not 7 or 9 ?

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/Color.java
Resci: We have 8 different color in enum structure, so i used random.nextInt(8) which return [0; 8). It is better to change it to variable? If do that, may appeaar situation, when someone add the new color but forget add it to random generator

RomanDubovskyi: so instead you can use enum length which will make code more readable and flexible 
https://stackoverflow.com/questions/17492079/how-to-get-number-of-possible-item-of-an-enum/17492102

src/main/java/core/basesyntax/figures/Figures.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax.figures;
+
+public abstract class Figures {
+
+    protected double area;
+    protected Color color;
+
+    public abstract void displayInfo();
+
+    public abstract void draw();
```
RomanDubovskyi, line 10: these methods better to move to interface and make this class implement it (interface - behaviour, abstract class state)

src/main/java/core/basesyntax/figures/FigureGenerator.java
```diff
@@ -0,0 +1,38 @@
+package core.basesyntax.figures;
+
+import java.util.Random;
+
+public class FigureGenerator {
+    private static Random random = new Random();
+
+    public static Figures getRandomFigure() {
+        switch (random.nextInt(5)) {
```
RomanDubovskyi, line 9: declare 5 as a constant, not it's magic number

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/figures/FigureGenerator.java
```diff
@@ -0,0 +1,39 @@
+package core.basesyntax.figures;
+
+import java.util.Random;
+
+public class FigureGenerator {
+    private static final int count = 5; //Number of inherited classes
```
RomanDubovskyi, line 6:
```suggestion
    private static final int FIGURES_AMOUNT / FIGURES_QUANTITY = 5; 
```
make code self description, constant naming - https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names

src/main/java/core/basesyntax/figures/Figures.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax.figures;
+
+public abstract class Figures implements Displayable, Drawable {
+
```
RomanDubovskyi, line 4: this empty line here is redundant

RomanDubovskyi (changes requested):
Good job, almost there 👍

src/main/java/core/basesyntax/figures/Figures.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax.figures;
+
+public abstract class Figures implements Displayable, Drawable {
```
RomanDubovskyi, line 3:
```suggestion
public abstract class Figure implements Displayable, Drawable {
```

src/main/java/core/basesyntax/figures/IsoscelesTrapezoid.java
```diff
@@ -0,0 +1,33 @@
+package core.basesyntax.figures;
+
+public class IsoscelesTrapezoid extends Figures {
```
RomanDubovskyi, line 3: For future - it's better to have getters and setters as well, so you can change figure state after creation,

RomanDubovskyi (approved):

PR #9, opened 2020-12-01
add solution

antoniukivan:


src/main/java/core/basesyntax/Test.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.Random;
+
+/**
+ * Feel free to remove this class and create your own.
+ */
+public class Test {
+    public static void main(String[] args) {
+        Random random = new Random();
+        Shape[] shapes = new Shape[random.nextInt(10)];
+        for (int i = 0; i < shapes.length; i++) {
```
RomanDubovskyi, line 12: move the logic of creating random figure into a separate class, your main method is too large

src/main/java/core/basesyntax/Shape.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax;
+
+public abstract class Shape {
```
RomanDubovskyi, line 3: let's use abstract class to store common state, do we have common fields among our child classes?

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public Circle(int radius, String color) {
+        this.radius = radius;
+        this.color = color;
+    }
+
+    public int getRadius() {
+        return radius;
+    }
+
```
RomanDubovskyi, line 15: better to use @Override annotation to make code more informative

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/Circle.java
antoniukivan: but only circle has radius

antoniukivan: > better to use @override annotation to make code more informative

but only circle has radius

RomanDubovskyi: that was concerning method "draw" you didin't have the annotation in prev commit

src/main/java/core/basesyntax/RightTriangle.java
```diff
@@ -0,0 +1,41 @@
+package core.basesyntax;
+
+public class RightTriangle extends Shape implements Figure {
```
RomanDubovskyi, line 3: you may implement the interface in the Shape class, this way it will implemented in the child classes by default

src/main/java/core/basesyntax/Shape.java
```diff
@@ -0,0 +1,10 @@
...
+    String color;
+    double area;
```
RomanDubovskyi, line 5: what about access modifiers should they be default ? 
what happens if you move your child classes to a different package ?

src/main/java/core/basesyntax/Shape.java
```diff
@@ -0,0 +1,10 @@
...
+
+    abstract double getArea();
```
RomanDubovskyi, line 7: same here

src/main/java/core/basesyntax/Test.java
```diff
@@ -0,0 +1,73 @@
...
+        fillShapes(shapes);
+
+        for (Shape shape : shapes) {
+            System.out.println(shape);
+        }
+    }
+
+    public static void fillShapes(Shape[] shapes) {
+        Random random = new Random();
+        for (int i = 0; i < shapes.length; i++) {
```
RomanDubovskyi, line 21: you didin't complete this comment, this logic should be in the different class

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/Shape.java
```diff
@@ -0,0 +1,12 @@
...
+
+public abstract class Shape {
+    protected Color color;
+    protected double area;
+
+    public abstract double getArea();
+
+    public abstract Color getColor();
+
+    public abstract void draw();
```
RomanDubovskyi, line 11: for common behaviour it's better to use interface (in your case declare methods draw() and getArea() in the interface and then implement it in this class)

src/main/java/core/basesyntax/Shape.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+public abstract class Shape {
+    protected Color color;
+    protected double area;
```
RomanDubovskyi, line 5: we don't use this field at all, let's remove it

src/main/java/core/basesyntax/ShapeProducer.java
```diff
@@ -0,0 +1,34 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ShapeProducer {
+    public static Shape getShape() {
+        Random random = new Random();
+        int variant = random.nextInt(5);
```
RomanDubovskyi, line 8: declare 5 as a constant, now it's magic number

RomanDubovskyi (changes requested):
nice, almost there, left last comments

RomanDubovskyi:
all good, but fix travis, check failed

RomanDubovskyi (changes requested):
travis

RomanDubovskyi (approved):

PR #20, opened 2020-12-02
added solution

ivlytvynets:


src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,19 @@
...
+        this.color = color;
+    }
+
+    public Color getColor() {
+        return color;
+    }
+
+    public abstract void setArea();
+
+    public abstract void getInfo();
```
RomanDubovskyi, line 18: let's move this method to an interface, they are usually used to declare a behavior

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Main {
+    public static void main(String[] args) {
+        FigureProducer figureProducer = new FigureProducer();
+        for (int i = 0; i < new Random().nextInt(1000); i++) {
```
RomanDubovskyi, line 8: we should create an array with figures

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,28 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
```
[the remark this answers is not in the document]
ivlytvynets: I didn't understand. What problem is this?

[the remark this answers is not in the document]
ivlytvynets: I have `protected double area` in class Figure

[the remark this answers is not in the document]
ivlytvynets: fixed

src/main/java/core/basesyntax/Figure.java
ivlytvynets: fixed

src/main/java/core/basesyntax/Main.java
ivlytvynets: fixed

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,28 @@
...
+        }
+        this.radius = radius;
+    }
+
+    public double getRadius() {
+        return radius;
+    }
+
+    @Override
+    public void setArea() {
```
RomanDubovskyi, line 19:
```suggestion
    public void calculateArea() {
```
we use setter when we insert value into field, for example public void setArea(double area)

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,23 @@
...
+        return color;
+    }
+
+    public double getArea() {
+        return area;
+    }
+
+    public abstract void setArea();
+
+    public abstract void getInfo();
```
RomanDubovskyi, line 22: previous comment is still not fixed, getInfo method should be in the interface, same with calculateArea(), let;s remove it from here

src/main/java/core/basesyntax/FigureProducer.java
```diff
@@ -0,0 +1,68 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureProducer {
+
+    private static final int COUNT_OF_FIGURE_IMPLEMENTATIONS = 5;
```
RomanDubovskyi, line 7:
```suggestion
    private static final int COUNT_OF_FIGURE = 5;
```

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/Figure.java
ivlytvynets: fixed

RomanDubovskyi (approved):

PR #26, opened 2020-12-02
Oop hw advanced by Oleh Kravets

Ketch92:


src/main/java/core/basesyntax/figure/figures/Circle.java
```diff
@@ -0,0 +1,45 @@
...
+        return radius;
+    }
+
+    public void setRadius(int radius) {
+        this.radius = radius;
+    }
+
+    @Override
+    public String toString() {
+        return getColor() + " " + NAME
```
RomanDubovskyi, line 41:
```suggestion
        return getColor() + " " + getClass().getSimpleName()
```
there is no need to create enum

src/main/java/core/basesyntax/figure/producers/FigureProducer.java
```diff
@@ -0,0 +1,74 @@
...
+import core.basesyntax.figure.figures.Circle;
+import core.basesyntax.figure.figures.IsoscelesTrapezoid;
+import core.basesyntax.figure.figures.Rectangle;
+import core.basesyntax.figure.figures.RightTriangle;
+import core.basesyntax.figure.figures.Square;
+import java.util.Random;
+
+public class FigureProducer {
+
+    public static Figure getRandomFigure() {
```
RomanDubovskyi, line 13:
```suggestion
    public Figure getRandomFigure() {
```
let's not use static methods here

src/main/java/core/basesyntax/figure/producers/FigureProducer.java
```diff
@@ -0,0 +1,74 @@
...
+            case 4:
+            default:
+                return new IsoscelesTrapezoid(ColorProducer.getRandomColor(),
+                        randomizer.nextInt(bound),
+                        randomizer.nextInt(bound),
+                        randomizer.nextInt(bound));
+        }
+    }
+
+    public static Figure[] getRandomFigures() {
```
RomanDubovskyi, line 41: same with other methods

RomanDubovskyi, line 41: we have duplicated code here, you can just hardcode amount value and call your getRandomFigures(int amount) method from here

src/main/java/core/basesyntax/figure/producers/FigureProducer.java
```diff
@@ -0,0 +1,74 @@
...
+        int index = 0;
+
+        while (index < figures.length) {
+            figures[index++] = getRandomFigure();
+        }
+
+        return figures;
+    }
+
+    public static void print(Figure figure) {
```
RomanDubovskyi, line 65: it's not logical to have print method in FigureProducer class, it has different purpose according to its name

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/figure/figures/Circle.java
Ketch92: I thought that enum isn't fitting there. Ok thanks

src/main/java/core/basesyntax/figure/producers/FigureProducer.java
Ketch92: we shall not use static methods there!

src/main/java/core/basesyntax/figure/producers/FigureProducer.java
Ketch92: Nice tip, thx

src/main/java/core/basesyntax/figure/producers/FigureProducer.java
Ketch92: I though who produces also is able to print, but as you say

src/main/java/core/basesyntax/figure/figures/Circle.java
RomanDubovskyi: we don't need to use enum here getClass().getSimpleName() is just a method

src/main/java/core/basesyntax/figure/producers/FigureProducer.java
RomanDubovskyi: if you have producer class it should create new objcets/values, if you need to print something you can do in the class where you call print, or create some util class. This way you follow Single responsibility principle

src/main/java/core/basesyntax/figure/producers/FigureProducer.java
```diff
@@ -0,0 +1,56 @@
...
+import core.basesyntax.figure.figures.Square;
+import java.util.Random;
+
+public class FigureProducer {
+
+    public Figure getRandomFigure() {
+        Random randomizer = new Random();
+        int bound = 1000;
+
+        switch (new Random().nextInt(5)) {
```
RomanDubovskyi, line 17: if 5 is a number of your figures it's better to declare as a constant 
private static int FIGURE_NUMBER = 5;
this way it won't be a magic number in the code

RomanDubovskyi (approved):
cool, left last comment

src/main/java/core/basesyntax/figure/producers/FigureProducer.java
Ketch92: Ok, got it. Also had something like that on mind, I mean to get read of magic numbers.
Is there a way to find a number of classes who extends that class so the figure number will update it self?

PR #27, opened 2020-12-02
Hw 3 2

IgorBochkor:
added solution


src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import core.basesyntax.constant.Color;
+import core.basesyntax.producers.ColorProducer;
+
+public abstract class Figure implements Drawable {
```
RomanDubovskyi, line 6: you figure should have field - color (it's a state for each object )

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,13 @@
...
+    public abstract double getArea();
```
RomanDubovskyi, line 7: move method to an interface - that's behavior, interfaces are used to represent behavior

src/main/java/core/basesyntax/producers/FigureProducer.java
```diff
@@ -0,0 +1,31 @@
...
+import core.basesyntax.Figure;
+import core.basesyntax.shape.Circle;
+import core.basesyntax.shape.IsoscelesTrapezoid;
+import core.basesyntax.shape.Rectangle;
+import core.basesyntax.shape.RightTriangle;
+import core.basesyntax.shape.Square;
+import java.util.Random;
+
+public class FigureProducer {
+    public static final int COUNT_OF_FIGURE_IMPLEMENTATIONS = 5;
```
RomanDubovskyi, line 12:
```suggestion
    public static final int COUNT_OF_FIGURE = 5;
```
Figure is a class, we don't implement classes only interfaces

src/main/java/core/basesyntax/producers/NumberProducer.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax.producers;
+
+import java.util.Random;
+
+public class NumberProducer {
+    public static int getNumber() {
+        return new Random().nextInt(15) + 1;
+    }
+
+    public static int getNumberListFigure() {
```
RomanDubovskyi, line 10:
```suggestion
    public static int getFigureListIndex() {
```

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/Figure.java
IgorBochkor: done

src/main/java/core/basesyntax/Figure.java
IgorBochkor: done!

src/main/java/core/basesyntax/producers/FigureProducer.java
IgorBochkor: done

src/main/java/core/basesyntax/producers/NumberProducer.java
IgorBochkor: done

RomanDubovskyi (approved):

PR #29, opened 2020-12-02
created all classes

YuraBondarenko:


src/main/java/core/basesyntax/Drawable.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Drawable {
+    void drawingFigures();
```
RomanDubovskyi, line 4:
```suggestion
    void drawFigure();
```

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum Colors {
```
RomanDubovskyi, line 3:
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/FigureProducer.java
```diff
@@ -0,0 +1,30 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureProducer {
+    public static final int figure = 5;
```
RomanDubovskyi, line 6:
```suggestion
    public static final int FIGURE_COUNT = 5;
```
we declare constant in upper case
do we use this constant only in this class, if yes let's make private

src/main/java/core/basesyntax/FigureProducer.java
```diff
@@ -0,0 +1,30 @@
...
+
+    public static Figure getFigure() {
```
RomanDubovskyi, line 8: why methods are static in this class?

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/FigureProducer.java
YuraBondarenko: to access this method in the class

RomanDubovskyi: you can access not only static methods 
FigureProducer figureProducer = new FigureProducer();
figureProducer.getFigure() 
will work in main mehtod

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
```
RomanDubovskyi, line 3: add getRadius() method, so we can get this value from object

src/main/java/core/basesyntax/FigureProducer.java
```diff
@@ -0,0 +1,30 @@
...
+                return new IsoscelesTrapezoid(getColor(), getRandomNumber(),
+                        getRandomNumber(), getRandomNumber());
+        }
+    }
+
+    public static int getRandomNumber() {
+        return new Random().nextInt(10) + 1;
+    }
+
+    public static Color getColor() {
```
RomanDubovskyi, line 26: if we use this method only inside of this class make it private 
```suggestion
    public Color getColor() {
```

src/main/java/core/basesyntax/FigureProducer.java
```diff
@@ -0,0 +1,30 @@
...
+            case 3: return new Square(getColor(), getRandomNumber());
+            case 4: return new RightTriangle(getColor(), getRandomNumber(), getRandomNumber());
+            case 5:
+            default:
+                return new IsoscelesTrapezoid(getColor(), getRandomNumber(),
+                        getRandomNumber(), getRandomNumber());
+        }
+    }
+
+    public static int getRandomNumber() {
```
RomanDubovskyi, line 22: same here

src/main/java/core/basesyntax/FigureProducer.java
```diff
@@ -0,0 +1,30 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureProducer {
+    private static final int FIGURE_COUNT = 5;
+
+    public static Figure getFigure() {
```
RomanDubovskyi, line 8:
```suggestion
    public Figure getFigure() {
```

RomanDubovskyi (changes requested):

RomanDubovskyi (approved):

PR #32, opened 2020-12-02
HW-3 

Baldinini:


src/main/java/core/basesyntax/CountFigure.java
```diff
@@ -0,0 +1,29 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class CountFigure {
+    public static final int CountFigure = 5;
```
RomanDubovskyi, line 6:
```suggestion
    public static final int COUNT_FIGURE = 5;
```
we declare constants with upper case

src/main/java/core/basesyntax/CountFigure.java
```diff
@@ -0,0 +1,29 @@
...
+                return new Rectangle(random(), random(), colorProducer.getColor());
+            case 3:
+                return new Circle(random(), colorProducer.getColor());
+            default:
+                return new IsoscelesTrapezoid(random(),
+                        random(), random(), colorProducer.getColor());
+        }
+    }
+
+    public int random() {
```
RomanDubovskyi, line 26: we usually use verbs for method names. And if we use this method only in this class it's better to declare it as a private

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,15 @@
...
+
+    public Figure(Color color) {
+        this.color = color;
+    }
+
+    public Color getColor() {
+        return color;
+    }
+
+    public abstract double getArea();
```
RomanDubovskyi, line 14: let's move this method to the interface and change interface name (interfaces are commonly used to represent behavior)

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Main {
+    public static void main(String[] args) {
+        Figure[] figures = new Figure[new Random().nextInt(20)];
+        for (Figure i : figures) {
```
RomanDubovskyi, line 8: "i" is not an informative naming

src/main/java/core/basesyntax/CountFigure.java
```diff
@@ -0,0 +1,29 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class CountFigure {
```
RomanDubovskyi, line 5: do we count figure here or create them?

RomanDubovskyi (changes requested):

RomanDubovskyi (approved):

PR #34, opened 2020-12-02
Created required classes,methods,interfaces

AndriiBardin:


src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,21 @@
+package core.basesyntax;
+
+public class Circle extends Figure implements Drawable {
```
RomanDubovskyi, line 3: if every class that extends Figure implements Drawable, can we implement Drawable from Figure?

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,21 @@
...
+    private final double radius;
+
+    public Circle(double radius,Color color) {
```
RomanDubovskyi, line 6:
```suggestion
    public Circle(double radius, Color color) {
```

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,21 @@
...
+        this.radius = radius;
+        this.color = color;
+        this.area = Math.pow(radius,2) * Math.PI;
```
RomanDubovskyi, line 9: we won't be able to get access to area from a class in a different package, add method getArea()

src/main/java/core/basesyntax/RandomGenerator.java
```diff
@@ -0,0 +1,43 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class RandomGenerator {
+    private static final int TOTALNUMBER = 5;
```
RomanDubovskyi, line 6:
```suggestion
    private static final int TOTAL_NUMBER = 5;
```
let's improve naming, it's not clear total number of what?

src/main/java/core/basesyntax/RandomGenerator.java
```diff
@@ -0,0 +1,43 @@
...
+        }
+    }
+
+    public static int getRandomLength() {
+        return new Random().nextInt(50);
+    }
+
+    public static Color getColor() {
+        Color[] hues = Color.values();
+        return hues[random.nextInt(hues.length)];
```
RomanDubovskyi, line 41:
```suggestion
        return random.nextInt(Color.values().length);
```

src/main/java/core/basesyntax/RandomGenerator.java
```diff
@@ -0,0 +1,43 @@
...
+            default:
+                return new Circle(0,Color.AMETHYST);
+        }
+    }
+
+    public static int getRandomLength() {
+        return new Random().nextInt(50);
+    }
+
+    public static Color getColor() {
```
RomanDubovskyi, line 39: if we use this method only inside of the class let's make it private

src/main/java/core/basesyntax/Trapezoid.java
```diff
@@ -0,0 +1,26 @@
+package core.basesyntax;
+
+public class Trapezoid extends Figure implements Drawable {
+
+    private double baseA;
```
RomanDubovskyi, line 5: why in some classes your figure parameters are final and in some not. Let's keep code consistent and use same approach everywhere unless there is a particular reason not to do it

RomanDubovskyi (changes requested):

RomanDubovskyi (approved):

PR #35, opened 2020-12-02
done hw

Alex-Korzun:


src/main/java/core/basesyntax/FigureProducer.java
```diff
@@ -0,0 +1,30 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureProducer {
+
+    public static final int COUNT_OF_FIGURE_IMPLEMENTATIONS = 5;
```
RomanDubovskyi, line 7:
```suggestion
    public static final int COUNT_OF_FIGURES = 5;
```
Figure does not have implamentation, cause it's a class

src/main/java/core/basesyntax/Area.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Area {
+    public double calculateArea();
```
RomanDubovskyi, line 4:
```suggestion
    double calculateArea();
```
methods in an interface are public abstract by default

src/main/java/core/basesyntax/Draw.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Draw {
+    public void draw();
```
RomanDubovskyi, line 4: same here

RomanDubovskyi (changes requested):
good job left minor comments

src/main/java/core/basesyntax/Draw.java
Alex-Korzun: done

src/main/java/core/basesyntax/FigureProducer.java
```diff
@@ -0,0 +1,30 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureProducer {
+
+    public static final int COUNT_OF_FIGURE = 5;
```
RomanDubovskyi, line 7:
```suggestion
    private static final int COUNT_OF_FIGURE = 5;
```
if we use some element only once in the class we should keep it as strict as possible

RomanDubovskyi (approved):

PR #38, opened 2020-12-02
added solution

oroboro-cpu:


src/main/java/core/basesyntax/Figures.java
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    public String getColor() {
+        return color.toString();
+    }
+
+    public abstract double getArea();
+
+    @Override
+    public void draw() {
```
RomanDubovskyi, line 17: remove this method, we are not obligated to override interfaces method in the abstract class if we gonna do it in child classes

src/main/java/core/basesyntax/Figures.java
```diff
@@ -0,0 +1,26 @@
...
+
+    public Figures(Color color) {
+        this.color = color;
+    }
+
+    public String getColor() {
+        return color.toString();
+    }
+
+    public abstract double getArea();
```
RomanDubovskyi, line 14: create a separate interface and move this method there

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/Drawable.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Drawable {
+    public void draw();
```
RomanDubovskyi, line 4: methods in interface are public abstract by default

src/main/java/core/basesyntax/Figures.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public abstract class Figures implements Drawable, Displayable, Calculable {
```
RomanDubovskyi, line 3:
```suggestion
public abstract class Figure implements Drawable, Displayable, Calculable {
```

RomanDubovskyi (approved):

src/main/java/core/basesyntax/Drawable.java
oroboro-cpu: All corrections are done. Thank you.

PR #43, opened 2020-12-02
Issue2

Alekearl:


src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,36 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Circle extends Figure implements Area, ColorCreator {
```
RomanDubovskyi, line 5: make your Figure implement Area and you won't need to do it in each class

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,36 @@
...
+        this.radius = radius;
+    }
+
+    @Override
+    public double countArea() {
+        return Math.PI * (radius * radius);
+    }
+
+    @Override
+    public Color getColor() {
```
RomanDubovskyi, line 26: this method should return value of color field and not generte random color each

src/main/java/core/basesyntax/ColorCreator.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface ColorCreator {
+    Color getColor();
```
RomanDubovskyi, line 4: make it as a simple getter inside of abstract class

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public abstract class Figure {
+    public abstract String onScreen();
```
RomanDubovskyi, line 4: this class should have common state (fields), for example you can store color here

RomanDubovskyi, line 4: change method name, it's not clear what it does and it should be a verb. Plus, move it to a separate interface

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/Circle.java
Alekearl: Done. =)

src/main/java/core/basesyntax/Circle.java
Alekearl: Changed

src/main/java/core/basesyntax/ColorCreator.java
Alekearl: Changed

src/main/java/core/basesyntax/Figure.java
Alekearl: Changed

Alekearl: Changed

RomanDubovskyi (approved):

PR #45, opened 2020-12-03
hw3-draw figures

tbondare:


src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,29 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Circle extends Figure {
+    private int radius = 0;
```
RomanDubovskyi, line 6: init this value inside of the constructor, but don't generate a random value here, it should be passed from the outside to constructor

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    @Override
+    public String getName() {
+        return "circle";
+    }
+
+    @Override
+    public double getArea() {
+        setData();
```
RomanDubovskyi, line 21: we can remove setDate() after you create a constructor with radius initialization

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,29 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public abstract class Figure implements FigureBehaviour {
+
+    public static Figure getFigure() {
+        int randomFigure = new Random().nextInt(5);
```
RomanDubovskyi, line 8: 5 is a magic number, let's declare it as a constant instead

src/main/java/core/basesyntax/FigureBehaviour.java
```diff
@@ -0,0 +1,12 @@
...
+
+public interface FigureBehaviour {
+
+    String getName();
+
+    double getArea();
+
+    void printUniqueParameters();
+
+    Color getColor();
```
RomanDubovskyi, line 11: we can declare this method in the abstract class

src/main/java/core/basesyntax/FigureBehaviour.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+public interface FigureBehaviour {
+
+    String getName();
+
+    double getArea();
+
+    void printUniqueParameters();
```
RomanDubovskyi, line 9: let's remove this method, instead create in each class unique method like getRadius()

src/main/java/core/basesyntax/IsoscelesTrapezoid.java
```diff
@@ -0,0 +1,41 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class IsoscelesTrapezoid extends Figure {
+    private int sides = 0;
+    private int bottomBase = 0;
+    private int topBase = 0;
```
RomanDubovskyi, line 8: these values should be initialized in the constructor

RomanDubovskyi (changes requested):
for generating random figures creat FigureProducer, like Bohdan showed you

src/main/java/core/basesyntax/IsoscelesTrapezoid.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    @Override
+    public double getArea() {
+        return (topBase + bottomBase) / 2 * height;
+    }
+
+    @Override
+    public void draw() {
+        System.out.println("Shape: " + name + ", area: "
```
RomanDubovskyi, line 34:
```suggestion
        System.out.println("Shape: " + getClass().getSimpleName() + ", area: "
```
we can remove name variable

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -0,0 +1,29 @@
+package core.basesyntax;
+
+public class Rectangle extends Figure {
+    private String name = "rectangle";
```
RomanDubovskyi, line 4: same here and in other classes

RomanDubovskyi (approved):

PR #50, opened 2021-01-09
added solution

programmer28:


src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,23 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+    private int radius = 5;
```
boroda4436, line 4: why 5? This is hardcode

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,23 @@
...
+
+    public Circle() {
+        this.setName("circle");
```
boroda4436, line 7: hardcode. Better to accept values as a constructor parameters

src/main/java/core/basesyntax/IsoscelesTrapezoid.java
```diff
@@ -0,0 +1,24 @@
+package core.basesyntax;
+
+public class IsoscelesTrapezoid extends Figure {
+    private int height = 2;
```
boroda4436, line 4: hardcode. This field value should be received from user

boroda4436 (approved):
approved but code is not good

src/main/java/core/basesyntax/Circle.java
programmer28: will it be right when I replace initialization to constructor and make it with random positive integer?

src/main/java/core/basesyntax/Circle.java
programmer28: Okay, I got it

src/main/java/core/basesyntax/IsoscelesTrapezoid.java
programmer28: Okay, I got it

src/main/java/core/basesyntax/Circle.java
programmer28: Okay, I`ve understood

PR #52, opened 2021-03-01
Solution from Sak Vladyslav

SakVladyslav:


src/main/java/core/basesyntax/Behaviour.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax;
+
+public interface Behaviour {
+    double getArea();
```
boroda4436, line 4: this method could be moved to BaseFigure class

src/main/java/core/basesyntax/Behaviour.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax;
+
+public interface Behaviour {
```
boroda4436, line 3:
```suggestion
public interface Drawable {
```

src/main/java/core/basesyntax/IsoscelesTrapezoid.java
```diff
@@ -0,0 +1,28 @@
+package core.basesyntax;
+
+public class IsoscelesTrapezoid extends BaseFigure implements Behaviour {
+    private int base1;
+    private int base2;
+    private int height;
+
+    public IsoscelesTrapezoid(int base1, int base2, int height) {
+        super();
```
boroda4436, line 9: this constructor should be called by default. No needs to call it manually

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -0,0 +1,24 @@
+package core.basesyntax;
+
+public class Rectangle extends BaseFigure implements Behaviour {
+    private int width;
+    private int height;
+
+    public Rectangle(int width, int height) {
+        super();
```
boroda4436, line 8: same as for IsoscelesTrapezoid

src/main/java/core/basesyntax/RightTriangle.java
```diff
@@ -0,0 +1,25 @@
+package core.basesyntax;
+
+public class RightTriangle extends BaseFigure implements Behaviour {
+    private int base;
+    private int height;
+
+    public RightTriangle(int base, int height) {
+        super();
```
boroda4436, line 8: same as for IsoscelesTrapezoid

src/main/java/core/basesyntax/Square.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax;
+
+public class Square extends BaseFigure implements Behaviour {
+    private double side;
+
+    public Square(double side) {
+        super();
```
boroda4436, line 7: same as for IsoscelesTrapezoid

boroda4436 (approved):

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private static final String[] colorList =
+            new String[]{"yellow", "orange", "blue", "white", "black", "brown", "purple", "green"};
+
+    public static String getRandomColor() {
+        return colorList[new Random().nextInt(colorList.length - 1)];
```
kieraCurtis, line 10: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [DMI: Random object created and used only once (DMI_RANDOM_USED_ONLY_ONCE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#dmi-random-object-created-and-used-only-once-dmi-random-used-only-once).
This code creates a java.util.Random object, uses it to generate one random number, and then discards the Random object. This produces mediocre quality random numbers and is inefficient. If possible, rewrite the code so that the Random object is created once and saved, and each time a new random number is required invoke a method on the existing Random object to obtain it.
If it is important that the generated Random numbers not be guessable, you *must* not create a new Random for each random number; the values are too easily guessable. You should strongly consider using a java.security.SecureRandom instead (and avoid allocating a new SecureRandom for each random number needed).

PR #126, opened 2021-06-02
HW solution

aaananas:


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public class Main {
+    public static void main(String[] args) {
+        int numberOfFigures = 5;
```
Sofasmile, line 5: let's create a constant for 5 
https://mate-academy.github.io/style-guides/java/java.html#s5.2.4-constant-names

src/main/java/core/basesyntax/Main.java
```diff
@@ -1,12 +1,13 @@
 package core.basesyntax;
 
 public class Main {
+    static final int NUMBER_OF_FIGURES = 5;
```
Sofasmile, line 4:
```suggestion
    private static final int NUMBER_OF_FIGURES = 5;
```

Sofasmile (approved):

PR #161, opened 2021-07-13
OOP_Advanced solution

lena-kravchenko:


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,35 @@
...
+    public static final int FIGURE_SIDE = 50;
+    private Random random = new Random();
+
+    public int getRandomNumber() {
+        return random.nextInt(FIGURE_SIDE);
+    }
+
+    public Figure getRandomFigure() {
+        int figureNumber = random.nextInt(FIGURE_COUNT);
+        ColorSupplier supplier = new ColorSupplier();
```
RomanDubovskyi, line 16: make color supplier a class variable as well. Like you did with Random object. We don't need to create each object per method call

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,35 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    public static final int FIGURE_COUNT = 5;
```
RomanDubovskyi, line 6: If we don't use these constants outside of class it's recommended to make access modifier as strict as possible to follow encapsulation principle

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/FigureSupplier.java
lena-kravchenko: done

src/main/java/core/basesyntax/FigureSupplier.java
lena-kravchenko: done

RomanDubovskyi (approved):

PR #170, opened 2021-08-09
Jv oop advanced hw2 solution

vadim9482:
Hi, please check my pull request, CONSTANTs were moved to proper class and access modifier was changed to "private", return type of getArea was changed to "double" type, draw methods were updated with double formatting.

### DIFF
--- a/src/main/java/core/basesyntax/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/FigureSupplier.java
@@ -1,40 +1,46 @@
 package core.basesyntax;
 
+import static core.basesyntax.MainApp.FIGURE_COUNT;
+
 public class FigureSupplier {
     public Figure getRandomFigure() {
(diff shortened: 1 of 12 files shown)


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,48 @@
+package core.basesyntax;
+
+public class FigureSupplier {
+    private static final int FIGURE_COUNT = 5;
+    private static final int MAX_LENGTH = 5;
+    private static final int MIN_LENGTH = 1;
+
```
RomanDubovskyi, line 7:
```suggestion
    private static final int MIN_LENGTH = 1;
    private ColorSupplier colorSupplier = new ColorSupplier();
```
move declaration to a class instead of method

RomanDubovskyi (approved):
Great!

PR #175, opened 2021-08-16
created Figure class and 5 subclasses. Created AreCalculator and Figu…

AlexRozv:
…reDrawer interfaces. Created classes to get random color and figure.

### DIFF
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,0 +1,35 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    public static final int COLOR_COUNT = 10;
+    private Random random = new Random();
+
+    public String getRandomColor() {
+        int colorNumber = random.nextInt(COLOR_COUNT);
+
+        switch (colorNumber) {
+            case 1:
+                return "Black";
+            case 2:
+                return "Blue";
+            case 3:
+                return "Yellow";
+            case 4:
+                return "Red";
+            case 5:
+                return "Green";
+            case 6:
+                return "Pink";
+            case 7:
+                return "Brown";
+            case 8:
+                return "White";
+            case 9:
+                return "Violet";
+            default:
+                return "Orange";
+        }
+    }
+}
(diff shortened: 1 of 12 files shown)


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,35 @@
...
+public class ColorSupplier {
+    public static final int COLOR_COUNT = 10;
+    private Random random = new Random();
+
+    public String getRandomColor() {
+        int colorNumber = random.nextInt(COLOR_COUNT);
+
+        switch (colorNumber) {
+            case 1:
+                return "Black";
```
Max-Pochepets, line 14: we can have less colors, but let's create a separate `enum Color` and take values from it, without switch

src/main/java/core/basesyntax/FigureDrawer.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface FigureDrawer {
+    void drawFigure();
```
Max-Pochepets, line 4:
```suggestion
    void draw();
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,31 @@
...
+    public static final int FIGURE_COUNT = 5;
+    public static final int MAX_SIDE_LENGTH = 100;
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        int figureNumber = random.nextInt(FIGURE_COUNT);
+        int firstSide = random.nextInt(MAX_SIDE_LENGTH);
+        int secondSide = random.nextInt(MAX_SIDE_LENGTH);
+        int thirdSide = random.nextInt(MAX_SIDE_LENGTH);
```
Max-Pochepets, line 15: what if it will be a `Circle`? We'll need only one side, but we will still create three more

Max-Pochepets (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
AlexRozv: created enum Color instead of switch case

src/main/java/core/basesyntax/FigureSupplier.java
AlexRozv: optional sides are now created only in case where they are needed

Max-Pochepets (changes requested):
👍

Max-Pochepets (approved):

PR #189, opened 2021-08-18
Added jv-oop-adv solution

andrii-demchenko:


src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,30 @@
+package core.basesyntax;
+
+public class Circle extends Figure implements FigureDrawer, AreaCalculator {
```
Max-Pochepets, line 3:
```suggestion
public class Circle extends Figure {
```
these interfaces should be accessible through parent class

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,51 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    public static final int FIGURE_COUNT = 5;
```
Max-Pochepets, line 6:
```suggestion
    public static final int FIGURE_TYPE_COUNT = 5;
```

Max-Pochepets (changes requested):
👍

Max-Pochepets (approved):

PR #232, opened 2021-09-09
added solution hw-2

YarmoshykYa:


src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax;
+
+public interface Figure {
```
MaliukDaria, line 3: I think Figure its something that has state, maybe its better to make figure as abstract class, not interface? And lets create 2 interfaces for `getArea()` and `draw()` methods (some `AreaCalculator` and `Drawer` interfaces, or smth like this)

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,46 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    public static String getRandomFigure() {
```
MaliukDaria, line 6: it's better to awoid to use `static` methods, lets make `getRandomFigure()` not static

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,37 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    public String getRandomColor() {
+        Random r = new Random();
```
MaliukDaria, line 7:
```suggestion
        Random random = new Random();

```

dont forget about informative variable names

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,37 @@
...
+                break;
+            case (6):
+                color = "blue";
+                break;
+            case (7):
+                color = "purple";
+                break;
+            default:
+                color = "";
+        }
```
MaliukDaria, line 34: we can replace this huge switch-case with array of colors `private static final String[] COLORS = new String[]{"red", "green", "blue", ...}`, than get random number from 0 to COLORS.length as index and get array element of that index

src/main/java/core/basesyntax/FigureSupplier.java
MaliukDaria, line 6: I think this method should return Figure, not String from behaviour() method

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class MainApp extends FigureSupplier {
+    public static void main(String[] args) {
+        Random r = new Random();
+        int qf = r.nextInt(5) + 1;
```
MaliukDaria, line 8: not informative variable name

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax;
+
+public class Circle extends ColorSupplier implements Figure {
```
MaliukDaria, line 3: lets use out ColorSupplier in FigureSupplier class(and remove it from all figures classes), and generate Random colour for each figure which is created there

MaliukDaria (changes requested):

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,29 @@
+package core.basesyntax;
+
+public class Circle extends Figure implements AreaCalculator, Draw {
```
MaliukDaria, line 3:
```suggestion
public class Circle extends Figure {

```

if Figure implements AreaCalculator and Draw than all it's children also implements them

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    public String getRandomColor() {
+        Random random = new Random();
+        int number = random.nextInt(7) + 1;
+        String[] colors = {"red","orange","yellow","green","light blue","blue","purple"};
```
MaliukDaria, line 9: lets make this array constant field

src/main/java/core/basesyntax/Draw.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Draw {
```
MaliukDaria, line 3:
```suggestion
public interface Drawer {
```

dont forget about interface naming

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,17 @@
...
+        int sideB;
+        int height;
+        int firstLeg;
+        int secondLeg;
+        int radius;*/
+    }
+
+    public String behaviour() {
+        return null;
+    }
```
MaliukDaria, line 16: lets remove this methods, we dont need them. Instead lets create here `color` field with setters

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,43 @@
...
+    public Figure getRandomFigure() {
+        Figure temp = null;
+        Random figure = new Random();
+        int num = figure.nextInt(5) + 1;
+        int sideA = Math.abs(figure.nextInt(10) + 1);
+        int sideB = Math.abs(figure.nextInt(10) + 1);
+        int height = Math.abs(figure.nextInt(10) + 1);
+        int firstLeg = Math.abs(figure.nextInt(10) + 1);
+        int secondLeg = Math.abs(figure.nextInt(10) + 1);
+        int radius = Math.abs(figure.nextInt(10) + 1);
```
MaliukDaria, line 15: lets generate this fields when we create a certain figure

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,43 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier extends ColorSupplier {
+    public Figure getRandomFigure() {
+        Figure temp = null;
```
MaliukDaria, line 7:
```suggestion
        Figure figure = null;

```

not informative variable names

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -0,0 +1,16 @@
...
+public class MainApp {
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Random figure = new Random();
+        int num = figure.nextInt(5) + 1;
+
+        for (int i = 0; i < num; i++) {
+            Figure randomFigure = figureSupplier.getRandomFigure();
+            System.out.println(randomFigure.behaviour());
+        }
```
MaliukDaria, line 14: you need to generate random number of random figures

MaliukDaria (changes requested):

src/main/java/core/basesyntax/MainApp.java
YarmoshykYa: but it's generate random number of random figures, or I don't understand something. Please explain what should I do

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class MainApp {
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Random figure = new Random();
```
RomanDubovskyi, line 8:
```suggestion
        Random random = new Random();

```
that's not quite a figure

RomanDubovskyi (approved):

PR #239, opened 2021-09-14
Sp01 jv oop advanced

SerhiiPet:


src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,22 @@
...
+    @Override
+    public double getArea() {
+        return Math.PI * radius * radius;
+    }
+
+    @Override
+    public String draw() {
+        String a = "Figure: " + this.getClass().getSimpleName() + ", area: " + getArea();
+        String b = " sq.units, radius: " + radius + " units, color: " + getRandomColor();
+        return a + b;
```
RomanDubovskyi, line 19:
```suggestion
        return "Figure: " + this.getClass().getSimpleName() + ", area: " + getArea();
        + " sq.units, radius: " + radius + " units, color: " + getRandomColor();
```
a and b are not informative, it's better to return a string, or give it some meaningful name

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax;
+
+public class ColorSupplier {
+
```
RomanDubovskyi, line 4:
```suggestion
   private String[] colors = { "white", "black".... };
```
we don't need to create new array and fill it with values per each method call

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,20 @@
...
+        stringsColor[1] = "black";
+        stringsColor[2] = "green";
+        stringsColor[3] = "yellow";
+        stringsColor[4] = "blue";
+        stringsColor[5] = "red";
+        stringsColor[6] = "brown";
+        stringsColor[7] = "gray";
+        stringsColor[8] = "purple";
+        stringsColor[9] = "violet";
+        int colorNumber = (int) (Math.random() * 10);
```
RomanDubovskyi, line 17:
```suggestion
        int colorNumber = (int) (Math.random() * stringColors.length);
```

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public abstract class Figure extends FigureSupplier {
```
RomanDubovskyi, line 3:
```suggestion
public abstract class Figure {
```
Figure does not extend the functionality of FigureSupplier, they have different purpose. Figure - object representing real life figure. FigureSupplier - generating random values and figures based on them.

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,8 @@
...
+
+    public abstract double getArea();
```
RomanDubovskyi, line 5: behaviour is better to more to interface. 
Here you can store common state of all figures (color for example)

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,34 @@
+package core.basesyntax;
+
+public class FigureSupplier extends ColorSupplier {
```
RomanDubovskyi, line 3:
```suggestion
public class FigureSupplier {
    private ColorSupplier colorSupplier = new ColorSupplier();
```
The same here they have different purpose. You can use one inside of another but it won't be correct to extend it.

src/main/java/core/basesyntax/RightTriangle.java
```diff
@@ -0,0 +1,23 @@
...
+    @Override
+    public double getArea() {
+        return firstLeg * secondLeg / 2;
+    }
+
+    @Override
+    public String draw() {
+        return "Figure: " + this.getClass().getSimpleName() + ", area: "
+                + getArea() + " sq.units, firstLeg: " + firstLeg
+                + " units, secondLeg: " + secondLeg + " units, color: " + getRandomColor();
```
RomanDubovskyi, line 21: this figure doesn't have color, it just generates random one for console print. Create field according to task description

src/main/java/core/basesyntax/Square.java
```diff
@@ -0,0 +1,25 @@
...
+    public String draw() {
+        return "Figure: " + this.getClass().getSimpleName() + ", area: "
+                + getArea() + " sq.units, side: " + side + " units, color: " + getRandomColor();
+    }
+}
+
+
+
+
+
```
RomanDubovskyi, line 25: too many empty lines at the end, leave only single one after 
}

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,39 @@
+package core.basesyntax;
+
+public class FigureSupplier {
+
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    private double randomDouble() {
```
RomanDubovskyi, line 7: приватні методи повинні йти після публічних

src/main/java/core/basesyntax/Square.java
RomanDubovskyi: > too many empty lines at the end, leave only single one after
> }

цей комент не пофікшений

src/main/java/core/basesyntax/Figure.java
RomanDubovskyi: > behaviour is better to more to interface.
> Here you can store common state of all figures (color for example)

цей комент не пофікшений (винесення методів getArea(), draw() ) в інтерфейси

RomanDubovskyi (changes requested):
almost there

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,21 @@
+package core.basesyntax;
+
+public class Circle extends Figure implements AreaCalculator {
```
RomanDubovskyi, line 3:
```suggestion
public class Circle extends Figure {
```
це теж можна в Figure винести

RomanDubovskyi (approved):

PR #240, opened 2021-09-23
solution is done

alexgash8:
Can't understand, why another solution with extended loop (commented under working solution with classical for-loop statements in file start.java) doesn`t work and doesn`t assign Figure object to fig variable.

### DIFF
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,0 +1,8 @@
+package core.basesyntax;
+
+public class ColorSupplier {
+    public static String getRandomColor() {
+        String[] colors = {"red", "blue", "white", "black", "pink", "green"};
+        return colors[(int)(Math.random() * 5)];
+    }
+}
--- a/src/main/java/core/basesyntax/Drawable.java
+++ b/src/main/java/core/basesyntax/Drawable.java
@@ -1,0 +1,7 @@
+package core.basesyntax;
+
+public interface Drawable {
+    default void draw() {
+
+    }
+}
--- a/src/main/java/core/basesyntax/Figure.java
+++ b/src/main/java/core/basesyntax/Figure.java
@@ -1,0 +1,9 @@
+package core.basesyntax;
+
+public abstract class Figure implements Drawable, AreaCalculator {
+    private final String color = ColorSupplier.getRandomColor();
+
+    public String getColor() {
+        return color;
+    }
+}
--- a/src/main/java/core/basesyntax/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/FigureSupplier.java
@@ -1,0 +1,24 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    public static final int FIGURE_COUNT = 4;
+    private Random random = new Random();
+
+    public Figure getRandomFigure() {
+        int figureNumber = random.nextInt(FIGURE_COUNT);
+        switch (figureNumber) {
+            case 0:
+                return new Square();
+            case 1:
+                return new Circle();
+            case 2:
+                return new Triangle();
+            case 3:
+                return new Trapezoid();
+            default:
+                return new Square();
+        }
+    }
+}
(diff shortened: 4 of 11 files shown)


src/main/java/core/basesyntax/Drawable.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax;
+
+public interface Drawable {
+    default void draw() {
```
RomanDubovskyi, line 4: let's make this method abstract, so it will be necessary to override it in implementations

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public abstract class Figure implements Drawable, AreaCalculator {
+    private final String color = ColorSupplier.getRandomColor();
```
RomanDubovskyi, line 4:
```suggestion
    private final String color;
```
let's initialize this field in the constructor

src/main/java/core/basesyntax/Triangle.java
```diff
@@ -0,0 +1,33 @@
+package core.basesyntax;
+
+public class Triangle extends Figure implements Drawable, AreaCalculator {
```
RomanDubovskyi, line 3:
```suggestion
public class Triangle extends Figure {
```
that's redundant since Figure already implements these interfaces

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public class ColorSupplier {
+    public static String getRandomColor() {
+        String[] colors = {"red", "blue", "white", "black", "pink", "green"};
```
RomanDubovskyi, line 5:
```suggestion
public class ColorSupplier {
	private String[] colors = {"red", "blue", "white", "black", "pink", "green"};

    public String getRandomColor() {
```
it's better to make this field a class variable. This way we won't create new object of color per each method call

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public abstract class Figure implements Drawable, AreaCalculator {
+    private final String color;
+
+    public Figure() {
+        this.color = ColorSupplier.getRandomColor();
```
RomanDubovskyi, line 7:
```suggestion
    public Figure(String color) {
        this.color = color;
```
not recommended to use random as default behavior

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,31 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    public static final int FIGURE_COUNT = 4;
+    private Random random = new Random();
```
RomanDubovskyi, line 7: this variable is not used in the code

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,31 @@
...
+
+    public Figure getRandomFigure() {
+        int figureNumber = random.nextInt(FIGURE_COUNT);
+        switch (figureNumber) {
+            case 0:
+                double rnd = (Math.random() * 10 + 1);
```
RomanDubovskyi, line 13: it's better to make 10 a constant, in case you want to change it to 12 for example - you will need to do it only in one place instead of all code occurrence

src/main/java/core/basesyntax/Start.java
```diff
@@ -0,0 +1,36 @@
...
+            figures[i] = supplier.getRandomFigure();
+        }
+
+        System.out.println(figures.length);
+
+        for (int i = 0; i < figures.length; i++) {
+            figures[i].draw();
+        }
+
+        //        for (Figure fig : figures) {
```
RomanDubovskyi, line 21: let's remove commented code

RomanDubovskyi (changes requested):
almost there 👍

src/main/java/core/basesyntax/FigureSupplier.java
alexgash8: it is used in 10 & 11 lines

PR #242, opened 2021-09-29
Lost branch

alexgash8:
I have made some changes, and it was a mess with detached HEAD, and branches, hope i've solved everything clearly...

### DIFF
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,0 +1,8 @@
+package core.basesyntax;
+
+public class ColorSupplier {
+    public static String getRandomColor() {
+        String[] colors = {"red", "blue", "white", "black", "pink", "green"};
+        return colors[(int)(Math.random() * 5)];
+    }
+}
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,8 +1,9 @@
 package core.basesyntax;
 
 public class ColorSupplier {
-    public static String getRandomColor() {
-        String[] colors = {"red", "blue", "white", "black", "pink", "green"};
+	private String[] colors = {"red", "blue", "white", "black", "pink", "green"};
+
+    public String getRandomColor() {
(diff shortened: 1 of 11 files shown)


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public class ColorSupplier {
+    private String[] colors = {"red", "blue", "white", "black", "pink", "green"};
+
+    public String getRandomColor() {
+        return colors[(int)(Math.random() * 5)];
```
RomanDubovskyi, line 7:
```suggestion
        return colors[(int)(Math.random() * colors.length)];
```
this way if you add/remove smt from the colors it won't break this code

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,33 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    public static final int SIDE = 10;
```
RomanDubovskyi, line 6: if variable is used only inside the class it's recommended to make it priave to follow the encapsulation principle

RomanDubovskyi (changes requested):
GJ, almost there

src/main/java/core/basesyntax/ColorSupplier.java
alexgash8: super!

src/main/java/core/basesyntax/FigureSupplier.java
alexgash8: at least got it))

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,33 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int SIDE = 10;
```
RomanDubovskyi, line 6:
```suggestion
    private static final int SIDE_LIMIT = 10;
```

RomanDubovskyi (approved):

PR #258, opened 2021-10-20
added classes: Figure, Circle, Rectangle, Square, RightTriangle, Isos…

TankevychO:
…celesTrapezoid; created an array of figures and generated random parameters

### DIFF
--- a/src/main/java/core/basesyntax/Figure.java
+++ b/src/main/java/core/basesyntax/Figure.java
@@ -1,6 +1,6 @@
 package core.basesyntax;
 
-public abstract class Figure implements GetArea, GetInfo {
+public abstract class Figure implements AreaCalculator, PrintInfo {
     private String color;
 
     public Figure(String color) {
--- a/src/main/java/core/basesyntax/AreaCalculator.java
+++ b/src/main/java/core/basesyntax/AreaCalculator.java
@@ -1,5 +1,5 @@
 package core.basesyntax;
 
-public interface GetArea {
+public interface AreaCalculator {
     double getArea();
 }
--- a/src/main/java/core/basesyntax/PrintInfo.java
+++ b/src/main/java/core/basesyntax/PrintInfo.java
@@ -1,5 +1,5 @@
 package core.basesyntax;
 
-public interface GetInfo {
+public interface PrintInfo {
     String getInfo();
 }


src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,21 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+    private int radius;
+
+    public Circle() {
+        this.radius = getNumber();
+    }
```
Sofasmile, line 8: we shouldn't hardcode values in the constructor
```suggestion
    public Circle(String color, int radius) {
        ...
    }
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    public static final int COLOR_COUNT = 8;
```
Sofasmile, line 6: do we need this field?

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,21 @@
...
+        return color.getRandomColor();
+    }
+
+    public int getNumber() {
+        return random.nextInt(MAX_NUMBER);
+    }
+
+    public abstract double getArea();
+
+    public abstract String toString();
```
Sofasmile, line 20: the correct way is to create two separate interfaces with these methods 
then we should implement these interfaces in the Figure class and override in Circle, Rectuncle,... classes

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,24 @@
...
+                return new Circle();
+            case ("RECTANGLE"):
+                return new Rectangle();
+            case ("SQUARE"):
+                return new Square();
+            case ("RIGHT_TRIANGLE"):
+                return new RightTriangle();
+            default:
+                return new IsoscelesTrapezoid();
+        }
```
Sofasmile, line 22:
```suggestion
        switch (figureNumber) {
            case 1:
                return new Circle();
            case 2:
                return new Rectangle();
            ...
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,24 @@
...
+
+public class FigureSupplier {
+    public static final int FIGURE_COUNT = 5;
+    private Random random = new Random();
+
+    public Figure getRandomFigure() {
+        int figureNumber = random.nextInt(FIGURE_COUNT);
+        switch (FigureName.values()[figureNumber].toString()) {
+            case ("CIRCLE"):
+                return new Circle();
```
Sofasmile, line 13: we should pass the parameters in the constructor (color, radius...)

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+public class Main {
+    public static void main(String[] args) {
+        Figure[] figures = new Figure[6];
```
Sofasmile, line 5: let's create a constant  for 6 number

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,12 @@
...
+        FigureSupplier figure = new FigureSupplier();
+        for (int i = 0; i < figures.length; i++) {
+            figures[i] = figure.getRandomFigure();
```
Sofasmile, line 8: please read task more attentive, we should have getDefaultFigure() method too

Sofasmile (changes requested):

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,35 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private Random random = new Random();
+
+    public Figure getDefaultFigure() {
+        return new Circle("WHITE", 10);
```
Sofasmile, line 9:
```suggestion
        return new Circle(Color.WHITE.toString(), 10);
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    public int getNumber() {
```
Sofasmile, line 12:
```suggestion
    private int getNumber() {
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,35 @@
...
+        return random.nextInt(10);
+    }
+
+    public Figure getRandomFigure() {
+        ColorSupplier colorSupplier = new ColorSupplier();
+        String color = colorSupplier.getRandomColor();
+        int figureNumber = random.nextInt(FigureName.values().length);
+        switch (figureNumber) {
+            case (0):
```
Sofasmile, line 21: the same in other case blockes
```suggestion
            case 0:
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,18 @@
+package core.basesyntax;
+
+public class Main {
+    public static final int FIGURE_COUNT = 6;
```
Sofasmile, line 4:
```suggestion
    private static final int FIGURE_COUNT = 6;
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,18 @@
...
+
+    public static void main(String[] args) {
+        Figure[] figures = new Figure[FIGURE_COUNT];
+        FigureSupplier figure = new FigureSupplier();
```
Sofasmile, line 8:
```suggestion
        FigureSupplier figureSupplier = new FigureSupplier();
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    public int getNumber() {
+        return random.nextInt(10);
+    }
+
+    public Figure getRandomFigure() {
+        ColorSupplier colorSupplier = new ColorSupplier();
+        String color = colorSupplier.getRandomColor();
+        int figureNumber = random.nextInt(FigureName.values().length);
```
Sofasmile, line 19: we can use FigureName to generate `figureNumber` or we can create and use constant to do the same
```suggestion
        int figureNumber = random.nextInt(NUMBER_OF_FIGURES);
```

PR #271, opened 2021-10-20
my solution v1.0

PaulMartsiniv:


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,22 @@
...
+import core.basesyntax.figure.Figure;
+import core.basesyntax.figure.FigureSupplier;
+
+public class Main {
+    private static final int MAX_NUMBER = 6;
+
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Figure randomFigure = figureSupplier.getRandomFigure();
+        Figure defaultFigure = figureSupplier.getDefaultFigure();
```
Sofasmile, line 12: let's create all objects in the for loop

src/main/java/core/basesyntax/figure/Figure.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax.figure;
+
+import core.basesyntax.misc.AreaCalculator;
+import core.basesyntax.misc.Drawable;
+
+public abstract class Figure implements Drawable,AreaCalculator {
```
Sofasmile, line 6:
```suggestion
public abstract class Figure implements Drawable, AreaCalculator {
```

src/main/java/core/basesyntax/figure/Figure.java
```diff
@@ -0,0 +1,22 @@
...
+    private String color;
+    private String type;
```
Sofasmile, line 8: we can skip type field because it the same as class name

src/main/java/core/basesyntax/figure/FigureSupplier.java
```diff
@@ -0,0 +1,42 @@
...
+
+public class FigureSupplier {
+    private static final int UNITS = 100;
+    private Random random = new Random();
+    private ColorSupplier colorSupply = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        int r1 = random.nextInt(UNITS);
+        int r2 = random.nextInt(UNITS);
+        int r3 = random.nextInt(UNITS);
```
Sofasmile, line 19: let's generate random value only if we need it
in case when we will have Decagon will we create ten new variables with a random values?

PR #274, opened 2021-10-21
Initial commit, according to the task

MTWIB:


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+    
+    public Figure getRandomRightTriangle() {
+        String color = new ColorSupplier().getRandomColor();
+        int randomFirstLeg = random.nextInt(MAX_RANDOM_NUMBER);
+        int randomSecondLeg = random.nextInt(MAX_RANDOM_NUMBER);
+        return new RightTriangle(randomFirstLeg, randomSecondLeg, color);
+    }
+    
+    public Figure getRandomSquare() {
```
Nick97-git, line 39:
```suggestion
    private Figure getRandomSquare() {
```

Nick97-git: the same with other methods that used only in this class

PR #348, opened 2022-01-18
Hw solution

Sasha-Bachynskyi:


src/main/java/ColorSupplier.java
```diff
@@ -0,0 +1,7 @@
+import java.util.Random;
+
+public class ColorSupplier {
+    public String getRandomColor() {
+        return String.valueOf(Colors.values()[new Random().nextInt(Colors.values().length)]);
```
AlexRozv, line 5: it is optional, but I'd suggest using method toString() instead of String.valueOf()

src/main/java/FigureInfo.java
```diff
@@ -0,0 +1,3 @@
+public interface FigureInfo {
```
AlexRozv, line 1: let's fix naming FigureInfo -> FigurePrinter or Printable

src/main/java/FigureSupplier.java
```diff
@@ -0,0 +1,37 @@
+import java.util.Random;
+
+public class FigureSupplier {
+    public static final int AMOUNT_OF_FIGURES = 5;
+    public static final int BORDER = 5;
+    private Random random = new Random();
+
+    public Figure getRandomFigure() {
+        int randomNumber = random.nextInt(AMOUNT_OF_FIGURES - 1);
```
AlexRozv, line 9: if you're decreasing it by 1, why don't you just change AMOUNT_OF_FIGURES value from 5 to 4?

src/main/java/FigureSupplier.java
```diff
@@ -0,0 +1,37 @@
...
+
+        switch (randomNumber) {
+            case 1:
+                return new Square(new Random().nextInt(BORDER - 1) + 1,
+                        new ColorSupplier().getRandomColor());
```
AlexRozv, line 14: mate-academy.github.io/jv-program-common-mistakes/java-core/abstract-class-interface/oop-advanced#%D1%81reating-a-figure-dont-pass-expressions-in-constructor

we can create separate methods to create random figures, for example, getRandomSquare, inside this method you should create random variables that you will pass to the constructor and then return triangle with random values. And then you can use these methods inside your switch/case

AlexRozv (changes requested):

src/main/java/ColorSupplier.java
Sasha-Bachynskyi: I've tried, and the code hasn't compiled. What is the difference between these two methods?
Bohdan used String.valueOf() method in his example

src/main/java/FigureSupplier.java
Sasha-Bachynskyi: It's a border between 1 (including) and 5 (not including)

src/main/java/FigureSupplier.java
```diff
@@ -0,0 +1,37 @@
...
+                return new RightTriangle(new Random().nextInt(BORDER - 1) + 1,
+                        new Random().nextInt(BORDER - 1) + 1,
+                        new ColorSupplier().getRandomColor());
+            case 3:
+                return new Circle(new Random().nextInt(BORDER - 1) + 1,
+                        new ColorSupplier().getRandomColor());
+            default:
+                return new IsoscelesTrapezoid(new Random().nextInt(BORDER - 1) + 1,
+                        new Random().nextInt(BORDER - 1) + 1,
+                        new Random().nextInt(BORDER - 1) + 1,
```
AlexRozv, line 29: let's create Random outside the switch/case and use this object

AlexRozv (changes requested):

AlexRozv (approved):
good job 🔥

PR #359, opened 2022-01-19
1) created: parent class Figure and

Viktoriia-Zinchenko:
inherited it with five classes with the type of figures; enum Colors; 2 interfaces with methods
2) created method which random generates colors in class ColorSupplier
3) created getDefaultFigure method which by default will return a circle with the given parameters and getRandomFigure method which generates figures with random properties
4) created in the Main method an array of figures

### DIFF
--- a/src/main/java/core/basesyntax/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/FigureSupplier.java
@@ -1,0 +1,36 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    public static final int FIGURE_COUNT = 4;
+    public static final int MAX_VALUE = 10;
+
+    public Figure getDefaultFigure() {
+        return new Circle("white", MAX_VALUE);
+    }
+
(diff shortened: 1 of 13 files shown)


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,36 @@
...
+
+    public Figure getRandomFigure() {
+        switch (new Random().nextInt(FIGURE_COUNT)) {
+            case 0:
+                return new Square(new ColorSupplier().getRandomColor(),
+                        new Random().nextInt(MAX_VALUE));
+            case 1:
+                return new Rectangle(new ColorSupplier().getRandomColor(),
+                        new Random().nextInt(MAX_VALUE),
+                        new Random().nextInt(MAX_VALUE));
```
AlexRozv, line 21: let's create Random outside the switch/case and use this object

AlexRozv (changes requested):

AlexRozv (approved):
great job 🔥

PR #364, opened 2022-01-20
Realised oop-advanced task

DenTima:


src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,32 @@
+package core.basesyntax;
+
+public class Circle extends Figure implements DrawFigure, AreaCalculator {
```
AlexRozv, line 3: if you implement DrawFigure and AreaCalculator in Figure class you don't have to implement them in every separate figure class because they will be inherited

AlexRozv (approved):
good job 🔥

PR #368, opened 2022-01-20
First commit of task Figure

o240789:


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,41 @@
...
+            case 3:
+                randomFigure = new Rectangle(randomColor, random.nextDouble(), random.nextDouble());
+                break;
+            case 4:
+                randomFigure = new RightTriangle(randomColor,
+                        random.nextDouble(), random.nextDouble());
+                break;
+            default:
+                randomFigure = null;
+        }
```
AlexRozv, line 34: it's a bad practice to return null, let's use default case instead of case 4 to return one of the random figures. But don't forget to decrease FIGURE_QUANTITY by 1

AlexRozv (changes requested):

AlexRozv (approved):
good job 🔥

PR #381, opened 2022-02-14
added solution for jv-oop-advanced task

dmnichz:


src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,28 @@
+package core.basesyntax;
+
+public class Circle extends Figure implements Area {
```
MaliukDaria, line 3:
```suggestion
public class Circle extends Figure {
```

if class Figure implements some interface, then all it's heirs will implement this interface automatically

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    public String getRandomColor() {
+        return Color.values()[new Random().nextInt(Color.values().length)].toString().toLowerCase();
```
MaliukDaria, line 7: Lets make Random class level parameter (not local), than we won't create new instance of Random class every time when we call getRandomColor() method

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,29 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    public Figure getRandomFigure() {
+        ColorSupplier colorSupplier = new ColorSupplier();
```
MaliukDaria, line 7: same here, lets make ColorSupplier and Random as class level parameters

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,29 @@
...
+                new Random().nextInt(100));
+        Rectangle rectangle = new Rectangle(colorSupplier.getRandomColor(),
+                new Random().nextInt(100),
+                new Random().nextInt(100));
+        RightTriangle rightTriangle = new RightTriangle(colorSupplier.getRandomColor(),
+                new Random().nextInt(100),
+                new Random().nextInt(100));
+        Square square = new Square(colorSupplier.getRandomColor(),
+                new Random().nextInt(100));
+        Figure[] figures = {circle, isoscelesTrapezoid, rectangle, rightTriangle, square};
```
MaliukDaria, line 22: We create here 5 shapes in order to choose one of them (it turns out we don’t use the other 4 anywhere). Let's create 5 private methods in this class getRandomCircle(), getRandomSquare(),... and call these methods in the switch-case construction.

MaliukDaria (changes requested):

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,64 @@
...
+    private RightTriangle getRandomRightTriangle() {
+        return new RightTriangle(colorSupplier.getRandomColor(),
+                random.nextInt(100),
+                random.nextInt(100));
+    }
+
+    private Square getRandomSquare() {
+        return new Square(colorSupplier.getRandomColor(),
+                random.nextInt(100));
+    }
```
MaliukDaria, line 35: private methods should follow after public as helper methods. Also we can make number 100 as constant field

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,64 @@
...
+                figure = getRandomRectangle();
+                break;
+            case 3 :
+                figure = getRandomRightTriangle();
+                break;
+            case 4 :
+                figure = getRandomSquare();
+                break;
+            default:
+                break;
```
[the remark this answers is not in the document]
MaliukDaria: what about this comment?
```suggestion
                figure = getDefaultFigure();
                break;
```

MaliukDaria (changes requested):

MaliukDaria (approved):
👍🏻

PR #398, opened 2022-03-17
Implement figures model and logic

dSkarzh:


src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+    private final int radius;
```
Sofasmile, line 4: It is optional to make this field final, better don't do it often

Sofasmile (approved):
awesome

PR #407, opened 2022-03-23
add solution to hw

DimaPilipenko:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+
```
Sofasmile, line 6: remove empty line after class declaration

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,67 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    public static final int FIGURE_COUNT = 5;
+    public static final int DEFAULT_RADIUS = 10;
+    public static final int RANDOM_MAX_NUMBER = 100;
```
Sofasmile, line 8: we use these constants only in this class so we can make them private

Sofasmile (changes requested):

Sofasmile (approved):

PR #412, opened 2022-03-24
Added solution

GerasimovTA-92:


Sofasmile (changes requested):
ти не запушив всі файли, додай будь ласка решту

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,28 @@
...
+ * Feel free to remove this class and create your own.
+ */
+public class Main {
+    public static void main(String[] args) {
+        firstHalf();
+        System.out.println("-------------------------------------");
+        secondHalf();
+    }
+
+    public static void firstHalf() {
```
Sofasmile, line 15:
```suggestion
    private static void firstHalf() {
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,28 @@
...
+        FigureSupplier figure = new FigureSupplier();
+        for (int i = 0; i < 5; i++) {
+            figure.getRandomFigure().drawing();
+        }
+    }
+
+    public static void secondHalf() {
```
Sofasmile, line 22:
```suggestion
    private static void secondHalf() {
```

src/main/java/model/Area.java
```diff
@@ -0,0 +1,5 @@
+package model;
+
+public interface Area {
```
Sofasmile, line 3:
```suggestion
public interface AreaCalculator {
```

src/main/java/model/Circle.java
```diff
@@ -0,0 +1,26 @@
+package model;
+
+public class Circle extends Figure implements Area, Draw {
```
Sofasmile, line 3:
```suggestion
public class Circle extends Figure {
```
we can implement these interfaces once in the class Figure
it will be enough

src/main/java/model/Draw.java
```diff
@@ -0,0 +1,5 @@
+package model;
+
+public interface Draw {
+    void drawing();
```
Sofasmile, line 4: the name of the interface should be a noun  or adjective
and the name of the method should start from the verb

src/main/java/model/Figure.java
```diff
@@ -0,0 +1,21 @@
...
+        return color;
+    }
+
+    public double getArea() {
+        return 0;
+    }
+
+    public void drawing() {
+        System.out.println("This is figure!");
+    }
```
Sofasmile, line 20: we can remove the implementation of these methods from the Figure class, because class is abstract

src/main/java/model/Trapezoid.java
```diff
@@ -0,0 +1,24 @@
+package model;
+
+public class Trapezoid extends Figure implements Area, Draw {
```
Sofasmile, line 3:
```suggestion
public class Trapezoid extends Figure {
```

src/main/java/model/Square.java
```diff
@@ -0,0 +1,22 @@
+package model;
+
+public class Square extends Figure implements Draw, Area {
```
Sofasmile, line 3:
```suggestion
public class Square extends Figure {
```

src/main/java/model/RightTriangle.java
```diff
@@ -0,0 +1,22 @@
+package model;
+
+public class RightTriangle extends Figure implements Area, Draw {
```
Sofasmile, line 3:
```suggestion
public class RightTriangle extends Figure {
```

src/main/java/model/Rectangle.java
```diff
@@ -0,0 +1,25 @@
+package model;
+
+public class Rectangle extends Figure implements Area, Draw {
```
Sofasmile, line 3:
```suggestion
public class Rectangle extends Figure {
```

Sofasmile (changes requested):
🔥

Sofasmile (approved):

PR #417, opened 2022-03-25
Added all needed classes and methods

NesstorFedak:


src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,26 @@
+package core.basesyntax;
+
+public class Circle extends Figure implements Description, GetArea {
```
Sofasmile, line 3:
```suggestion
public class Circle extends Figure {
```
we have already implemented these interfaces in class Figure, it is enough

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,26 @@
...
+        this.radius = radius;
+    }
+
+    public int getRadius() {
+        return radius;
+    }
+
+    @Override
+    public void getDescription() {
+        System.out.println("Name: " + name + " radius: " + getRadius() + " units "
```
Sofasmile, line 18:
```suggestion
        System.out.println("Name: circle " + " radius: " + getRadius() + " units "
```
and let's remove field `private String name = "circle";` from the class
and do the same for all other figures

src/main/java/core/basesyntax/IsolateTrapezoid.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax;
+
+public class IsolateTrapezoid extends Figure implements Description, GetArea {
```
Sofasmile, line 3:
```suggestion
public class IsolateTrapezoid extends Figure {
```
the same everywhere

Sofasmile (changes requested):

Sofasmile (approved):

KseniiaMakarova (approved):

PR #422, opened 2022-03-27
Create app to display figures with random and default values

lexzane:


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,43 @@
...
+    public Figure getDefaultFigure() {
+        return defaultCircle;
+    }
+
+    public int getRandomValue() {
+        return random.nextInt(MAX_UNIT_LIMIT) + MIN_UNIT_LIMIT;
+    }
+
+    public String getRandomColor() {
+        return new ColorSupplier().getRandomColor();
```
KseniiaMakarova, line 41: let's make ColorSupplier a field - we don't need to create new object each time on method call

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,43 @@
...
+    }
```
KseniiaMakarova, line 42: can we make these methods private? they are just helper methods

src/main/java/core/basesyntax/model/Circle.java
```diff
@@ -0,0 +1,25 @@
+package core.basesyntax.model;
+
+import java.text.DecimalFormat;
+
+public class Circle extends Figure implements AreaCalculator {
```
KseniiaMakarova, line 5: let's make Figure class implement 2 interfaces: AreaCalculator and Drawable

KseniiaMakarova, line 5:
```suggestion
public class Circle extends Figure {
```

src/main/java/core/basesyntax/model/Circle.java
```diff
@@ -0,0 +1,25 @@
...
+    private int radius;
+    private DecimalFormat decimalFormat = new DecimalFormat("0.0");
```
KseniiaMakarova, line 7: is it a constant field?

src/main/java/core/basesyntax/model/Circle.java
```diff
@@ -0,0 +1,25 @@
...
+        this.radius = radius;
+    }
+
+    @Override
+    public double getArea() {
+        return Double.parseDouble(decimalFormat.format(Math.PI * radius * radius));
+    }
+
+    @Override
+    public String toString() {
```
KseniiaMakarova, line 20:
```suggestion
    public void draw() {
```

src/main/java/core/basesyntax/model/Figure.java
```diff
@@ -0,0 +1,15 @@
...
+
+    public Figure(String color) {
+        this.color = color;
+    }
+
+    public String getColor() {
+        return color;
+    }
+
+    public abstract String toString();
```
KseniiaMakarova, line 14: there is no need to declare methods that are already present in Object class

KseniiaMakarova (changes requested):
nice job!

KseniiaMakarova (approved):

PR #424, opened 2022-03-27
added solution

mordast7:


src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,29 @@
+package core.basesyntax;
+
+public class Circle extends Figure implements BehaviorFigure {
```
KseniiaMakarova, line 3:
```suggestion
public class Circle extends Figure {
```
let Figure class implement interfaces, so all its children are also implementing interfaces

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,29 @@
...
+        return "Figure: "
+                + "circle, "
+                + "area: "
+                + getArea()
+                + " sq.units, "
+                + "radius: "
+                + radius
+                + " units, "
+                + "color: "
+                + getColor();
```
KseniiaMakarova, line 27:
```suggestion
    public void draw() {
        System.out.println("Figure: "
                + "circle, "
                + "area: "
                + getArea()
                + " sq.units, "
                + "radius: "
                + radius
                + " units, "
                + "color: "
                + getColor());
```
let's create a separate interface Drawable with method draw()
and make Figure class implement 2 interfaces, because the purposes of the interfaces are different: 1st calculates the area, second is printing out whole information

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+public enum Colors {
```
KseniiaMakarova, line 3:
```suggestion
public enum Color {
```
enum name should better be singular

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,23 @@
+package core.basesyntax;
+
+public class Figure {
```
KseniiaMakarova, line 3:
```suggestion
public abstract class Figure implements AreaCalculator, Drawable {
```
abstract class doesn't have to override methods

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.Random;
+
+public class FigureSupplier {
+    public static final int MAX_BOUND = 100;
+
+    public Figure getRandomFigure() {
+        switch (new Random().nextInt(5)) {
+            case 0 :
+                return new Square(new ColorSupplier().getRandomColor(),
+                        new Random().nextInt(MAX_BOUND));
```
KseniiaMakarova, line 12: let's save Random object into a local variable
we can reuse it

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,41 @@
...
+                return new Circle(new ColorSupplier().getRandomColor(),
+                        new Random().nextInt(MAX_BOUND));
+
+            case 4 :
+                return new IsoscelesTrapezoid(new ColorSupplier().getRandomColor(),
+                        new Random().nextInt(MAX_BOUND),
+                        new Random().nextInt(MAX_BOUND),
+                        new Random().nextInt(MAX_BOUND));
+            default:
+                return new Figure(new ColorSupplier().getRandomColor());
```
KseniiaMakarova, line 34: same with ColorSupplier object

src/main/java/core/basesyntax/BehaviorFigure.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface BehaviorFigure {
```
KseniiaMakarova, line 3:
```suggestion
public interface AreaCalculator {
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+/**
+ * Feel free to remove this class and create your own.
+ */
```
KseniiaMakarova, line 5: remove this comment

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,15 @@
...
+public class Main {
+    public static void main(String[] args) {
+        Figure[] array = new Figure[8];
+
+        for (int i = 0; i < array.length; i++) {
+            array[i] = new FigureSupplier().getRandomFigure();
```
KseniiaMakarova, line 11: second half should be default figures

KseniiaMakarova (changes requested):

PR #489, opened 2022-05-19
hw-3

zimmboco:


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,66 @@
...
+            case RECTANGLE:
+                return getRandomRectangle();
+            case SQUARE:
+                return getRandomSquare();
+            case TRAPEZOID:
+                return getRandomTrapezoid();
+            case TRIANGLE:
+                return getRandomTriangle();
+            default:
+                return null;
```
MaliukDaria, line 30:
```suggestion
                return getDefaultFigure();
```

[checklist:](https://mate-academy.github.io/jv-program-common-mistakes/java-core/abstract-class-interface/oop-advanced) Don’t return null from a method.
Returning null from a method is a bad practice. If you use switch-case in your solution, you may just put the last possible option in the default case.

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax;
+
+public class Main {
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] figuresArray = new Figure[6];
+
+        for (int i = 0; i < figuresArray.length; i++) {
+            if (i < 3) {
```
MaliukDaria, line 9:
```suggestion
            if (i < figuresArray.length/2) {
```

MaliukDaria (changes requested):

Nick97-git (changes requested):

Nick97-git (approved):

PR #496, opened 2022-05-19
Jv oop advanced solution

mykola8426:


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,31 @@
...
+            return new Square(colorSupplier.getRandomColor(), random.nextInt(100));
+        } else if ("rightTriangle".equals(array[index])) {
+            return new RightTriangle(colorSupplier.getRandomColor(),
+                    random.nextInt(100), random.nextInt(100));
+        } else if ("rectangle".equals(array[index])) {
+            return new Rectangle(colorSupplier.getRandomColor(),
+                    random.nextInt(100), random.nextInt(100));
+        } else if ("isoscelesTrapezoid".equals(array[index])) {
+            return new IsoscelesTrapezoid(colorSupplier.getRandomColor(), random.nextInt(100),
+                     random.nextInt(100), random.nextInt(100));
```
KostyaNikiforov, line 23: Can we move some logic to a separate method to avoid code duplication?

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,31 @@
...
+        }
+        return new Circle(colorSupplier.getRandomColor(),random.nextInt(100));
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle("WHITE",10);
```
KostyaNikiforov, line 29: Let's make these values constant too

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private Random random;
```
KostyaNikiforov, line 6:
```suggestion
    private final Random random;
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,31 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private final Random random = new Random();
+    private final ColorSupplier colorSupplier = new ColorSupplier(random);
```
KostyaNikiforov, line 7: Let's initialize these fields in the constructor just as you would in the ColorCupplyer class.

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum Colors {
```
KostyaNikiforov, line 3: Enum name must be singular
```suggestion
public enum Color {
```

KostyaNikiforov (changes requested):

PR #505, opened 2022-05-23
Implemented FigureSupplier class who generate random figure with rand…

DrTymkivstain:
…om properties, 5 classes of figures, enum Colors with colors, 2 interfaces

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -48,6 +48,14 @@
                     <linkXRef>false</linkXRef>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <source>17</source>
+                    <target>17</target>
+                </configuration>
+            </plugin>
         </plugins>
         <pluginManagement>
             <plugins>
--- a/pom.xml
+++ b/pom.xml
@@ -19,7 +19,6 @@
     </properties>
 
     <dependencies>
-        <!-- https://mvnrepository.com/artifact/junit/junit -->
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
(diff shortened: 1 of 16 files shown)


src/main/java/core/basesyntax/Draw.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Draw {
```
liliia-ponomarenko, line 3:
```suggestion
public interface Drawable  {
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,16 @@
...
+public class ColorSupplier {
+    private final Random random;
+
+    public ColorSupplier(Random random) {
+        this.random = random;
+    }
+
+    public String getRandomColor() {
+        int index = random.nextInt(Colors.values().length);
+        return String.valueOf(Colors.values()[index]);
```
liliia-ponomarenko, line 14: You can use method name() to get the name of the enum element

pom.xml
```diff
@@ -19,6 +19,7 @@
     </properties>
 
     <dependencies>
+        <!-- https://mvnrepository.com/artifact/junit/junit -->
```
liliia-ponomarenko, line 22:
```suggestion
```

src/main/java/core/basesyntax/Draw.java
```diff
@@ -0,0 +1,5 @@
...
+    void name();
```
liliia-ponomarenko, line 4:
```suggestion
    void drawFigure();
```

liliia-ponomarenko (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
DrTymkivstain: Done?

PR #510, opened 2022-06-01
jv-oop-advanced homework

denis-kravziuk:


pom.xml
```diff
@@ -9,7 +9,7 @@
     <version>1.0-SNAPSHOT</version>
 
     <properties>
-        <jdk.version>11</jdk.version>
+        <jdk.version>1.8</jdk.version>
```
MaliukDaria, line 12: why do we need to change this version?

pom.xml
```diff
@@ -22,7 +22,7 @@
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
-            <version>4.12</version>
+            <version>4.13.2</version>
```
MaliukDaria, line 25: why do we need to change this version?

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,31 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+
+    private static final double PI = Math.PI;
```
MaliukDaria, line 5: I dont think we need to make this value as constant, lets just use `Math.PI` where we need

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,13 @@
...
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private final Random random = new Random();
+
+    public String getRandomColor() {
+        int index = random.nextInt(Color.values().length);
+        return Color.values()[index].getColor();
+
```
MaliukDaria, line 11: redundant empty line

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,29 @@
...
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final Figure circle = new Circle(parameter, colorSupplier.getRandomColor());
+    private final Figure square = new Square(parameter, colorSupplier.getRandomColor());
+    private final Figure rectangle = new Rectangle(parameter, parameter,
+            colorSupplier.getRandomColor());
+    private final Figure rightTriangle = new RightTriangle(parameter, parameter,
+            colorSupplier.getRandomColor());
+    private final Figure isoscelesTrapezoid = new IsoscelesTrapezoid(parameter,
+            parameter, parameter, colorSupplier.getRandomColor());
+    private final Figure[] figures = {circle, square, rectangle, rightTriangle, isoscelesTrapezoid};
```
MaliukDaria, line 19: you create 5 figures just to pick one of them. And every time you will choose figure from the same circle, square, ... when you call getRandomFigure() several times.  Lets make this variables locas and use switch-case construction. We can create 5 private methods getRandomCircle(), getRandomSquare(), ... and call them in switch-case construction, this will increase code readability

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,29 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    public static final int FIGURE_COUNT = 5;
+    public static final int RADIUS = 10;
```
MaliukDaria, line 7:
```suggestion
    public static final int DEFAULT_RADIUS = 10;
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,29 @@
...
+    private final Random random = new Random();
+    private final double parameter = random.nextDouble();
```
MaliukDaria, line 9: same, lets make this variable local

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,25 @@
...
+            } else {
+                figures[i] = defaultFigure;
+            }
+        }
+
+        for (Figure figure : figures) {
+            figure.draw();
+        }
+    }
+
```
MaliukDaria, line 24: loots of redundant empty lines

MaliukDaria (changes requested):

pom.xml
denis-kravziuk: in previous homework it caused problems. command mvn clean package gave me errors. so in pom file i changed these two settings and it worked. in this home work i made it from the start

pom.xml
denis-kravziuk: in previous homework it caused problems. command mvn clean package gave me errors. so in pom file i changed these two settings and it worked. in this home work i made it from the start

src/main/java/core/basesyntax/FigureSupplier.java
MaliukDaria: Only `private final double parameter = random.nextDouble();`

MaliukDaria: Only `private final double parameter = random.nextDouble();`

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.util.Random;
+
+public class FigureSupplier {
+    public static final int DEFAULT_RADIUS = 10;
+    private final Random random = new Random();
+    private final double parameter = random.nextDouble();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    private Figure getRandomCircle() {
```
MaliukDaria, line 11: all private methods should follow after public as helper ones

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #512, opened 2022-06-03
Version 3. Common mistakes fixed

AngelinaRo21:


src/main/java/core/basesyntax/AreaCalculator.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface AreaCalculator {
+    double getSquare();
```
MaliukDaria, line 4:
```suggestion
    double getArea();

```

src/main/java/core/basesyntax/Drawer.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Drawer {
+    void getDraw();
```
MaliukDaria, line 4:
```suggestion
    void draw();

```

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax;
+
+public class Figure implements Drawer, AreaCalculator {
```
MaliukDaria, line 3: do we need to have ability to create instances of Figure class?

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,29 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+    private int radius;
+
+    public Circle(int radius) {
```
MaliukDaria, line 6: I think we also need to set color to circle(same for other figures)

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,59 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int MAX_NUMBER = 100;
+    private static Type type;
```
MaliukDaria, line 7: Why do we need to make this variable class level?

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,59 @@
...
+
+public class FigureSupplier {
+    private static final int MAX_NUMBER = 100;
+    private static Type type;
+
+    public FigureSupplier(Type type) {
+        this.type = type;
+    }
+
+    public static Figure getRandomFigure() {
```
MaliukDaria, line 13:
```suggestion
    public Figure getRandomFigure() {

```

it is better to avoid using `static` https://medium.com/att-israel/should-you-avoid-using-static-ae4b58ca1de5

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,59 @@
...
+        Random random = new Random();
+        Random random3 = new Random();
+        Random random2 = new Random();
```
MaliukDaria, line 16: why do we need 3 random instances?

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,59 @@
...
+                figure = new RightTriangle(value1);
+                figure.setColor(String.valueOf(colorSupplier.getRandomColor()));
+                break;
+
+            default :
+                figure = new IsoscelesTrapezoid(value1, value2, value3);
+                figure.setColor(String.valueOf(colorSupplier.getRandomColor()));
+                break;
+        }
+        return figure;
```
MaliukDaria, line 51: Lets create 5 private methods getRandomCircle(), getRandomSquare(), ... and call them in switch-case construction, this will increase code readability. (Remember private methods should follow after public as helper ones)

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    public static Figure getDefaultFigure() {
+        Figure figure = new Circle(10);
+        figure.setColor("WHITE");
```
MaliukDaria, line 56: lets make number 10 and "WHITE" as constant fields

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,59 @@
...
+    public static Figure getRandomFigure() {
+        Random random = new Random();
+        Random random3 = new Random();
+        Random random2 = new Random();
+        int value1 = random.nextInt(MAX_NUMBER) + 1;
+        int value2 = random.nextInt(MAX_NUMBER) + 1;
+        int value3 = random.nextInt(MAX_NUMBER) + 1;
+        int index = random.nextInt(Type.values().length);
+        type = Type.values()[index];
+        ColorSupplier colorSupplier = new ColorSupplier();
```
MaliukDaria, line 22: same, let's make Random and ColorSupplier as class level variables, not to create new instances every time we call getRandomFigure() method.

src/main/java/core/basesyntax/Type.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum Type {
```
MaliukDaria, line 3:
```suggestion
public enum FigureType {

```

MaliukDaria (changes requested):

PR #546, opened 2022-07-09
jv-oop-advanced

Alina-Nikiforova:


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+public class Main {
+    public static final int SIZE_OF_ARRAY = 6;
```
MaliukDaria, line 4:
```suggestion
    public static final int NUMBER_OF_FIGURES = 6;
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.util.Random;
+
+public class FigureSupplier {
+    public static final int DEFAULT_RADIUS = 5;
+    public static final int MAX_NUMBER = 30;
+    public static final int AMOUNT_FIGURE = 5;
+    public static final String COLOR_CIRCLE = Color.WHITE.name();
+    private Random random = new Random();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
```
MaliukDaria, line 11:
```suggestion
    private static final int DEFAULT_RADIUS = 5;
    private static final int MAX_NUMBER = 30;
    private static final int AMOUNT_FIGURE = 5;
    private static final String COLOR_CIRCLE = Color.WHITE.name();
    private final Random random = new Random();
    private final ColorSupplier colorSupplier = new ColorSupplier();
```

we can make constants private if we don't use them outside the class

PR #550, opened 2022-07-10
Figure

oleh-satoshi:


src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,32 @@
...
+
+    @Override
+    public void setColor(Color color) {
+        super.setColor(color);
+    }
+
+    @Override
+    public Color getColor() {
+        return super.getColor();
+    }
```
KseniiaMakarova, line 30: there is no need to override these methods, they are inherited anyway

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,59 @@
...
+            case 2:
+                return getRandomRectangle();
+            case 3:
+                return getRandomRightTriangle();
+            default:
+                return getRandomSquare();
+        }
+    }
+
+    private double getRandomValue() {
```
KseniiaMakarova, line 29: let's move all private methods below public

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/IsoscelesTrapezoid.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Override
+    public void setColor(Color color) {
+        super.setColor(color);
+    }
+
+    @Override
+    public Color getColor() {
+        return super.getColor();
+    }
```
KseniiaMakarova, line 36: we can remove these overrides from all child figures

KseniiaMakarova (approved):

PR #574, opened 2022-07-27
created classes (all inherited from abstract class Figure), abstract …

veers-droid:
…class Figure has field color cause all classes have it, also this class implements interfaces: AreaCalculator and Canvas.

AreaCalculator has method countArea
Canvas has method Draw
in all classes I overrided methods what they inherited.
Created classes FigureSupplier which creates a random figure with random parameters or a figure with default parameters. class ColorSupplier help the last one to create a random color from enum class Color.
In Main I created object of FigureSupplier class and with its help created a list of figures and show their info via loop.

### DIFF
--- a/src/main/java/core/basesyntax/AreaCalculator.java
+++ b/src/main/java/core/basesyntax/AreaCalculator.java
@@ -1,0 +1,6 @@
+package core.basesyntax;
+
+public interface AreaCalculator {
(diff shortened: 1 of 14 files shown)


Noct2000 (changes requested):
don't use master or main branch for your commits

PR #583, opened 2022-07-28
Mate2

Litvik21:


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Feel free to remove this class and create your own.
+ */
+public class Main {
+    private static void print(Figure[] figures) {
+        for (Figure figure : figures) {
+            System.out.println(figure.getInfo());
+        }
+    }
```
Sofasmile, line 14: better move all private methods after public

Sofasmile (approved):
awesome

PR #610, opened 2022-07-29
home work done

badyaxa:


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,26 @@
...
+
+public class Main {
+    public static final int NUMBER_FIGURES = 12;
+    public static final int MAXIMUM_LENGTH = 100;
+
+    public static void main(String[] args) {
+        Figure[] figures = new Figure[NUMBER_FIGURES];
+        FigureSupplier figureSupplier = new FigureSupplier();
+        for (int i = 0; i < figures.length; i++) {
+            if (i % 2 == 0) {
```
KseniiaMakarova, line 15: The first half of figures in this array should be generated with random parameters.

src/main/java/core/basesyntax/model/Rectangle.java
```diff
@@ -0,0 +1,49 @@
+package core.basesyntax.model;
+
+public class Rectangle extends Figure {
+    private int firstLeg;
+    private int secondLeg;
```
KseniiaMakarova, line 5:
```suggestion
    private int width;
    private int height;
```

src/main/java/core/basesyntax/model/Square.java
```diff
@@ -0,0 +1,38 @@
+package core.basesyntax.model;
+
+public class Square extends Figure {
+    private int leg;
```
KseniiaMakarova, line 4:
```suggestion
    private int side;
```

KseniiaMakarova (changes requested):
good job

KseniiaMakarova (approved):

PR #616, opened 2022-07-31
added solution for task

Maksym-Korkosh:


src/main/java/core/basesyntax/Color.java
```diff
@@ -0,0 +1,12 @@
...
+
+public enum Color {
+    Black,
+    White,
+    Brown,
+    Yellow,
+    Red,
+    Green,
+    Blue,
+    Pink,
```
KseniiaMakarova, line 11: let's name them in UPPER_CASE (because these values are constants)

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/AreaFinder.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface AreaFinder {
+    double area();
```
KseniiaMakarova, line 4:
```suggestion
    double getArea();
```

KseniiaMakarova (approved):

PR #639, opened 2022-08-24
The task was completed according to the conditions

oleksandr-kravchenko:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    public static String getRandomColor() {
+        Random random = new Random();
+        int number = random.nextInt(Colors.values().length);
```
aaananas, line 8:
```suggestion
    private static final Random random = new Random();
    
    public static String getRandomColor() {
        int number = random.nextInt(Colors.values().length);
```

aaananas (changes requested):

PR #662, opened 2022-09-19
Solved the task using OOP principles

igorgalenko:
not allowed use "*" in imports
asked single class import instead
for instance:
import core.basesyntax.shapes.Circle;
import core.basesyntax.shapes.IsoscelesTrapezoid;
import core.basesyntax.shapes.Rectangle;
import core.basesyntax.shapes.RightTriangle;
import core.basesyntax.shapes.Shape;
import core.basesyntax.shapes.Square;
vs
import core.basesyntax.shapes.*;
why?

### DIFF
--- a/src/main/java/core/basesyntax/supliers/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/supliers/ColorSupplier.java
@@ -1,0 +1,11 @@
+package core.basesyntax.supliers;
+
+import core.basesyntax.interfaces.Color;
+import java.util.Random;
+
+public class ColorSupplier {
+    public String getRandomColor() {
+        Random random = new Random();
+        return Color.values()[random.nextInt(Color.values().length)].name().toLowerCase();
+    }
+}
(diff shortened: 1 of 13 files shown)


src/main/java/core/basesyntax/supliers/ColorSupplier.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax.supliers;
+
+import core.basesyntax.interfaces.Color;
+import java.util.Random;
+
+public class ColorSupplier {
+    public String getRandomColor() {
+        Random random = new Random();
```
hronograf, line 8: We can add field with type `Random` to our class, so we won't need to create an instance every time when method invoked

src/main/java/core/basesyntax/supliers/FigureSupplier.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int RANDOM_RADIUS = 10;
+    private static final int RANDOM_INT_BOUND = 7;
+    private static final int TOTAL_SHAPES = 5;
+
+    public Shape getRandomFigure() {
+        Random random = new Random();
+        String randomColor = new ColorSupplier().getRandomColor();
```
hronograf, line 19: We can add field with type `Random` to our class, so we won't need to create an instance every time when method invoked
same for `ColorSupplier`

hronograf (changes requested):
Cool! Let's improve this solution a bit)

PR #668, opened 2022-09-22
implemented task

vov4ik89:


src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,22 @@
...
+import core.basesyntax.interfaces.FigureInterface;
+import java.util.Random;
+
+public abstract class Figure implements FigureInterface {
+    protected Random random = new Random();
+    private String color;
+
+    public void setRandomProperties() {
+        this.color = new ColorSupplier().getRandomColor();
+    }
```
hronograf, line 12: do we need this method? (we can set fields through constructor)

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,27 @@
...
+import java.util.Random;
+
+public class FigureSupplier {
+    public static final int FIGURE_COUNT = 5;
+    public static final int DEFAULT_RADIUS = 10;
+    private final Random random = new Random();
+
+    public Figure getRandomFigure() {
+        Figure[] figures = new Figure[] {new Circle(), new Square(), new Rectangle(),
+                new RightTriangle(), new IsoscelesTrapezoid()};
```
hronograf, line 18: should we create instances of all figures if we want to return only one of them?

src/main/java/core/basesyntax/figures/Circle.java
```diff
@@ -0,0 +1,35 @@
...
+    public Circle(String color, int radius) {
+        super.setColor(color);
+        this.radius = radius;
+    }
+
+    @Override
+    public void setRandomProperties() {
+        super.setRandomProperties();
+        this.radius = random.nextInt(BOUND);
+    }
```
hronograf, line 22: do we need this method? (we can set fields through constructor)
Please, check it for all classes

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax;
+
+import core.basesyntax.interfaces.FigureInterface;
+import java.util.Random;
+
+public abstract class Figure implements FigureInterface {
```
hronograf, line 6: we also want to draw our figure:
> behaviour - we can obtain the area of any figure and are able to draw it. To 'draw' means to print out all information about a figure using System.out.println() (you shouldn't override toString() method for this).

we can create and implement one more interface for this purpose

hronograf (changes requested):
Nice!
And we can make it better)

hronograf (approved):
Awesome! 🚀

PR #675, opened 2022-09-26
solution was added

vyshnevsky:


liliia-ponomarenko (changes requested):
Let’s  improve your solution 😉

PR #677, opened 2022-09-27
Homework

alex-anofriev:


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,35 @@
...
+public class FigureSupplier {
+    private static final int MAX_FIGURE_SIZE = 10;
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final Random random = new Random();
+
+    public int figureSize() {
+        return random.nextInt(MAX_FIGURE_SIZE);
+    }
+
+    public Figure whichFigure() {
```
hronograf, line 14: why did you choose this name? (in the task was`getRandomFigure`)

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,35 @@
...
+            case CIRCLE:
+                return new Circle(colorSupplier.getRandomColor(), figureSize());
+            case SQUARE:
+                return new Square(colorSupplier.getRandomColor(), figureSize());
+            case RECTANGLE:
+                return new Rectangle(colorSupplier.getRandomColor(), figureSize(), figureSize());
+            case RIGHTTRIANGLE:
+                return new RightTriangle(colorSupplier.getRandomColor(),
+                        figureSize(), figureSize());
+            default: ISOSCELESTRAPEZOID:
```
hronograf, line 25: it would be better if for default case we will return default figure (`getDefaultFigure()` method)

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,35 @@
...
+    private static final int MAX_FIGURE_SIZE = 10;
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final Random random = new Random();
+
+    public int figureSize() {
+        return random.nextInt(MAX_FIGURE_SIZE);
+    }
+
+    public Figure whichFigure() {
+        switch (Figures.values()[random.nextInt(Figures.values().length)]) {
```
hronograf, line 15: it would be more readable if we introduced a variable for this statement

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,23 @@
+package core.basesyntax;
+
+public class Main {
+    private static final int taskNumber = 6;
+    private static final int halfOfTaskNumber = taskNumber / 2 - 1;
```
hronograf, line 5: it is better to change style of naming
Please, look at [Constant names](https://mate-academy.github.io/style-guides/java/java.html#s5.2.4-constant-names)

hronograf (changes requested):
Nice!
Let's improve solution a bit)

src/main/java/core/basesyntax/FigureSupplier.java
alex-anofriev: sorry, I did not pay attention there was prepared name in task for that

src/main/java/core/basesyntax/FigureListToDo.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public interface FigureListToDo {
+
+    String getName();
+
+    double area();
+}
```
hronograf, line 8: Sorry, my fault
in previous comment I meant that we need an interface only for `area` method
`getName` can be inside `Figure` class 

also, let's try to find a better name for this interface (after moving `getName` to the `Figure` class)

hronograf (changes requested):
Good!
One more change)

alex-anofriev:
Unfortunately I hardly see differences beween area and name and why area less personal thing for figure than name

hronograf:
it is ok and this difference isn't obvious
in this example let's imagine that we also have class `Flat`, that should have area but shouldn't have name
and also we have method `int calculateAreaDifference(FigureArea firstArea, FigureArea secondArea)` (just simple example)
so this method can calculate are difference for flats and figures, but flats don't have to have names

hronograf (approved):
Great! 💥

PR #680, opened 2022-09-28
homework oop advanced

Bodddd:


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,68 @@
...
+            case 3:
+                return getRandomRightTriangle();
+            case 4:
+                return getRandomSquare();
+            default:
+                return getDefaultFigure();
+        }
+    }
+
+    public Circle getRandomCircle() {
```
liliia-ponomarenko, line 31: make this and other  methods private

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,68 @@
...
+        String color = colorSupplier.getRandomColor();
+        double radius = random.nextDouble() * MAX_VALUE;
```
liliia-ponomarenko, line 33: Let’s  create a private method that returns a random value and avoid  code duplicates

liliia-ponomarenko (changes requested):
Great 😊  let’s  fix a few mistakes😉

liliia-ponomarenko (approved):
good job 😎

PR #683, opened 2022-09-28
Created abstract class Figure and interfaces Drawable and AreaCalcula…

PavloPolovyi:
…ble. Created classes Square, Rectangle, RightTriangle, Circle, IsoscelesTrapezoid. Created ColorSupplier and FigureSupplier. Created classes FigureDimensionError for custom error output and FigureDimensionsValidator for validation of figures dimensions.

### DIFF
--- a/src/main/java/core/basesyntax/Circle.java
+++ b/src/main/java/core/basesyntax/Circle.java
@@ -1,0 +1,24 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+    private final double radius;
+
+    public Circle(Color color, double radius) {
+        super(color);
+        if (!FigureDimensionsValidator.isValidDimensions(radius)) {
+            throw new FigureDimensionError(this.getClass().getSimpleName());
+        }
+        this.radius = radius;
+    }
+
+    @Override
+    public double getArea() {
+        return Math.PI * radius * radius;
+    }
+
+    @Override
+    public String draw() {
+        return super.draw() + " { radius: " + radius + " units, area: "
+                + getArea() + " square units, color: " + getColor() + " }";
+    }
+}
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private final Random randomGen = new Random();
+    private final Color[] colors = Color.values();
+    private final int numOfColor = colors.length;
+
+    public Color getRandomColor() {
+        return colors[randomGen.nextInt(numOfColor)];
+    }
+}
--- a/src/main/java/core/basesyntax/Drawable.java
+++ b/src/main/java/core/basesyntax/Drawable.java
@@ -1,0 +1,7 @@
+package core.basesyntax;
+
+public interface Drawable {
+    default String draw() {
+        return this.getClass().getSimpleName();
+    }
+}
--- a/src/main/java/core/basesyntax/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/FigureSupplier.java
@@ -1,0 +1,61 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    public static final Color DEFAULT_FIGURE_COLOR = Color.WHITE;
+    public static final double DEFAULT_SIDE_DIMENSION = 10.0;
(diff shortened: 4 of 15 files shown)


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,61 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    public static final Color DEFAULT_FIGURE_COLOR = Color.WHITE;
+    public static final double DEFAULT_SIDE_DIMENSION = 10.0;
+    private final String[] figuresClassList = {"Circle", "Square", "Rectangle",
+            "RightTriangle", "IsoscelesTrapezoid"};
+    private final int numberOfFigures = figuresClassList.length;
```
liliia-ponomarenko, line 10: the same about variables in fields

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,61 @@
...
+    private final Random randGen = new Random();
```
liliia-ponomarenko, line 11: Don’t use abbreviations in names

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,18 @@
+package core.basesyntax;
+
+public class Main {
+    public static final int TOTAL_NUMBER_OF_FIGURES = 6;
```
liliia-ponomarenko, line 4:
```suggestion
    private static final int TOTAL_NUMBER_OF_FIGURES = 6;
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,18 @@
...
+    public static void main(String[] args) {
+        FigureSupplier supplier = new FigureSupplier();
+        Figure[] figureList = new Figure[TOTAL_NUMBER_OF_FIGURES];
+        for (int i = 0; i < TOTAL_NUMBER_OF_FIGURES; i++) {
+            figureList[i] = i < NUMBER_OF_RANDOM_FIGURES ? supplier.getRandomFigure()
+                    : supplier.getDefaultFigure();
+        }
+        for (Figure figure : figureList) {
+            System.out.println(figure.draw());
+        }
```
liliia-ponomarenko, line 16: Can we use only one loop for this?

liliia-ponomarenko (changes requested):
Let’s  improve your solution😉

src/main/java/core/basesyntax/Drawable.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax;
+
+public interface Drawable {
+    default String draw() {
```
liliia-ponomarenko, line 4: why is this method default?

src/main/java/core/basesyntax/FigureSupplier.java
PavloPolovyi: What is abbreviations?

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,24 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+    private final double radius;
+
+    public Circle(Color color, double radius) {
+        super(color);
+        if (!FigureDimensionsValidator.isValidDimensions(radius)) {
+            throw new FigureDimensionError(this.getClass().getSimpleName());
+        }
```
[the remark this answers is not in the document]
PavloPolovyi: then how to validate the arguments of the constructor, if "validation is business logic and has no place in the constructor" ?

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private final Random randomGen = new Random();
+    private final Color[] colors = Color.values();
+    private final int numOfColor = colors.length;
```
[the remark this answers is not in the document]
PavloPolovyi: ColorSupplier - is only utility class and shouldn't be initialized by user. If i put these variables in method, they will be always created again when method called. In my case, FigureSupplier create instance of ColorSupplier only once and these variables are created once too.

src/main/java/core/basesyntax/FigureSupplier.java
PavloPolovyi: Same as my previous answer.

ChumakAn (approved):

PR #684, opened 2022-09-28
created figures classes, interfaces and enums

GrishaSeniv:
1) created 5 figures classes and AbstractFigure and moved it to figures package 
2) created AreaCalculator, PerimeterCalculator, Drawable interfaces 
3) implemented ColorSupplier, FigureSupplier
4) created Color and FigureNames enum

### DIFF
--- a/src/main/java/core/basesyntax/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/FigureSupplier.java
@@ -1,0 +1,49 @@
+package core.basesyntax;
+
+import core.basesyntax.figures.AbstractFigure;
+import core.basesyntax.figures.Circle;
+import core.basesyntax.figures.IsoscelesTrapezoid;
+import core.basesyntax.figures.Rectangle;
+import core.basesyntax.figures.RightTriangle;
+import core.basesyntax.figures.Square;
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int NUMBER_OF_FIGURES_BOUND = 4;
+    private static final int MAX_VALUE_BOUND = 100;
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final Random random = new Random();
+
+    public AbstractFigure getRandomFigure() {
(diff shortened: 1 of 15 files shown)


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,51 @@
...
+    private static final int NUMBER_OF_FIGURES_BOUND = 4;
+    private static final int MAX_VALUE_BOUND = 100;
+    private static final String DEFAULT_COLOR = "white";
+    private static final double DEFAULT_RADIUS = 10.0;
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final Random random = new Random();
+
+    public AbstractFigure getRandomFigure() {
+        String randomColor = colorSupplier.getRandomColor();
+        int randomFigureNumber = random.nextInt(NUMBER_OF_FIGURES_BOUND);
```
liliia-ponomarenko, line 21: this looks like code duplicates. Let’s  move it to a private  method

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,51 @@
...
+        if (randomFigureNumber == 0) {
```
liliia-ponomarenko, line 22: Let’s  use switch-cases

liliia-ponomarenko (changes requested):
wow, you add PerimeterCalculator😎 great job 👏  let’s  fix a few mistakes

src/main/java/core/basesyntax/FigureSupplier.java
GrishaSeniv: Is it good that I divided the classes into packages?

liliia-ponomarenko: It’s not good it is great 🔥 but give names of packages in the singular😉

liliia-ponomarenko (approved):
LGTM🔥

PR #685, opened 2022-09-28
Oop advanced task completed, all tests is passed

arthur4655:
maven wanted me to make the variables private, but private variables can't be inherited, so i use protected as more restrictive modifier.

### DIFF
--- a/src/main/java/core/basesyntax/Circle.java
+++ b/src/main/java/core/basesyntax/Circle.java
@@ -1,0 +1,28 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+    private int radius;
+
+    public Circle() {
+        this.radius = random.nextInt(VALUE_LIMIT);
+        this.color = colorSupplier.getRandomColor();
+        area();
+    }
+
+    public Circle(int radius, String color) {
+        this.radius = radius;
+        this.color = color;
+        area();
+    }
+
+    @Override
+    public void area() {
+        this.area = Math.PI * Math.pow(radius, 2);
+    }
+
+    @Override
+    public void draw() {
+        System.out.printf("Figure: circle, area: %.1f sq.units, radius: %d units, color: %s%n",
+                area, radius, color);
+    }
+}
--- a/src/main/java/core/basesyntax/Figure.java
+++ b/src/main/java/core/basesyntax/Figure.java
@@ -1,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+abstract class Figure implements AreaCalculator {
+    public static final int VALUE_LIMIT = 50;
+    protected String color;
+    protected double area;
+    protected Random random = new Random();
+    protected ColorSupplier colorSupplier = new ColorSupplier();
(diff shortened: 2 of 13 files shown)


src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,28 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+    private int radius;
+
+    public Circle() {
+        this.radius = random.nextInt(VALUE_LIMIT);
+        this.color = colorSupplier.getRandomColor();
+        area();
+    }
```
hronograf, line 10: do we need this constructor?
and should constructor without parameters returns figure with random fields?
it isn't obvious behaviour
(for all classes)

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+abstract class Figure implements AreaCalculator {
+    public static final int VALUE_LIMIT = 50;
```
hronograf, line 6: this field also will be redundant, after we remove `Random` and `ColorSupplier` fields

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,14 @@
...
+    protected String color;
+    protected double area;
```
hronograf, line 8: do we need area `field` here?
maybe we will be able to return our area just from the method

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,30 @@
...
+            case 2:
+                return new Square();
+            case 3:
+                return new IsoscelesTrapezoid();
+            case 4:
+                return new RightTriangle();
+            default:
+        }
+        return null;
+    }
```
hronograf, line 25: we should change it after other comments will be fixed
also, `return null` may be redundant, we can return default figure for the default case

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,30 @@
...
+
+    public Figure getDefaultFigure() {
+        return new Circle(10, "white");
```
hronograf, line 28: it would be better to use `ColorList` here, not just string

src/main/java/core/basesyntax/Quadrilateral.java
```diff
@@ -0,0 +1,14 @@
...
+    protected int side2;
+    protected int side3;
+    protected int side4;
+
+    @Override
+    public void draw() {
+        System.out.printf("area: %.1f sq.units, sides are: %d-%d-%d-%d units, color: %s%n",
+                area, side1,side2, side3, side4, color);
+    }
+}
```
hronograf, line 14: cool!

hronograf (approved):
Nice! 💥 
and @arthur4655 please, read the comment)

PR #687, opened 2022-09-28
Implemented the task

hpavlo:


hronograf (approved):
Wow, wow, wow!
On the first try!!! 🚀

PR #690, opened 2022-09-29
Solution

OleksiyMasaltsev:


src/main/java/core/basesyntax/figures/Printable.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax.figures;
+
+public interface Printable {
+    void draw();
+
+    default String getFormatted(double d) {
+        return String.format("%.2f", d);
+    }
```
hronograf, line 8: we shouldn't use interfaces as a place for utility methods
interfaces should say what methods object must have

src/main/java/core/basesyntax/suppliers/FigureSupplier.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    private Circle getRandomCircle() {
+        return new Circle(colorSupplier.getRandomColor(),
+                getRandomDouble());
+    }
+
+    private double getRandomDouble() {
+        return MIN_SIZE + (MAX_SIZE - MIN_SIZE) * random.nextDouble();
+    }
```
hronograf, line 79: nice

hronograf (changes requested):
Good!

src/main/java/core/basesyntax/figures/Figure.java
```diff
@@ -0,0 +1,23 @@
...
+        return color;
+    }
+
+    public void setColor(Color color) {
+        this.color = color;
+    }
+
+    String getFormatted(double d) {
+        return String.format("%.2f", d);
+    }
```
hronograf, line 22: I think we can use `String.format` in each `draw` method in subclasses

hronograf (approved):
Good! 🔥 
and @OleksiyMasaltsev check my comment)

src/main/java/core/basesyntax/figures/Figure.java
OleksiyMasaltsev: @hronograf  So, even if we have a similar implementation in every class, is it better to copy the method?

PR #693, opened 2022-09-29
create figure classes, Info and FigureOperations interfaces, implemen…

yehbreus:
…ted getRandomColor,getRandomFigure methods

### DIFF
--- a/src/main/java/core/basesyntax/Circle.java
+++ b/src/main/java/core/basesyntax/Circle.java
@@ -1,0 +1,20 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+    private int radius;
+
+    public Circle(int radius) {
+        this.radius = radius;
+        this.setType("circle");
+    }
+
+    public double getArea() {
+        return Math.PI * radius * radius;
+    }
+
+    @Override
+    public void printInfo() {
+        System.out.println("Figure: " + getType() + ", area: " + getArea()
+                + " units, radius: " + radius + " units, color: " + getColor());
+    }
+}
--- a/src/main/java/core/basesyntax/Classes.java
+++ b/src/main/java/core/basesyntax/Classes.java
@@ -1,0 +1,9 @@
+package core.basesyntax;
+
+public enum Classes {
+    Circle,
+    IsoscelesTrapezoid,
+    RightTriangle,
+    Rectangle,
+    Square
+}
--- a/src/main/java/core/basesyntax/Figure.java
+++ b/src/main/java/core/basesyntax/Figure.java
@@ -1,0 +1,22 @@
+package core.basesyntax;
+
+public abstract class Figure implements FigureOperations, Info {
+    private String color;
+    private String type;
+
+    public String getColor() {
+        return color;
+    }
+
+    public void setColor(String color) {
+        this.color = color;
+    }
+
+    public String getType() {
+        return type;
+    }
+
+    public void setType(String type) {
+        this.type = type;
+    }
+}
--- a/src/main/java/core/basesyntax/FigureOperations.java
+++ b/src/main/java/core/basesyntax/FigureOperations.java
@@ -1,0 +1,5 @@
+package core.basesyntax;
+
+public interface FigureOperations {
+    public double getArea();
+}
--- a/src/main/java/core/basesyntax/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/FigureSupplier.java
@@ -1,0 +1,45 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int DEFAULT_CIRCLE_RADIUS = 10;
+    private static final String DEFAULT_CIRCLE_COLOR = "White";
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        return createRandomFigure(Classes.values()[random.nextInt(Classes.values().length)]);
+    }
+
+    public Figure createRandomFigure(Classes types) {
+        Figure figure = null;
+        switch (types) {
+            case Circle :
+                figure = new Circle(random.nextInt());
+                break;
+            case Square:
+                figure = new Square(random.nextInt());
+                break;
+            case Rectangle:
+                figure = new Rectangle(random.nextInt(), random.nextInt());
+                break;
+            case RightTriangle:
+                figure = new RightTriangle(random.nextInt(), random.nextInt());
+                break;
+            case IsoscelesTrapezoid:
+                figure = new IsoscelesTrapezoid(random.nextInt(),
+                        random.nextInt(), random.nextInt());
+                break;
+            default:getDefaultFigure();
+        }
+        figure.setColor(colorSupplier.getRandomColor());
+        return figure;
+    }
+
+    public Figure getDefaultFigure() {
+        Figure figure = new Circle(DEFAULT_CIRCLE_RADIUS);
+        figure.setColor(DEFAULT_CIRCLE_COLOR);
+        return figure;
+    }
+}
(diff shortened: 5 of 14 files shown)


src/main/java/core/basesyntax/Info.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Info {
+    public void printInfo();
```
hronograf, line 4: `public` is redundant

src/main/java/core/basesyntax/FigureOperations.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface FigureOperations {
+    public double getArea();
```
hronograf, line 4: same

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,20 @@
...
+    private int radius;
+
+    public Circle(int radius) {
+        this.radius = radius;
+        this.setType("circle");
+    }
+
+    public double getArea() {
+        return Math.PI * radius * radius;
+    }
```
hronograf, line 13: it would be better if we had `@Override` annotation here 
(and in other classes)

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax;
+
+public abstract class Figure implements FigureOperations, Info {
+    private String color;
+    private String type;
```
hronograf, line 5: `Figure` class shouldn't know anything about a type of the figure
also it would be better if we could set `color` parameter by constructor

src/main/java/core/basesyntax/Classes.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum Classes {
+    Circle,
+    IsoscelesTrapezoid,
+    RightTriangle,
+    Rectangle,
+    Square
```
hronograf, line 8: [Constant names](https://mate-academy.github.io/style-guides/java/java.html#s5.2.4-constant-names)

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,45 @@
...
+                figure = new Rectangle(random.nextInt(), random.nextInt());
+                break;
+            case RightTriangle:
+                figure = new RightTriangle(random.nextInt(), random.nextInt());
+                break;
+            case IsoscelesTrapezoid:
+                figure = new IsoscelesTrapezoid(random.nextInt(),
+                        random.nextInt(), random.nextInt());
+                break;
+            default:getDefaultFigure();
```
hronograf, line 34: should we write here 
> defafault:
>     figure = ...

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,45 @@
...
+        }
+        figure.setColor(colorSupplier.getRandomColor());
```
hronograf, line 36: it would be better to pass color through constructor

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,16 @@
...
+
+public class Main {
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] figuresArray = {figureSupplier.getRandomFigure(),
+                figureSupplier.getRandomFigure(),
+                figureSupplier.getRandomFigure(),
+                figureSupplier.getDefaultFigure(),
+                figureSupplier.getDefaultFigure(),
+                figureSupplier.getDefaultFigure()};
```
hronograf, line 11: let's use cycle to fill this array

hronograf (changes requested):
Good!

src/main/java/core/basesyntax/Classes.java
```diff
@@ -1,9 +1,9 @@
 package core.basesyntax;
 
 public enum Classes {
```
KostyaNikiforov, line 3: Enum names must be singular 
And let's make this name more informative. What kind of classes? Your project can grow and it can be challenging to understand how to use it

src/main/java/core/basesyntax/Info.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+interface Info {
```
KostyaNikiforov, line 3: Doesn't this interface have to be accessed in other packages?

src/main/java/core/basesyntax/Info.java
```diff
@@ -0,0 +1,5 @@
...
+    public void printInfo();
```
KostyaNikiforov, line 4: Usually IDE highlights in a grey color those parts of code which can be deleted. Doesn't your EDI do it with the public key work here?

src/main/java/core/basesyntax/Square.java
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    @Override
+    public double getArea() {
+        return side * side;
+    }
+
+    @Override
+    public void printInfo() {
+        System.out.println("Figure: " + getType() + ", area: " + getArea() + " units, side: " + side
```
KostyaNikiforov, line 28: You can have a direct access to this field as it is in the save class. But i think it could be moved to the parent class Figure because this field is common for each class of this type.
```suggestion
        System.out.println("Figure: " + type + ", area: " + getArea() + " units, side: " + side
```

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/Draw.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Draw {
```
KostyaNikiforov, line 3:
```suggestion
public interface Drawable {
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,46 @@
...
+            case RIGHT_TRIANGLE:
+                figure = new RightTriangle(colorSupplier.getRandomColor(), random.nextInt(),
+                        random.nextInt());
+                break;
+            case ISOSCELES_TRAPEZOID:
+                figure = new IsoscelesTrapezoid(colorSupplier.getRandomColor(), random.nextInt(),
+                        random.nextInt(), random.nextInt());
+                break;
+            default:
+                figure = getDefaultFigure();
```
KostyaNikiforov, line 38:
```suggestion
            default:
                figure = new IsoscelesTrapezoid(colorSupplier.getRandomColor(), random.nextInt(),
                        random.nextInt(), random.nextInt());
                break;
```

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/Square.java
KostyaNikiforov: Okay, let's leave it as it is

src/main/java/core/basesyntax/FigureSupplier.java
KostyaNikiforov: It was in common mistakes

yehbreus: I think my code here is right. because readme.md file say next: The other half of figures should have the same, default parameters.For this purpose create a new method in the FigureSupplier class:
public Figure getDefaultFigure() - this method should always return white circle with radius 10. 
Also : https://github.com/mate-academy/jv-oop-advanced/pull/693#discussion_r984499728

yehbreus: oh, sorry. I see it. My bad.

KostyaNikiforov (approved):
😌

PR #704, opened 2022-09-30
Create Classes wtih implemented behaviours, random Figures and their …

Andew-Miroshnikov:
…methos

### DIFF
--- a/src/main/java/core/basesyntax/Draw.java
+++ b/src/main/java/core/basesyntax/Draw.java
@@ -1,0 +1,5 @@
+package core.basesyntax;
+
+public interface Draw {
+    public void draw();
+}
--- a/src/main/java/core/basesyntax/Figure.java
+++ b/src/main/java/core/basesyntax/Figure.java
@@ -1,0 +1,19 @@
+package core.basesyntax;
+
+public abstract class Figure implements AreaCalculator, Draw {
+    private String color;
+
+    public Figure(String color) {
+        this.color = color;
+    }
+
+    public String getColor() {
+        return color;
+    }
+
+    @Override
+    public abstract double getArea();
+
+    @Override
+    public abstract void draw();
+}
--- a/src/main/java/core/basesyntax/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/FigureSupplier.java
@@ -1,0 +1,30 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int FIGURE_LENGTH = Figures.values().length;
+    private static final Random random = new Random();
+    private final ColorSupplier supplier = new ColorSupplier();
+    private static final int randomSide = random.nextInt(100);
+
+    public Figure getRandomFigure() {
+        int index = random.nextInt(FIGURE_LENGTH);
+        if (Figures.values()[index].equals(Figures.CIRCLE)) {
+            return new Circle(randomSide, supplier.getRandomColor());
+        } else if (Figures.values()[index].equals(Figures.ISOSCELESTRAPEZOID)) {
+            return new IsoscelesTrapezoid(randomSide, randomSide, supplier.getRandomColor());
+        } else if (Figures.values()[index].equals(Figures.RIGHTTRIANGLE)) {
+            return new RightTriangle(randomSide, randomSide, supplier.getRandomColor());
+        } else if (Figures.values()[index].equals(Figures.RECTANGLE)) {
+            return new Rectangle(randomSide, randomSide, supplier.getRandomColor());
+        } else {
+            return new Square(randomSide, supplier.getRandomColor());
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(10, "white");
+    }
+
+}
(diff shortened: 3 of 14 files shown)


src/main/java/core/basesyntax/Draw.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Draw {
```
KostyaNikiforov, line 3:
```suggestion
public interface Drawable {

```

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,19 @@
...
+
+    public String getColor() {
+        return color;
+    }
+
+    @Override
+    public abstract double getArea();
+
+    @Override
+    public abstract void draw();
```
KostyaNikiforov, line 18: Do you really need this code here?)

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,30 @@
...
+            return new Circle(randomSide, supplier.getRandomColor());
+        } else if (Figures.values()[index].equals(Figures.ISOSCELESTRAPEZOID)) {
+            return new IsoscelesTrapezoid(randomSide, randomSide, supplier.getRandomColor());
+        } else if (Figures.values()[index].equals(Figures.RIGHTTRIANGLE)) {
+            return new RightTriangle(randomSide, randomSide, supplier.getRandomColor());
+        } else if (Figures.values()[index].equals(Figures.RECTANGLE)) {
+            return new Rectangle(randomSide, randomSide, supplier.getRandomColor());
+        } else {
+            return new Square(randomSide, supplier.getRandomColor());
+        }
```
KostyaNikiforov, line 23: Can you use switch-case instead of if-bloks?

src/main/java/core/basesyntax/Figures.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum Figures {
```
KostyaNikiforov, line 3: Enum names must be singular

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,19 @@
...
+
+        Figure[] figureArray = new Figure[figureAmount];
+        for (int i = 0; i < figureArray.length; i++) {
+            if (i < 3) {
+                figureArray[i] = figureSupplier.getRandomFigure();
+                figureArray[i].draw();
+            } else {
+                figureArray[i] = figureSupplier.getDefaultFigure();
+                figureArray[i].draw();
+            }
```
KostyaNikiforov, line 16:
```suggestion
            if (i < 3) {
                figureArray[i] = figureSupplier.getRandomFigure();
            } else {
                figureArray[i] = figureSupplier.getDefaultFigure();
            }
            figureArray[i].draw();
```

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -0,0 +1,34 @@
...
+    public double getArea() {
+
+        return width * height;
+    }
+
+    @Override
+    public void draw() {
+        System.out.println("Figure: rectangle, area: " + getArea() + " sq.units, "
+                + "height: " + getHeight() + " units, "
+                + "width: " + getWidth() + " units, color: "
```
KostyaNikiforov, line 31: You can get direct access to your fields, do not use getter inside the class

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -3,23 +3,24 @@
...
-            return new Square(randomSide, supplier.getRandomColor());
+        switch (index) {
+            case 0 :
+                return new Circle(randomSide, supplier.getRandomColor());
+            case 1 :
+                return new IsoscelesTrapezoid(randomSide, randomSide, supplier.getRandomColor());
+            case 2 :
+                return new RightTriangle(randomSide, randomSide, supplier.getRandomColor());
+            case 3 :
+                return new Rectangle(randomSide, randomSide, supplier.getRandomColor());
```
KostyaNikiforov, line 21: randomSide - is always the same value. It is that behavior that we expect from the method with the name getRandomFigure. Let's generate a new value inside the method

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/FigureSupplier.java
KostyaNikiforov: So your Rectangle will always have sides of the same length)
It is still not the expected behaviour

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -4,23 +4,25 @@
 
 public class FigureSupplier {
     private static final int FIGURE_LENGTH = 5;
+    private static final int MAX_RANDOM_INT = 100;
     private static final Random random = new Random();
+    private int randomSide1 = random.nextInt(MAX_RANDOM_INT);
+    private int randomSide2 = random.nextInt(MAX_RANDOM_INT);
```
KostyaNikiforov, line 10: It shoulb be on the method level

KostyaNikiforov (approved):

PR #708, opened 2022-09-30
Implemented Figure interface & FugiresSupplier

sequencerr:


src/main/java/core/basesyntax/suppliers/FigureSupplier.java
```diff
@@ -0,0 +1,47 @@
...
+                double height = getRandomDoubleValue();
+                return new IsoscelesTrapezoid(randomColor, randomDouble1, randomDouble2, height);
+        }
+    }
+
+    private double getRandomDoubleValue() {
+        return random.nextDouble() * (random.nextInt(MAX_RANDOM_MULTIPLIER) + 1);
+    }
+
+    public static Figure getDefaultFigure() {
```
sequencerr, line 43: Why we shouldn't use that method as static? 
Bohdan said, it much harder to create unit tests with static methods, but IMHO, it can cause performance issues and breaks the logic.

src/main/java/core/basesyntax/FigureInterface.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+interface FigureInterface {
```
ChumakAn, line 3:
```suggestion
interface Drawable {
```

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,27 @@
...
+
+    protected String getName() {
+        return getClass().getSimpleName().toLowerCase();
+    }
+
+    protected static String formatValue(double val) {
+        return String.format("%,.1f", val);
+    }
+
+    public abstract void draw();
```
ChumakAn, line 24: why do you repeat this?

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,27 @@
...
+
+    protected abstract double getArea();
```
ChumakAn, line 26: consider adding an interface AreaCalculator for this method

src/main/java/core/basesyntax/suppliers/ColorSupplier.java
```diff
@@ -0,0 +1,22 @@
...
+    public static final String[] COLORS = {
+            DEFAULT_COLOR,
+            "black",
+            "red",
+            "orange",
+            "grey",
+            "brown",
+            "lime",
+            "violet"
+    };
```
ChumakAn, line 16: it's better to make enum in a separate file 
also, all enums have their convention, you should write it like BLACK, RED , e.t.c

src/main/java/core/basesyntax/suppliers/FigureSupplier.java
```diff
@@ -0,0 +1,47 @@
...
+            case 2:
+                return new Square(randomColor, randomDouble1);
+            case 3:
+                return new Rectangle(randomColor, randomDouble1, randomDouble2);
+            default:
+                double height = getRandomDoubleValue();
+                return new IsoscelesTrapezoid(randomColor, randomDouble1, randomDouble2, height);
+        }
+    }
+
```
ChumakAn, line 38: consider going through the enum of shapes for example to make It more clear what cases do you have
for example
`case CIRCLE: `, not `case 3:`

src/main/java/core/basesyntax/suppliers/FigureSupplier.java
```diff
@@ -0,0 +1,47 @@
...
+    private double getRandomDoubleValue() {
+        return random.nextDouble() * (random.nextInt(MAX_RANDOM_MULTIPLIER) + 1);
+    }
+
+    public static Figure getDefaultFigure() {
+        return new Circle(DEFAULT_CIRCLE_RADIUS);
```
ChumakAn, line 44: it should also have a white color

src/main/java/core/basesyntax/suppliers/FigureSupplier.java
ChumakAn: you should remember about the rule that we don't use static method if it's not necessary 
we should not break the principles of clean code and some best practices, so in this situation it's better not to use static keyword and  create the object of Figure and then refer to the method getDefaultFigure()

ChumakAn (changes requested):
Let's make some changes to make your code better

src/main/java/core/basesyntax/suppliers/FigureSupplier.java
ChumakAn: it's the optional one, like a good practice

src/main/java/core/basesyntax/Figure.java
sequencerr: I have thought, that it's necessary. We need to specify that the method we implementing is abstract, or maybe change access modifier. Like so.

src/main/java/core/basesyntax/suppliers/FigureSupplier.java
sequencerr: I agree, that is a good practice.
But it is making code harder to support (every time you add new Figure class, you should update Figures enum), isn't it?
upd: oh, wait, I have to update figures_count constant every time too.

The part in mate code style recommendations, saying that the last Figure should be returned in default branch, is also confusing me. It breaks the random if we have wrong amount of figures. I would log an Error for default branch.

src/main/java/core/basesyntax/suppliers/FigureSupplier.java
```diff
@@ -0,0 +1,52 @@
...
+                return new Circle(randomColor, randomDouble1);
+            case RIGHT_TRIANGLE:
+                return new RightTriangle(randomColor, randomDouble1, randomDouble2);
+            case SQUARE:
+                return new Square(randomColor, randomDouble1);
+            case RECTANGLE:
+                return new Rectangle(randomColor, randomDouble1, randomDouble2);
+            default:
+                double height = getRandomDoubleValue();
+                return new IsoscelesTrapezoid(randomColor, randomDouble1, randomDouble2, height);
```
ChumakAn, line 40: use the getDefaultFigure() for default case

ChumakAn (changes requested):
Only one minor comment, well done

ChumakAn (approved):
LGTM

PR #715, opened 2022-10-02
implemented task

Veronika-Tretiak:


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,36 @@
...
+            case 1: return new Rectangle(colorSupplier.getRandomColor(),
+                    random.nextInt(LIMIT),
+                    random.nextInt(LIMIT));
+            case 2: return new RightTriangle(colorSupplier.getRandomColor(),
+                    random.nextInt(LIMIT),
+                    random.nextInt(LIMIT));
+            case 3: return new Circle(colorSupplier.getRandomColor(),
+                    random.nextInt(LIMIT));
+            case 4: return new IsoscelesTrapezoid(colorSupplier.getRandomColor(),
+                    random.nextInt(LIMIT),
```
ChumakAn, line 26: consider going through the enum of shapes for example to make It more clear what cases do you have
for example 
`case CIRCLE: , not case 3: `

ChumakAn (changes requested):
Only one small change, well done!

src/main/java/core/basesyntax/FigureSupplier.java
ChumakAn: it's an optional one, I can accept this and discuss it on Q&A

ChumakAn (approved):

PR #717, opened 2022-10-02
created classes of random shapes and created an array of these shapes

Matolch:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private static Random random;
+
+    public ColorSupplier(Random random) {
+        this.random = random;
+    }
```
ChumakAn, line 10: better to initialize this in 6 line

src/main/java/core/basesyntax/FigureArea.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax;
+
+public interface FigureArea {
```
ChumakAn, line 3:
```suggestion
public interface AreaCalculator {
```

src/main/java/core/basesyntax/ReturnMessage.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax;
+
+public interface ReturnMessage {
```
ChumakAn, line 3:
```suggestion
public interface Drawable {
```

src/main/java/core/basesyntax/ReturnMessage.java
```diff
@@ -0,0 +1,6 @@
...
+
+    void messageForReturn();
```
ChumakAn, line 5:
```suggestion
    void draw();
```

ChumakAn (changes requested):
good! Let's make it better

ChumakAn (changes requested):
you didn't resolve all my comments

src/main/java/core/basesyntax/ColorSupplier.java
Matolch: What did you mean? I try change it and did work/

PR #718, opened 2022-10-02
hw-oop-advanced

Iryna-a:


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,53 @@
...
+            int randomSecondLeg = random.nextInt(BOUND);
+            Figure triangle = new RightTriangle(randomFirstLeg, randomSecondLeg);
+            triangle.setColor(randomColor);
+            return triangle;
+        } else {
+            int randomSide = random.nextInt(BOUND);
+            Figure square = new Square(randomSide);
+            square.setColor(randomColor);
+            return square;
+        }
```
KostyaNikiforov, line 45: Can you use switch-case instead of if-blocks?

KostyaNikiforov, line 45: Also, i think we can improve the readability of this method by moving common logic to separate privet methods. 
For example, you can create separate methods for each type of figure.
Remember, your methods should not to be more than 15-20 lines long.

KostyaNikiforov (changes requested):
I like your solution! It is pretty clear and you fixed a lot of common mistakes:fire:

KostyaNikiforov (approved):
🔥

KostyaNikiforov (changes requested):
Hhh, don't forget to run `mvn clean package` before push

KostyaNikiforov (approved):

PR #719, opened 2022-10-02
Solved OOP Advanced task

HryhorashPavlo:


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,56 @@
...
+            case 2:
+                return getRandomRectangle();
+            case 3:
+                return getRandomTriangle();
+            case 4:
+                return getRandomTrapezoid();
+            default:
+                return getDefaultFigure();
+        }
+    }
```
KostyaNikiforov, line 31: Wow, it looks good)

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+public class Main {
+    public static final int FIGURE_AMOUNT = 10;
+
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] array = new Figure[FIGURE_AMOUNT];
```
KostyaNikiforov, line 8: Let's give this array a more personal name. I think it is not just any array, but the Array fo Figures. Don't make him offended)

KostyaNikiforov (changes requested):
Goot solution:fire:

KostyaNikiforov (approved):
🔥

PR #722, opened 2022-10-03
oop-advanced solution

Serhii-Prykhodko:
Created and implemented abstract class "Figure", 2 interface(Drawable and AreaCalculator), rename "HelloWorld" to "Main", and other.

### DIFF
--- a/src/main/java/core/basesyntax/Circle.java
+++ b/src/main/java/core/basesyntax/Circle.java
@@ -1,0 +1,21 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+    private int radius;
+
+    public Circle(int radius, String color) {
+        super(color);
+        this.radius = radius;
+    }
+
+    @Override
+    public double getArea() {
+        return (int) (Math.PI * radius * radius);
+    }
+
+    @Override
+    public void draw() {
+        System.out.println("Figure: Circle, area: " + getArea() + " sq.units, radius: " + radius
+                + " units, color: " + getColor());
+    }
+}
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private Random random;
+
+    public ColorSupplier(Random random) {
+        this.random = random;
+    }
+
+    public String getRandomColor() {
+        int colorNumber = new Random().nextInt(Color.values().length);
+        return (Color.values()[colorNumber]).name();
+    }
+}
--- a/src/main/java/core/basesyntax/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/FigureSupplier.java
@@ -1,0 +1,45 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int FIGURES_COUNT = 100;
+    private static final int CIRCLE_RADIUS = 10;
+    private static final String CIRCLE_COLOR = "WHITE";
+    private final ColorSupplier colorSupplier;
+    private Random random;
+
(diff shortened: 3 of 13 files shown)


src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,21 @@
...
+    private int radius;
+
+    public Circle(int radius, String color) {
+        super(color);
+        this.radius = radius;
+    }
+
+    @Override
+    public double getArea() {
+        return (int) (Math.PI * radius * radius);
```
ChumakAn, line 13: why are you casting to int if the return type is double?

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private Random random;
+
+    public ColorSupplier(Random random) {
+        this.random = random;
```
ChumakAn, line 9: it's better to initialize this variable in the same place where you declare this.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,45 @@
...
+    private static final int CIRCLE_RADIUS = 10;
+    private static final String CIRCLE_COLOR = "WHITE";
+    private final ColorSupplier colorSupplier;
+    private Random random;
+
+    public FigureSupplier(ColorSupplier colorSupplier, Random random) {
+        this.colorSupplier = colorSupplier;
+        this.random = random;
+    }
+
```
ChumakAn, line 16: it's better to initialize this variables with their declaration (line 12-15)

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,45 @@
...
+            case 2:
+                return new IsoscelesTrapezoid(getRandomValues(), getRandomValues(),
+                        getRandomValues(), colorSupplier.getRandomColor());
+            case 3:
+                return new Rectangle(getRandomValues(),getRandomValues(),
+                        colorSupplier.getRandomColor());
+            default:
+                return new Square(getRandomValues(), colorSupplier.getRandomColor());
+        }
+    }
```
ChumakAn, line 39: it's better to make cases more clear instead of random indexes. What about creating enum with different shapes and take a random one from there and go through switch-case depending on case

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,45 @@
...
+                return new RightTriangle(getRandomValues(), getRandomValues(),
+                        colorSupplier.getRandomColor());
+            case 2:
+                return new IsoscelesTrapezoid(getRandomValues(), getRandomValues(),
+                        getRandomValues(), colorSupplier.getRandomColor());
+            case 3:
+                return new Rectangle(getRandomValues(),getRandomValues(),
+                        colorSupplier.getRandomColor());
+            default:
+                return new Square(getRandomValues(), colorSupplier.getRandomColor());
```
ChumakAn, line 37: you should use getDefaultFigure() here too

ChumakAn (changes requested):
Good! Let's make the solution better!

src/main/java/core/basesyntax/FigureSupplier.java
ChumakAn: it's optional

src/main/java/core/basesyntax/ColorSupplier.java
ChumakAn: remove this constructor, we don't need this anymore

ChumakAn (changes requested):
One small comment

ChumakAn (approved):
Well done

Serhii-Prykhodko:
thanks))

PR #723, opened 2022-10-03
completed all task from README.md file

Danyl-Kniaziev:


src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,22 @@
...
+    @Override
+    public double getArea() {
+        return radius * radius * PI;
+    }
+
+    @Override
+    public String draw() {
+        return String.format(Pen.template, Shape.CIRCLE.name().toLowerCase(), getArea(),
+                String.format("radius: %d units", radius),
+                this.getColor());
```
KostyaNikiforov, line 20: Why do we need `this` ?

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public class Main {
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] arrayFigures = new Figure[6];
```
KostyaNikiforov, line 6:
```suggestion
        Figure[] figures = new Figure[6];
```

src/main/java/core/basesyntax/Pen.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax;
+
+public interface Pen {
+    String template = "Figure: %s, area: %.1f sq.units, %s, color: %s";
```
KostyaNikiforov, line 4:
```suggestion
    String TEMPLATE = "Figure: %s, area: %.1f sq.units, %s, color: %s";
```

src/main/java/core/basesyntax/Shape.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum Shape {
```
KostyaNikiforov, line 3:
```suggestion
public enum FegureName {
```

KostyaNikiforov (changes requested):
Good solution:fire:

src/main/java/core/basesyntax/Circle.java
Danyl-Kniaziev: Habit after Python😅

src/main/java/core/basesyntax/Pen.java
Danyl-Kniaziev: I think that we must use NAME_STYlE only for number constant

src/main/java/core/basesyntax/Shape.java
Danyl-Kniaziev: Why Shape is a bad name?

src/main/java/core/basesyntax/Circle.java
KostyaNikiforov: You learn bad things fast)

src/main/java/core/basesyntax/Pen.java
KostyaNikiforov: Why?)

Danyl-Kniaziev: *I thoutht.

src/main/java/core/basesyntax/Shape.java
KostyaNikiforov: Because you named elements the same way as you named classes. But they are figures and this is Shapes, why not vice versa? Or why don't you use the same word? I guess that is because it will look weird if an enum and an abstract class have the same name. 
Also, we know that the FegureName is a property of Figure, so the name i suggested is perfect for us.

KostyaNikiforov (approved):
🔥

PR #730, opened 2022-10-05
hw done

O1eksandrrr:
created array with 6 figures, 3 random and 3 default

### DIFF
--- a/src/main/java/Area.java
+++ b/src/main/java/Area.java
@@ -1,0 +1,3 @@
+public interface Area {
+    void drawFigure();
+}
--- a/src/main/java/Area.java
+++ b/src/main/java/Area.java
@@ -1,3 +1,3 @@
-public interface Area {
+public interface Drawable {
     void drawFigure();
 }
--- a/src/main/java/Draw.java
+++ b/src/main/java/Draw.java
@@ -1,3 +1,3 @@
-public interface Draw {
+public interface SquareCalculator {
     double getArea();
 }
--- a/src/main/java/Draw.java
+++ b/src/main/java/Draw.java
@@ -1,0 +1,3 @@
+public interface Draw {
+    double getArea();
+}
--- a/src/main/java/FigureSupplier.java
+++ b/src/main/java/FigureSupplier.java
@@ -1,0 +1,59 @@
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final double RADIUS = 10;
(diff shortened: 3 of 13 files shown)


src/main/java/Area.java
```diff
@@ -0,0 +1,3 @@
+public interface Area {
```
Rommelua, line 1:
```suggestion
public interface Drawable {
```

src/main/java/Area.java
```diff
@@ -0,0 +1,3 @@
...
+    void drawFigure();
```
Rommelua, line 2: Why drawFigure? Simple draw() - it is importent!
 What if tomorrow you will create class Plane implements Drawable ?

src/main/java/Draw.java
```diff
@@ -0,0 +1,3 @@
+public interface Draw {
```
Rommelua, line 1:
```suggestion
public interface SquareCalculator {
```

src/main/java/FigureSupplier.java
```diff
@@ -0,0 +1,59 @@
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final double RADIUS = 10;
+    private static final String COLOR = "WHITE";
```
Rommelua, line 5:
```suggestion
    private static final double DEFAULT_RADIUS = 10;
    private static final String DEFAULT_COLOR = "WHITE";
```

src/main/java/Main.java
```diff
@@ -0,0 +1,17 @@
+public class Main {
+    public static void main(String[] args) {
+        final int FN = 6;
```
Rommelua, line 3: What is FN?

src/main/java/Main.java
```diff
@@ -0,0 +1,17 @@
...
+        Figure []figures = new Figure[FN];
+        for (int i = 0; i < figures.length; i++) {
+            if (i < figures.length / 2) {
+                figures[i] = figureSupplier.getRandomFigure();
+                figures[i].drawFigure();
+            } else {
+                figures[i] = figureSupplier.getDefaultFigure();
+                figures[i].drawFigure();
+            }
+
```
Rommelua, line 14: empty line

Rommelua (changes requested):
image
Fix the build please.

aaananas (changes requested):
image

PR #732, opened 2022-10-05
fixed problems with maven version

O1eksandrrr:
fixed mistakes with interface names


aaananas (approved):
remember to create a separate branch for your hw

PR #737, opened 2022-10-06
Create random Figures

rozyzlav:


src/main/java/core/basesyntax/DiffColor.java
```diff
@@ -0,0 +1,14 @@
...
+    blue,
+    red,
+    green,
+    violet,
+    yellow,
+    black,
+    white,
+    brown,
+    golden,
+    grey,
```
ChumakAn, line 13: it's better to write with upper case in enum
etc BLUE, RED

ChumakAn (approved):
LGTM, but look at the comment

PR #740, opened 2022-10-09
Created required classes, interfaces etc. Were used in a class Main

BusyhinD:
I'm not sure about the implementation of the getDefaultFigure() method. Did not fully understand the required task.

### DIFF
--- a/src/main/java/core/basesyntax/AreaCalculator.java
+++ b/src/main/java/core/basesyntax/AreaCalculator.java
@@ -1,0 +1,5 @@
+package core.basesyntax;
+
+public interface AreaCalculator {
+    double getFigureArea();
+}
--- a/src/main/java/core/basesyntax/AreaCalculator.java
+++ b/src/main/java/core/basesyntax/AreaCalculator.java
@@ -1,5 +1,5 @@
 package core.basesyntax;
 
 public interface AreaCalculator {
-    double getFigureArea();
+    double getArea();
 }
--- a/src/main/java/core/basesyntax/Figure.java
+++ b/src/main/java/core/basesyntax/Figure.java
@@ -2,21 +2,13 @@ package core.basesyntax;
 
 public abstract class Figure implements AreaCalculator, InfoPrinter {
     private String figureName;
-    private String figureColor;
+    private String color;
 
-    public String getFigureName() {
-        return figureName;
+    public String getColor() {
+        return color;
     }
 
-    public void setFigureName(String figureName) {
-        this.figureName = figureName;
-    }
-
-    public String getFigureColor() {
-        return figureColor;
-    }
-
-    public void setFigureColor(String figureColor) {
-        this.figureColor = figureColor;
+    public void setColor(String color) {
+        this.color = color;
     }
 }
--- a/src/main/java/core/basesyntax/Figure.java
+++ b/src/main/java/core/basesyntax/Figure.java
@@ -1,0 +1,22 @@
+package core.basesyntax;
+
+public abstract class Figure implements AreaCalculator, InfoPrinter {
+    private String figureName;
+    private String figureColor;
+
+    public String getFigureName() {
+        return figureName;
+    }
+
+    public void setFigureName(String figureName) {
+        this.figureName = figureName;
+    }
+
+    public String getFigureColor() {
+        return figureColor;
+    }
+
+    public void setFigureColor(String figureColor) {
+        this.figureColor = figureColor;
+    }
+}
--- a/src/main/java/core/basesyntax/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/FigureSupplier.java
@@ -5,39 +5,57 @@ import java.util.Random;
 public class FigureSupplier {
     private static final int FIGURE_COUNT = 5;
     private static final int MAX_VALUE = 10;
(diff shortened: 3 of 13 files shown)


src/main/java/core/basesyntax/AreaCalculator.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface AreaCalculator {
+    double getFigureArea();
```
MaliukDaria, line 4:
```suggestion
    double getArea();
```

as we can use this interface not only for Figure class

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax;
+
+public abstract class Figure implements AreaCalculator, InfoPrinter {
+    private String figureName;
```
MaliukDaria, line 4: We can use `getClass().getSimpleName()` where needed, lets remove removeNode variable

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,22 @@
...
+    private String figureColor;
```
MaliukDaria, line 5:
```suggestion
    private String color;
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,43 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int FIGURE_COUNT = 5;
+    private static final int MAX_VALUE = 10;
+    private int[] figureParameters = new int[3];
```
MaliukDaria, line 8: can we implement logit without this array? (as we need only one parameter for Circle class)

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,43 @@
...
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    private Figure getFigure(int figureNumber) {
```
MaliukDaria, line 12: private method should follow after public as helper ones

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,43 @@
...
+        }
+        Figure figure = getFigure(random.nextInt(FIGURE_COUNT));
+        figure.setFigureColor(colorSupplier.getRandomColor());
+        return figure;
+    }
+
+    public Figure getDefaultFigure() {
+        figureParameters[0] = 10;
+        Figure figure = getFigure(0);
+        figure.setFigureColor(Color.WHITE.name().toLowerCase());
```
MaliukDaria, line 40: leta make number 10 and Color.WHITE as constant fields

src/main/java/core/basesyntax/InfoPrinter.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface InfoPrinter {
+    void getInfo();
```
MaliukDaria, line 4:
```suggestion
    void drawInfo();
```

this method is void, so we are not getting anything from it

MaliukDaria (changes requested):

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int FIGURE_COUNT = 5;
+    private static final int MAX_VALUE = 10;
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getDefaultFigure() {
+        final int figureParameter = 10;
```
zachar-chorny, line 12:
```suggestion
        final int figureParameter = 10;
```
This is not a constant but a variable in the constructor. Fix it and change the name to CIRCLE_AREA

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,61 @@
...
+                return getRandomRectangle();
+            case 3:
+                return getRandomIsoscelesTrapezoid();
+            default:
+                return getRandomIsoscelesRightTriangle();
+        }
+    }
+
+    private Figure getRandomCircle() {
+        return new Circle(random.nextInt(MAX_VALUE + 1));
```
zachar-chorny, line 41: If you want the maximum value to 11, specify MAX_VALUE as 11 instead of adding it everywhere. Fix everywhere

zachar-chorny (changes requested):

src/main/java/core/basesyntax/FigureSupplier.java
kateryna-deshchinska: You didn't fix it

kateryna-deshchinska (approved):
Fix issues that Zachar suggested

src/main/java/core/basesyntax/FigureSupplier.java
BusyhinD: If I do as suggested, the nextInt method might return 0. Does that make sense?

PR #748, opened 2022-10-23
Oop advanced solution

vgladkyi:


src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public abstract class Figure implements Drawable, AreaCalculator {
+    protected final String color;
+
+    public Figure(String color) {
+        this.color = color;
+    }
+
+    public abstract double getArea();
```
Rommelua, line 10:
```suggestion

```
[Common mistakes](https://mate-academy.github.io/jv-program-common-mistakes/java-core/abstract-class-interface/oop-advanced)
Don’t use abstract classes to set behaviour for classes
Abstract classes and interfaces have different use cases. Try to figure out when to use both in this task by yourself. If you’re blocked [this](https://stackoverflow.com/a/479168) may give you a hint.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.Random;
+
+public class FigureSupplier {
+    public static final int FIGURE_COUNT = 5;
+    public static final int MAX_UNITS = 50;
+    private final Random random = new Random();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getDefaultFigure() {
+        return new Circle("WHITE", 10);
```
Rommelua, line 12: Сreate constants DEFAULT_COLOR and DEFAULT_RADIUS

Rommelua (changes requested):

src/main/java/core/basesyntax/Figure.java
vgladkyi: So I need to remove "abstract" modificator from getArea method, create default method body and than override method in classes that use it ? 
```suggestion

    public double getArea(){
       return 0;
    }
```

vgladkyi: This method in Figure file is unnecessary, because I used it when I was building main logic. When I opened file with common mistakes, I created an interface with method getArea.

zachar-chorny (approved):

PR #774, opened 2022-11-11
Created Classes for Every figure.

Bogdan-Pokhilko:
Created an abstract class Figure (I could also use interface Class for getArea, toDraw. But I realized that all the figures will use those methods, so I stick to the abstract class)

ColorSupplier and FigureSupplier classes were created. In FigureSuplier I've created all the static valuables that were required, used switch to randomize Figures.

in Main I've created an array of Figures and manually declared positions in an array (I could've used for loop for that, if needed). and with for loop I've executed toDraw method.

### DIFF
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private Random random = new Random();
+
+    public String getRandomColor() {
+        Color [] value = Color.values();
+        return value[random.nextInt(value.length)].toString();
+    }
+}
--- a/src/main/java/core/basesyntax/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/FigureSupplier.java
@@ -1,0 +1,38 @@
+package core.basesyntax;
+
+import java.util.Random;
+public class FigureSupplier {
+    private static final String DEFAULT_COLOR = "WHITE";
+    private static final int DEFAULT_RADIUS = 10;
+      private static final int MAX_NUMBER = 6;
+    private Random random = new Random();
+    private  ColorSupplier randomColor = new ColorSupplier();
+
+    public Figure getRandomFigure(int value) {
+        int randomNum = random.nextInt(MAX_NUMBER);
+
+        switch (value) {
+            case 1:
+                return new Circle(randomColor.getRandomColor(), random.nextInt(randomNum));
+
+            case 2:
+                return new Rectangle(randomColor.getRandomColor(), random.nextInt(randomNum), random.nextInt(randomNum));
+
+            case 3:
+                return new Square(randomColor.getRandomColor(), random.nextInt(randomNum));
+
+            case 4:
+                return new RightTriangle(randomColor.getRandomColor(), random.nextInt(randomNum), random.nextInt(randomNum));
+
+            case 5:
+                return new IsoscelesTrapezoid(randomColor.getRandomColor(), random.nextInt(randomNum), random.nextInt(randomNum), random.nextInt(randomNum));
+
+            default:
+                return new Square(DEFAULT_COLOR, MAX_NUMBER);
+        }
+    }
+
+    public Figure getDefaultFigure () {
+        return new Circle(DEFAULT_COLOR, DEFAULT_RADIUS);
+    }
+}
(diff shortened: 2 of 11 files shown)


aaananas (changes requested):
image
run `mvn clean package`

Bogdan-Pokhilko:
> Can you, please, help me. I can't understand the type of error that mvn is giving me :( 

ex

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private Random random = new Random();
+
+    public String getRandomColor() {
+        Color [] value = Color.values();
+        return value[random.nextInt(value.length)].toString();
```
aaananas, line 10: image

src/main/java/core/basesyntax/Drawable.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Drawable {
+    void toDraw();
```
aaananas, line 4:
```suggestion
    void draw();
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,54 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+
```
aaananas, line 6:
```suggestion
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,54 @@
...
+    private static final String DEFAULT_COLOR = "WHITE";
+    private static final int DEFAULT_RADIUS = 10;
+    private static final int MAX_NUMBER = 10;
+    private Random random = new Random();
+    private ColorSupplier randomColor = new ColorSupplier();
+    private int randomNum = random.nextInt(MAX_NUMBER);
```
aaananas, line 12: this should be local variable

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,54 @@
...
+
+    private Figure getRandomCircle() {
+        return new Circle(randomColor.getRandomColor(), random.nextInt(randomNum));
+    }
+
+    private Figure getRandomTrapezoid() {
+        return new IsoscelesTrapezoid(randomColor.getRandomColor(),
+                random.nextInt(randomNum),
+                random.nextInt(randomNum), random.nextInt(randomNum));
+    }
```
aaananas, line 31: private methods should go after public methods

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,54 @@
...
+
+    public Figure getRandomFigure() {
+        int randomNumber = random.nextInt(3);
```
aaananas, line 34:
```suggestion
        int randomNumber = random.nextInt(MAX_NUMBER);
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,54 @@
...
+                return getRandomCircle();
+
+            case 2:
+                return getRandomTriangle();
+
+            case 3:
+                return getRandomSquare();
+
+            default:
+                return getRandomTrapezoid();
```
aaananas, line 47: you should start with 0

aaananas (changes requested):

src/main/java/core/basesyntax/FigureSupplier.java
Bogdan-Pokhilko: I do not quite understand it. Local var are vars declared inside of the methods, but random, randomNum or randomColor being used in different methods isn't it better to declare them as Class var?  Or I understood the reference wrong, sorry for that😊

aaananas: Dont write comments here - because mentors cant see them until your next re-request

aaananas: Random and ColorSupplier - should be class-level
randomNum should be generated each time and you need that new value there - there is no need to have a class level variable. Also that randomNumber has nothing to do with FigureSupplier class

aaananas (changes requested):
requested changes not fixed

aaananas (approved):

PR #798, opened 2022-11-28
Figures, Suppliers were created

katrevichdv:


src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+    private static final String FORM = "Figure: circle, area: %.1f sq.units, radius: %d units"
+            + ", color: %s";
```
zachar-chorny, line 5: You don't need to store this string as a constant, just write it in toString(). Fix everywhere

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private static final Colors[] COLORS = Colors.values();
+    private static final int COLORS_COUNT = COLORS.length;
```
zachar-chorny, line 7: These values should not be made constants

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,20 @@
...
+    private Random random = new Random();
+
+    private enum Colors {
+        BLACK,
+        RED,
+        WHITE,
+        YELLOW
+    }
```
zachar-chorny, line 15:
```suggestion
    private enum Color {
        BLACK,
        RED,
        WHITE,
        YELLOW
    }
```
This enum should not be in the CollorSupplier, create it separately

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,60 @@
...
+    private static final int DEFAULT_RADIUS = 10;
+    private static final int MIN_SIZE = 10;
+    private static final int MAX_SIZE = 20;
+    private static final int DIFF = MAX_SIZE - MIN_SIZE;
+    private static final String DEFAULT_COLOR = "white";
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        switch (random.nextInt(5)) {
```
zachar-chorny, line 15: 5 must be a constant

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,60 @@
...
+                return getRandomSquare();
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(DEFAULT_COLOR, DEFAULT_RADIUS);
+    }
+
+    private Figure getRandomCircle() {
+        return new Circle(colorSupplier.getRandomColor(), random.nextInt(DIFF) + MIN_SIZE);
```
zachar-chorny, line 35: Why make it so complicated and add 10 to a random number every time? Just pass the maximum desired number to nextInt at once. And delete DIFF & MIN_SIZE

katrevichdv: I just wanted to make sure that trapezoid area would never be NaN. I'll fix it.

src/main/java/core/basesyntax/Circle.java
katrevichdv: >  (you shouldn't override toString() method for this)

 So I moved them to draw() methods.

aaananas (approved):

PR #822, opened 2022-12-11
implemented task

Olegator191:


src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public abstract class Figure implements Draw, AreaCounter {
+    private String color;
```
KostyaNikiforov, line 4: Can u use the enum here? It is better to avoid primitives if it is possible

KostyaNikiforov (changes requested):
Good solution!:fire:

src/main/java/core/basesyntax/Figure.java
Olegator191: Yes,it`s done!)

src/main/java/core/basesyntax/Figure.java
```diff
@@ -1,13 +1,13 @@
...
 public abstract class Figure implements Draw, AreaCounter {
-    private String color;
+    private Color color;
 
     public Figure(String color) {
-        this.color = color;
+        this.color = Color.valueOf(color);
     }
 
     public String getColor() {
```
KostyaNikiforov, line 10: Why don't you use Enum here too?

KostyaNikiforov (changes requested):

PR #823, opened 2022-12-11
tasks were implemented

lysytskapolina:


src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,18 @@
+package core.basesyntax;
+
+public abstract class Figure implements Area, Drawing {
+    private String color;
+
+    public Figure() {
+        ColorSupplier colorSupplier = new ColorSupplier();
+        this.color = colorSupplier.getRandomColor();
```
KostyaNikiforov, line 8: A constructor must not have any business logic. It has only one responsibility - Object Initialization 
Please, pass all needed fields to parameters ;)

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,27 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+    private final int radius;
+
+    public Circle(int radius) {
+        super();
```
KostyaNikiforov, line 7: It is not required to call a parental constructor if you don't pass any params

KostyaNikiforov (changes requested):
Good job:fire:

lysytskapolina:
Requested changes are done, but I don't know if i could do it easier way?

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+    private final int radius;
+
+    public Circle(String color, int radius) {
+        super.setColor(color);
```
KostyaNikiforov, line 7:
```suggestion
        super(color);
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,39 @@
...
+        colorSupplier = new ColorSupplier();
+    }
+
+    private int getSide() {
+        return random.nextInt(MAX_SIDE);
+    }
+
+    private String getColor() {
+        return colorSupplier.getRandomColor();
+    }
```
KostyaNikiforov, line 23: It is better to place private method at the end of the class

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,39 @@
...
+    public FigureSupplier() {
+        random = new Random();
+        colorSupplier = new ColorSupplier();
+    }
+
+    private int getSide() {
+        return random.nextInt(MAX_SIDE);
+    }
+
+    private String getColor() {
```
KostyaNikiforov, line 21: Which color?) Let's make you method name more informative)

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public abstract class Figure implements Area, Drawing {
+    private String color;
+
```
KostyaNikiforov, line 5: Create a constructor here, which will receive a color as a parameter

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public abstract class Figure implements Area, Drawing {
+    private String color;
```
KostyaNikiforov, line 4: Also, let's use enum here. It is better to avoid primitives if it is possible

KostyaNikiforov (changes requested):

PR #830, opened 2022-12-12
oop advanced task completed

OleksiiZaloznyi:


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,19 @@
...
+        Figure[] figures = new Figure[FIGURES_AMOUNT];
+        for (int i = 0; i < figures.length / 2; i++) {
+            figures[i] = figureSupplier.getRandomFigure();
+        }
+        for (int i = figures.length / 2; i < figures.length; i++) {
+            figures[i] = figureSupplier.getDefaultFigure();
+        }
+        for (int i = 0; i < figures.length; i++) {
+            figures[i].draw();
+        }
```
KostyaNikiforov, line 17: Let's use for-each loop. We don't need an index here

KostyaNikiforov (changes requested):
Cool solution!:fire:

KostyaNikiforov (approved):

PR #868, opened 2023-01-07
implemented the task, created classes of figures, one abstract class …

0-zagar-0:
…and two interfaces and implemented methods

### DIFF
--- a/src/main/java/core/basesyntax/Main.java
+++ b/src/main/java/core/basesyntax/Main.java
@@ -1,0 +1,32 @@
+package core.basesyntax;
+
+import core.basesyntax.figure.Figure;
+import core.basesyntax.figure.FigureSupplier;
+import core.basesyntax.figure.name.RandomName;
+
+public class Main {
+    private static final int ARR_LENGTH = 6;
+
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        RandomName randomName = new RandomName();
+        Figure[] figures = new Figure[ARR_LENGTH];
+
+        for (int i = 0; i < ARR_LENGTH; i++) {
+            if (i < ARR_LENGTH / 2) {
+                String nameFigure = randomName.getRandomName();
+                Figure figure = figureSupplier.getRandomFigure(nameFigure);
+                figures[i] = figure;
+            }
+
+            if (i >= ARR_LENGTH / 2) {
+                Figure figure = figureSupplier.getDefaultFigure();
+                figures[i] = figure;
+            }
+        }
+
+        for (Figure figure : figures) {
+            figure.draw();
+        }
+    }
+}
--- a/src/main/java/core/basesyntax/figure/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/figure/FigureSupplier.java
@@ -1,0 +1,68 @@
+package core.basesyntax.figure;
+
+import core.basesyntax.figure.color.ColorSupplier;
+import core.basesyntax.figure.name.RandomName;
+import java.util.Random;
+
+public class FigureSupplier {
+    //Constant
+    private static final int BOUND_NUMBER = 10;
+    private static final int RADIUS_CIRCLE = 10;
+    //Object
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private RandomName randomName = new RandomName();
+    private Random random = new Random();
+    //Random variables
+    private String color = colorSupplier.getRandomColor();
+    private String nameFigure = randomName.getRandomName();
+
+    public Figure getCircle() {
+        int radius = random.nextInt(BOUND_NUMBER);
+        return new Circle(color, radius);
+    }
+
+    public Figure getSquare() {
+        int side = random.nextInt(BOUND_NUMBER);
+        return new Square(color, side);
+    }
+
+    public Figure getRightTriangle() {
+        int firstLeg = random.nextInt(BOUND_NUMBER);
+        int secondLeg = random.nextInt(BOUND_NUMBER);
+        return new RightTriangle(color, firstLeg, secondLeg);
+    }
+
+    public Figure getRectangle() {
+        int sideA = random.nextInt(BOUND_NUMBER);
+        int sideB = random.nextInt(BOUND_NUMBER);
+        return new Rectangle(color, sideA, sideB);
+    }
+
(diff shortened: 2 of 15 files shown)


src/main/java/core/basesyntax/figure/FigureSupplier.java
```diff
@@ -0,0 +1,68 @@
+package core.basesyntax.figure;
+
+import core.basesyntax.figure.color.ColorSupplier;
+import core.basesyntax.figure.name.RandomName;
+import java.util.Random;
+
+public class FigureSupplier {
+    //Constant
```
nacenik, line 8: Don't add this comment to your code. Fix it in all places

src/main/java/core/basesyntax/figure/FigureSupplier.java
```diff
@@ -0,0 +1,68 @@
...
+public class FigureSupplier {
+    //Constant
+    private static final int BOUND_NUMBER = 10;
+    private static final int RADIUS_CIRCLE = 10;
+    //Object
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private RandomName randomName = new RandomName();
+    private Random random = new Random();
+    //Random variables
+    private String color = colorSupplier.getRandomColor();
```
nacenik, line 16: Create this variable in method `getRandomFigure();`

src/main/java/core/basesyntax/figure/FigureSupplier.java
```diff
@@ -0,0 +1,68 @@
...
+    private String nameFigure = randomName.getRandomName();
```
nacenik, line 17: You don't need this variable, cause you have a figure name in the method `getRandomFigure()` and you can use that value

nacenik (changes requested):

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,30 @@
...
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        FigureNameSupplier figureNameSupplier = new FigureNameSupplier();
+        Figure[] figures = new Figure[ARR_LENGTH];
+
+        for (int i = 0; i < ARR_LENGTH; i++) {
+            if (i < ARR_LENGTH / 2) {
+                String nameFigure = figureNameSupplier.getRandomName();
+                Figure figure = figureSupplier.getRandomFigure(nameFigure);
+                figures[i] = figure;
```
okuzan, line 19: common part can be extracted from if

src/main/java/core/basesyntax/figure/interfaces/DrawFigure.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax.figure.interfaces;
+
+public interface DrawFigure {
```
okuzan, line 3:
```suggestion
public interface Drawable {

```

okuzan (changes requested):
Good job!

src/main/java/core/basesyntax/figure/interfaces/DrawFigure.java
0-zagar-0: Why don't you write about such errors in names right away?

src/main/java/core/basesyntax/figure/FigureSupplier.java
```diff
@@ -0,0 +1,65 @@
...
+public class FigureSupplier {
+    private static final int BOUND_NUMBER = 10;
+    private static final int DEFAULT_RADIUS = 10;
+    private static final String DEFAULT_COLOR = "white";
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private FigureNameSupplier figureNameSupplier = new FigureNameSupplier();
+    private Random random = new Random();
+    private String color;
+
+    private Figure getCircle() {
```
aleksandr-bevza, line 16: public methods should go first. Private methods should go after public methods

src/main/java/core/basesyntax/figure/FigureSupplier.java
```diff
@@ -0,0 +1,65 @@
...
+        int upperSide = random.nextInt(BOUND_NUMBER);
+        int downSide = random.nextInt(BOUND_NUMBER);
+        int height = random.nextInt(BOUND_NUMBER);
+        return new IsoscelesTrapezoid(color, upperSide, downSide, height);
+    }
+
+    public Figure getRandomFigure() {
+        color = colorSupplier.getRandomColor();
+        switch (figureNameSupplier.getRandomName()) {
+            case "SQUARE":
```
aleksandr-bevza, line 48: switch (figureNameSupplier.getRandomName()) gives us the value of enum. 
It is better to compare with values of the same enum in "case"
```suggestion
        switch (figureNameSupplier.getRandomName()) {
            case SQUARE:
```

src/main/java/core/basesyntax/figure/FigureSupplier.java
```diff
@@ -0,0 +1,65 @@
+package core.basesyntax.figure;
+
+import core.basesyntax.figure.color.ColorSupplier;
+import core.basesyntax.figure.name.FigureNameSupplier;
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int BOUND_NUMBER = 10;
+    private static final int DEFAULT_RADIUS = 10;
+    private static final String DEFAULT_COLOR = "white";
```
aleksandr-bevza, line 10: It's better not to write "white" here. Get it from enum.

src/main/java/core/basesyntax/figure/name/FigureNameSupplier.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax.figure.name;
+
+import java.util.Random;
+
+public class FigureNameSupplier {
+    private Random random = new Random();
+
+    public String getRandomName() {
+        int index = random.nextInt(FigureName.values().length);
+        return FigureName.values()[index].name();
```
aleksandr-bevza, line 10: This method is used only in switch - case. It's better to use enums in switch - case if possible
```suggestion
    public FigureName getRandomName() {
        int index = random.nextInt(FigureName.values().length);
        return FigureName.values()[index];
```

aleksandr-bevza (changes requested):

aleksandr-bevza (approved):

PR #874, opened 2023-01-13
Update HelloWorld.java

kromch:


Color.java
```diff
@@ -0,0 +1,32 @@
+package core.basesyntax;
+
+public enum Color {
```
Andriy-dz, line 3: divide your entire solution into separate classes

Andriy-dz (changes requested):

PR #880, opened 2023-01-18
Hw jv oop advanced

Oleksii-Velykyi:


Ivan95kos (changes requested):
Pay attention to description of task and common mistakes

PR #884, opened 2023-01-23
Hw 4 solution

serhii-shyian:


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,67 @@
...
+
+public class FigureSupplier {
+    public static final int FIGURE_COUNT = 5;
+    public static final int MAX_NUMBER = 100;
+    public static final Color DEFAULT_COLOR = Color.WHITE;
+    public static final int DEFAULT_RADIUS = 10;
+    private final Random random = new Random();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    private Circle getRandomCircle() {
```
andriy-shulepa, line 13: private methods should be below public ones

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private final Random random = new Random();
+
+    public Color getRandomColor() {
+        int index = random.nextInt(Color.values().length);
```
andriy-shulepa, line 9: Color.values() each time creates a new array instance. As an optimization, you can extract length to static variable

andriy-shulepa (changes requested):

src/main/java/core/basesyntax/Draw.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Draw {
+    void printInfo();
```
okuzan, line 4:
```suggestion
    void draw();
```

src/main/java/core/basesyntax/Draw.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Draw {
```
okuzan, line 3:
```suggestion
public interface Drawable {
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,20 @@
...
+            if (i < figures.length / 2) {
+                figures[i] = figureSupplier.getRandomFigure();
+            } else {
+                figures[i] = figureSupplier.getDefaultFigure();
+            }
+        }
+
+        for (Figure figure : figures) {
+            figure.printInfo();
+        }
```
okuzan, line 18: don't need to iterate twice, call this method outside of if-else statement inside of the first if

okuzan (changes requested):

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,67 @@
...
+        switch (index) {
+            case 0:
+                return getRandomCircle();
+            case 1:
+                return getRandomSquare();
+            case 2:
+                return getRandomRectangle();
+            case 3:
+                return getRandomIsoscelesTrapezoid();
+            case 4:
```
Ivan95kos, line 24: case 4: does not return any figure, will be called by default,
so you can delete it
```suggestion
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+public class Main {
+    public static void main(String[] args) {
+        Figure [] figures = new Figure[10];
```
Ivan95kos, line 5: For the future, all magic numbers, strings, symbols in your code should be constants.

Ivan95kos (approved):
Great job

PR #891, opened 2023-01-30
jv-oop-advanced solution

potrap:


src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,29 @@
+package core.basesyntax;
+
+public abstract class Figure implements Printable {
+    private String figureType;
```
karma-o, line 4: why do we need this field?

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,29 @@
...
+    public Color getColor() {
+        return color;
+    }
+
+    public void setColor(Color color) {
+        this.color = color;
+    }
+
+    abstract double countArea();
+
```
karma-o, line 28: we need 1 more interface for this method
And there is a redundant empty line

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private static final Random random = new Random();
```
karma-o, line 6: let's not make it static, because it does not quite fit in the dependency injection principle, for now, we will initialize similar objects like this, and only constants with types of primitives or wrappers will be static

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,28 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+    private int radius;
+
+    {
+        setFigureType("circle");
+    }
```
karma-o, line 8: first of all we do not need this filed, and the second thing is that static initialization blocks are not recommended to use

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,63 @@
...
+            case 2:
+                return getRandomIsoscelesTrapezoid();
+            case 3:
+                return getRandomRectangle();
+            case 4:
+                return getRandomRightTriangle();
+            case 5:
+                return getRandomSquare();
+            default:
+                return null;
```
karma-o, line 26: it is recommended to not return null in the default case, better to make case 5 default instead

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,63 @@
...
+    private static final Color DEFAULT_COLOR = Color.WHITE;
+    private static final int DEFAULT_RADIUS = 10;
+    private static final int NUMBER_OF_FIGURES = 5;
+    private static final int MAX_VALUE = 20;
+    private final Random random = new Random();
+
+    public Figure getRandomFigure() {
+        int randomFigureNumber = random.nextInt(NUMBER_OF_FIGURES) + 1;
+        switch (randomFigureNumber) {
+            case 1:
```
karma-o, line 15: we can avoid incrementing it, and just start with `case 0` instead

src/main/java/core/basesyntax/HelloWorld.java
```diff
@@ -1,8 +1,14 @@
...
-
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] figures = new Figure[]{figureSupplier.getDefaultFigure(),
+                figureSupplier.getDefaultFigure(),
+                figureSupplier.getDefaultFigure(), figureSupplier.getRandomFigure(),
+                figureSupplier.getRandomFigure(), figureSupplier.getRandomFigure()};
+        for (Figure figure : figures) {
+            figure.printInfo();
+        }
```
karma-o, line 12: can we fill our array in a loop instead of initializing it like this?

src/main/java/core/basesyntax/IsoscelesTrapezoid.java
```diff
@@ -0,0 +1,37 @@
...
+
+    @Override
+    public void printInfo() {
+        System.out.println("Figure: "
+                + getFigureType() + ", area: "
+                + countArea() + " sq.units, topParallelSide: "
+                + topParallelSide + " units, lowerParallelSide: "
+                + lowerParallelSide + " units, nonParallelSide: "
+                + nonParallelSide + " units, color: "
+                + getColor().toString().toLowerCase());
```
karma-o, line 35: Don't you think it is confusing to have different parameters in one code line? let's make it like
```
+ "area: " + getArea() + "sq.units"
+ ", someProperty: " + getSomeProperty()
```

karma-o (changes requested):

src/main/java/core/basesyntax/HelloWorld.java
```diff
@@ -1,8 +1,15 @@
...
-/**
- * Feel free to remove this class and create your own.
- */
 public class HelloWorld {
-
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] figures = new Figure[6];
+        for (int i = 0; i < figures.length; i++) {
+            figures[i] = i < 2 ? figureSupplier.getDefaultFigure()
```
KseniiaMakarova, line 8:
```suggestion
            figures[i] = i < figures.length / 2 ? figureSupplier.getDefaultFigure()
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+
+    public static Color getRandomColor() {
```
KseniiaMakarova, line 7: please let's not make this method static

KseniiaMakarova (changes requested):

karma-o (approved):
LGTM :+1:

PR #899, opened 2023-02-05
solved the task using OOP principles

tanneken:


src/main/java/core/basesyntax/Application.java
```diff
@@ -0,0 +1,23 @@
...
+            if (i < HALF_OF_FIGURES) {
+                figures[i] = figureSupplier.getRandomFigure();
+            } else {
+                figures[i] = figureSupplier.getDefaultFigure();
+            }
+        }
+
+        for (Figure figure: figures) {
+            figure.drawFigure();
+        }
```
okuzan, line 21: you can call it in the first loop

src/main/java/core/basesyntax/FigureDrawer.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface FigureDrawer {
+    void drawFigure();
```
okuzan, line 4: only figures can be drawn here
Let's call it Drawable, and method `draw()`

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,32 @@
+package core.basesyntax;
+
+public class FigureSupplier {
+    public static final int FIGURE_COUNT = 5;
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final Random random = new Random();
+
+    public Figure getRandomFigure() {
+        int figureNumber = new java.util.Random().nextInt(FIGURE_COUNT);
```
okuzan, line 9:
```suggestion
        int figureNumber = random.nextInt(FIGURE_COUNT);
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,32 @@
...
+        switch (figureNumber) {
+            case 1:
+                return new Square(colorSupplier.getRandomColor(), random.getProperty());
```
okuzan, line 12: create dedicated private methods for creation of random figure of each type

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,32 @@
...
+            case 5:
+                return new IsoscelesTrapezoid(colorSupplier.getRandomColor(),
+                        random.getProperty(), random.getProperty(), random.getProperty());
+            default:
+                return new Circle(Color.WHITE.name(), 10);
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(Color.WHITE.name(), 10);
```
okuzan, line 30: move WHITE, 10 to constants

src/main/java/core/basesyntax/Random.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public class Random {
```
okuzan, line 3: I don't think we need this class, we can use java.util.Random created once as a field in FigureSupplier. But if you have this class, avoid recreating java.util.Random instance per each method invocation, it should have been a field

okuzan (changes requested):

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,58 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int FIGURE_COUNT = 5;
+    private static final String DEFAULT_COLOR = Color.WHITE.name();
+    private static final int DEFAULT_RADIUS = 10;
```
okuzan, line 8: place fields of the same type together - all primitives then objects etc

okuzan (approved):
Good job!

PR #902, opened 2023-02-06
created classes of figures with their additional functionality

denys-mg:


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,79 @@
...
+        } else {
+            longerBase = baseSide2;
+            shorterBase = baseSide1;
+        }
+
+        return new IsoscelesTrapezoid(colorSupplier.getRandomColor(),
+                longerBase, shorterBase, height);
+    }
+
+    public Figure getRandomFigure() {
```
maxline, line 59: method ordering. public methods should be higher then private ones. Check please other classes too.

maxline (changes requested):
Good job! Left a several comments. Please take a look.

src/main/java/core/basesyntax/AreaCalculator.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface AreaCalculator {
+    double getArea();
```
petro-ushchuk, line 4: Better name `calculateArea` since you have calculations inside this method.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,88 @@
...
+                return randomRightTriangle();
+            case ISOSCELES_TRAPEZOID:
+                return randomIsoscelesTrapezoid();
+            default:
+                return randomCircle();
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(Color.WHITE, 10);
```
petro-ushchuk, line 36: `10` should be constant

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax;
+
+public class Main {
+    public static void main(String[] args) {
+        final int figureCount = 6;
```
petro-ushchuk, line 5: 6 should be constant

petro-ushchuk (changes requested):
Good work, just fix these small issues!

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+public abstract class Figure implements Drawable, AreaCalculator {
+    protected static final int DECIMAL_FORMAT = 100;
+    protected Color color;
```
karma-o, line 5: It can be protected but then it will also be accesibble from other classes in the same package, so we need to be mindful of that, and anyways we need getters and seters for that field

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,89 @@
...
+    private static final double MIN_VALUE = 1;
+    private static final int DECIMAL_FORMAT = 100;
+    private static final double DEFAULT_CIRCLE_RADIUS = 10;
+
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        int randomFigureNumber = random.nextInt(FigureShape.values().length);
+
```
karma-o, line 21: redundant empty line

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,89 @@
...
+        double baseSide2 = Math.round((MAX_VALUE * random.nextDouble() + MIN_VALUE)
+                * DECIMAL_FORMAT) / DECIMAL_FORMAT;
+
+        if (baseSide1 > baseSide2) {
+            longerBase = baseSide1;
+            shorterBase = baseSide2;
+        } else {
+            longerBase = baseSide2;
+            shorterBase = baseSide1;
+        }
```
karma-o, line 84: I don't see any reasonable difference except for fields naming in the model to make it that complicated, we can get rid of all this just by changing the name to something more neutral, like `BaseA` and `BaseB` for e.g.

src/main/java/core/basesyntax/figuretypes/Circle.java
```diff
@@ -0,0 +1,27 @@
+package core.basesyntax.figuretypes;
```
karma-o, line 1:
```suggestion
package core.basesyntax.model;
```

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
```
karma-o, line 1:
```suggestion
package core.basesyntax.model;
```

karma-o (changes requested):

sarakhmen (approved):
Good solution :)

PR #910, opened 2023-02-08
created figure classes and figures array in main

outref:


src/main/java/core/basesyntax/AreaCalculator.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface AreaCalculator {
+    double area();
```
KseniiaMakarova, line 4:
```suggestion
    double getArea();
```

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,21 @@
...
+
+    public Circle(String color, double radius) {
+        super("Circle", color);
+        this.radius = radius;
+    }
+
+    @Override
+    public String draw() {
+        return String.format("Figure: %s, area: %.1f sq.units, radius: %.0f units, color: %s",
+                getType(),area(),radius,getColor());
```
KseniiaMakarova, line 14: add whitespaces
same for other figures

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,21 @@
+package core.basesyntax;
+
+public class Circle extends AbstractFigure {
+    private double radius;
+
+    public Circle(String color, double radius) {
+        super("Circle", color);
```
KseniiaMakarova, line 7: can we use enum value here? same for other figures

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,40 @@
...
+            default:
+                return getDefaultFigure();
+        }
+    }
+
+    public AbstractFigure getDefaultFigure() {
+        return new Circle("white", 10);
+    }
+
+    private double randomDouble() {
```
KseniiaMakarova, line 37:
```suggestion
    private double getRandomDouble() {
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,40 @@
...
+            case TRAPEZOID:
+                return new IsoscelesTrapezoid(colorSupplier.getRandomColor(), randomDouble(),
+                        randomDouble(), randomDouble());
+            default:
+                return getDefaultFigure();
+        }
+    }
+
+    public AbstractFigure getDefaultFigure() {
+        return new Circle("white", 10);
```
KseniiaMakarova, line 34: use enum

src/main/java/core/basesyntax/IsoscelesTrapezoid.java
```diff
@@ -0,0 +1,26 @@
+package core.basesyntax;
+
+public class IsoscelesTrapezoid extends AbstractFigure {
+    private double baseA;
+    private double baseB;
```
KseniiaMakarova, line 5:
```suggestion
    private double firstBase;
    private double secondBase;
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax;
+
+public class Main {
+    private static final int FIGURES_NUMBER = 6;
+    
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+
+        Drawable[] figures = new Drawable[FIGURES_NUMBER];
```
KseniiaMakarova, line 9:
```suggestion
        AbstractFigure[] figures = new AbstractFigure[FIGURES_NUMBER];
```

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -0,0 +1,24 @@
+package core.basesyntax;
+
+public class Rectangle extends AbstractFigure {
+    private double sideA;
+    private double sideB;
```
KseniiaMakarova, line 5: change to length and width

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/AbstractFigure.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax;
+
+public abstract class AbstractFigure implements Drawable, AreaCalculator {
+    private String type;
```
karma-o, line 4: I don't think we need this field actually, enum is used only for random figure generation pretty much, and classes implementations of the `AbstractFigure` are defined by the class name itself, and it is string, which is not ok if we have an enum

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,21 @@
...
+
+    @Override
+    public String draw() {
+        return String.format("Figure: %s, area: %.1f sq.units, radius: %.0f units, color: %s",
+                getType(), getArea(), radius, getColor());
+    }
+
+    @Override
+    public double getArea() {
+        return Math.PI * radius * radius;
```
karma-o, line 19: not a mistake but check ou `Math.pow(int, int)` method

src/main/java/core/basesyntax/AbstractFigure.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax;
+
+public abstract class AbstractFigure implements Drawable, AreaCalculator {
+    private String type;
+    private String color;
```
karma-o, line 5: since we have an enum `Color` let's use it here as a type, it is always preferred to avoid raw types like string and use their object representation instead

karma-o (changes requested):

Andriy-dz (approved):

PR #914, opened 2023-02-08
classes were created, methods(getRandomFigure(), getDefaultFigure()) …

oleksiistepaniak:
…were created

### DIFF
--- a/src/main/java/core/basesyntax/AreaCalculator.java
+++ b/src/main/java/core/basesyntax/AreaCalculator.java
@@ -1,5 +1,5 @@
 package core.basesyntax;
 
 public interface AreaCalculator {
-    int obtainArea();
+    double obtainArea();
 }
--- a/src/main/java/core/basesyntax/AreaCalculator.java
+++ b/src/main/java/core/basesyntax/AreaCalculator.java
@@ -1,0 +1,5 @@
+package core.basesyntax;
+
+public interface AreaCalculator {
+    int obtainArea();
+}
--- a/src/main/java/core/basesyntax/Circle.java
+++ b/src/main/java/core/basesyntax/Circle.java
@@ -1,0 +1,23 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+    private int radius;
+
+    public Circle(String color, int radius) {
+        super(color);
+        this.radius = radius;
+    }
+
+    @Override
+    public int obtainArea() {
+        return (int) Math.PI * radius * radius;
+    }
+
+    @Override
+    public void draw() {
+        System.out.println("Figure: circle, area: "
+                + obtainArea() + " sq. units, radius: "
+                + radius + " units, color: "
+                + getColor());
+    }
+}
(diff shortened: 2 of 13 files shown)


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,62 @@
...
+            case 1:
+                return getRandomRectangle();
+            case 2:
+                return getRandomCircle();
+            case 3:
+                return getRandomRightTriangle();
+            case 4:
+                return getRandomIsoscelesTrapezoid();
+            default:
+                return null;
```
petro-ushchuk, line 30: Don't return `null` as default value, better replace it with `getRandomIsoscelesTrapezoid()` and remove `case 4:`

petro-ushchuk (changes requested):

src/main/java/core/basesyntax/AreaCalculator.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface AreaCalculator {
+    double obtainArea();
```
karma-o, line 4: Calculator -> calculate
```suggestion
    double calculateArea();

```

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,23 @@
...
+    private int radius;
+
+    public Circle(String color, int radius) {
+        super(color);
+        this.radius = radius;
+    }
+
+    @Override
+    public double obtainArea() {
+        return (int) Math.PI * radius * radius;
```
karma-o, line 13: we can make our figures fields of type double to avoid type casting

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,60 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    public static final int DEFAULT_RADIUS = 10;
+    public static final String DEFAULT_COLOR = "WHITE";
```
karma-o, line 7:
```suggestion
    public static final String DEFAULT_COLOR = Color.WHITE.name();

```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,60 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    public static final int DEFAULT_RADIUS = 10;
+    public static final String DEFAULT_COLOR = "WHITE";
+    public static final int FIGURE_BOUND = 5;
+    public static final int RANDOM_NUMBER = 10;
```
karma-o, line 9: strange naming for a constant don't you think? :)
```suggestion
    public static final int RANDOM_NUMBER_BOUND = 10;

```

karma-o (approved):
Approved with some comments

PR #919, opened 2023-02-09
Advanced OOP task

kromch:


Color.java
```diff
@@ -0,0 +1,32 @@
+package core.basesyntax;
```
Andriy-dz, line 1: this class is duplicated. Let's delete it

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    public String getRandomColor() {
+        Random random = new Random();
```
Andriy-dz, line 7: it is better to bring this variable to the class level with the 'private' access modifier

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,11 @@
...
+        Color[] colors = Color.values();
+        return colors[random.nextInt(colors.length)].toString();
```
Andriy-dz, line 9: in enum it is better to use the name() method
```suggestion
        return colors[random.nextInt(colors.length)].name();
```

src/main/java/core/basesyntax/FiguresBehavior.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax;
+
+interface FiguresBehavior {
```
Andriy-dz, line 3: it is better to call the interface Drawable

src/main/java/core/basesyntax/FiguresShape.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum FiguresShape {
+    Square,
```
Andriy-dz, line 4: UpperCase

src/main/java/core/basesyntax/HelloWorld.java
```diff
@@ -1,8 +1,13 @@
 package core.basesyntax;
 
-/**
- * Feel free to remove this class and create your own.
- */
 public class HelloWorld {
-
+    public static void main(String[] args) {
+        Figure[] figures = new Figure[6];
```
Andriy-dz, line 5: magic number

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,28 @@
...
+        calcSquare();
+    }
+
+    public void draw() {
+        System.out.println(this.getClass().getSimpleName() + ": square, area: "
+                + String.format("%.2f", square) + " sq.units, radius: "
+                + String.format("%.2f", radius) + " units, color: " + color);
+    }
+    
+    private void calcSquare() {
```
Andriy-dz, line 25: corrections here and in the following cases
```suggestion
    private void calculateArea() {
```

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+public abstract class Figure implements FiguresBehavior {
+    protected double square;
```
Andriy-dz, line 4: it is better not to create this variable right away, because in the future you may change the value of the radius or the side and accordingly you will need to call the calculation of the area again. Let's just create a method that will calculate the area for us

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,51 @@
...
+    private static final double DEFAULT_VALUE = 10.0;
+    private static final int MULTIPLAYER = 100;
+    private ColorSupplier colorSupplier;
+
+    public FigureSupplier() {
+        this.colorSupplier = new ColorSupplier();
+    }
+
+    public Figure getRandomFigure() {
+        Random random = new Random();
```
Andriy-dz, line 15: it is better to bring this variable to the class level with the 'private' access modifier

Color.java
```diff
@@ -0,0 +1,32 @@
...
+
```
Andriy-dz, line 2: you do not create AreaCalculator interface

Andriy-dz (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
kromch: done

src/main/java/core/basesyntax/ColorSupplier.java
kromch: i did change method to method  that proposal you

src/main/java/core/basesyntax/HelloWorld.java
kromch: done

src/main/java/core/basesyntax/FigureSupplier.java
kromch: done

src/main/java/core/basesyntax/FiguresShape.java
kromch: done

Color.java
kromch: done

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,26 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+    private static final double PI = 3.14159265359;
```
sarakhmen, line 4: There is no need to create a custom PI constant, you can use Math.PI for your purposes

src/main/java/core/basesyntax/Drawable.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax;
+
+interface Drawable {
+    final String defaultColor = "WHITE";
```
sarakhmen, line 4: This interface shouldn't be responsible for the default color.

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+public abstract class Figure implements AreaCalculator, Drawable {
+    protected String color;
+    private ColorSupplier colorSuppliers;
+
+    public Figure(ColorSupplier colorSuppliers) {
+        this.colorSuppliers = colorSuppliers;
```
sarakhmen, line 8: Figure class shouldn't have colorSupplier dependency. Consider passing color to the Figure directly.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final double DEFAULT_VALUE = 10.0;
+    private static final int MULTIPLAYER = 100;
+    private ColorSupplier colorSupplier;
+    private Random random = new Random();
+
+    public FigureSupplier() {
+        this.colorSupplier = new ColorSupplier();
```
sarakhmen, line 12: consider initializing random and colorSupplier in a unified way, either using a constructor or initializing in the fields.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    public Figure getRandomFigure() {
+        FiguresShape[] figureshape = FiguresShape.values();
+        Figure fg = null;
+        switch (figureshape[random.nextInt(figureshape.length)].name()) {
```
sarakhmen, line 18: Use local variable instead.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,51 @@
...
+    private ColorSupplier colorSupplier;
+    private Random random = new Random();
+
+    public FigureSupplier() {
+        this.colorSupplier = new ColorSupplier();
+    }
+
+    public Figure getRandomFigure() {
+        FiguresShape[] figureshape = FiguresShape.values();
+        Figure fg = null;
```
sarakhmen, line 17: Is this variable really necessary? I'd recommend return figure directly, without this var usage

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,51 @@
...
+        switch (figureshape[random.nextInt(figureshape.length)].name()) {
+            case "SQUARE": {
```
sarakhmen, line 19: Use enums instead of String representation.

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,26 @@
...
+        super(colorSuppliers);
+        this.radius = radius;
+    }
+
+    public Circle(double radius) {
+        super(defaultColor);
+        this.radius = radius;
+    }
+
+    public void draw() {
```
sarakhmen, line 17: Consider using of Override annotation here and in the following cases

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,26 @@
...
+        System.out.println(this.getClass().getSimpleName() + ": square, area: "
+                + String.format("%.2f", calculateArea()) + " sq.units, radius: "
+                + String.format("%.2f", radius) + " units, color: " + color);
+    }
+    
+    public double calculateArea() {
```
sarakhmen, line 23: Consider using of Override annotation here and in the following cases

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,51 @@
...
+                fg = new Square(random.nextDouble() * MULTIPLAYER, colorSupplier);
```
sarakhmen, line 20: It's better to use a separate method for each figure creation

sarakhmen (changes requested):
Good solution, but let's improve it :)

src/main/java/core/basesyntax/Circle.java
Andriy-dz: not fixed. Please add Override annotaion

src/main/java/core/basesyntax/Circle.java
Andriy-dz: not fixed

src/main/java/core/basesyntax/FigureSupplier.java
Andriy-dz: not fixed

Andriy-dz (changes requested):

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,46 @@
...
+            }
+            case ISOSCELESTRAPEZOID: {
+                return new IsoscelesTrapezoid(random.nextDouble() * NUMBER_RANGE,
+                    random.nextDouble() * NUMBER_RANGE, random.nextDouble() * NUMBER_RANGE,
+                    colorSupplier.getRandomColor());
+            }
+            default: {
+            }
+        }
+        return getDefaultFigure();
```
petro-ushchuk, line 40: Move `return getDefaultFigure();` to `default` case

petro-ushchuk (approved):

PR #938, opened 2023-02-11
Maksym Yashyn HomeWork

Nimakel:


src/main/java/core/basesyntax/ApplicationMain.java
```diff
@@ -0,0 +1,23 @@
...
+
+import core.basesyntax.figures.Figure;
+import core.basesyntax.supplier.FigureSupplier;
+import java.util.Random;
+
+public class ApplicationMain {
+    public static void main(String[] args) {
+        Random random = new Random();
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] figures = new Figure[random.nextInt(8) + 1];
```
karma-o, line 11: why random array size?

src/main/java/core/basesyntax/figures/AreaCalculator.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax.figures;
+
+public interface AreaCalculator {
+    double getArea();
```
karma-o, line 4: Calculator -> calculate
```suggestion
    double calculateArea();
```

src/main/java/core/basesyntax/supplier/FigureSupplier.java
```diff
@@ -0,0 +1,54 @@
...
+
+public class FigureSupplier {
+    private static final int NUMBER_OF_FIGURE = 5;
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private RandomSupplier randomSupplier = new RandomSupplier();
+
+    public Figure getRandomFigure() {
+        Figure figure;
+        int figureNumber = random.nextInt(NUMBER_OF_FIGURE);
```
karma-o, line 21: since we have an enum, there is no need to hardcode this number
```suggestion
        int figureNumber = random.nextInt(FigureName.values().length);
```

src/main/java/core/basesyntax/figures/FigureSupplier.java
```diff
@@ -0,0 +1,4 @@
+package core.basesyntax.figures;
+
+public class FigureSupplier {
```
karma-o, line 3: remove this class

karma-o (changes requested):

src/main/java/core/basesyntax/supplier/FigureSupplier.java
Nimakel: All magic numbers in code should be constants. In this case that is a magic number. 
изображение

KseniiaMakarova, line 21: I agree
`since we have an enum, there is no need to hardcode this number`
see previous comment

KseniiaMakarova (changes requested):
overall good job! left few comments

sarakhmen (approved):
Good solution :)

PR #939, opened 2023-02-11
added all figures with unique properties; added draw and area interfa…

redmanO-o:
…ces; realized main method; added 2 classes for generating random and default results; added 2 interfaces; maven checkstyle return BUILD SUCCESS.

public Figure getDefaultFigure() - this method should always return white circle with radius 10. After generating the array, we need to display the entire list of objects that we have, for example:

### DIFF
--- a/src/main/java/core/basesyntax/AreaCalculator.java
+++ b/src/main/java/core/basesyntax/AreaCalculator.java
@@ -1,0 +1,5 @@
+package core.basesyntax;
+
+public interface AreaCalculator {
+    double getArea();
+}
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private Random randomColor = new Random();
+
+    public Color getRandomColor() {
+        return Color.values()[randomColor.nextInt(Color.values().length)];
+    }
+}
--- a/src/main/java/core/basesyntax/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/FigureSupplier.java
@@ -1,0 +1,72 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final double DEFAULT_PARAMETER = 10.0;
+    private static final Color DEFAULT_COLOR = Color.WHITE;
+
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private Random randomParameter = new Random();
+
+    public Figure getRandomFigure() {
+        switch (FigureName.values()[randomParameter.nextInt(FigureName.values().length)]) {
+            case SQUARE:
+                return getRandomSquare();
+
+            case CIRCLE:
+                return getRandomCircle();
+
+            case RECTANGLE:
+                return getRandomRectangle();
+
+            case RIGHT_TRIANGLE:
+                return getRandomRightTriangle();
+
+            default:
+                return getRandomIsoscelesTrapezoid();
+        }
+    }
+
+    public Square getRandomSquare() {
+        return new Square(colorSupplier.getRandomColor(),
+                getRandomParameter()
+        );
+    }
+
+    public Circle getRandomCircle() {
+        return new Circle(colorSupplier.getRandomColor(),
+                getRandomParameter());
+    }
+
+    public Rectangle getRandomRectangle() {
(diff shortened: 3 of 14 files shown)


src/main/java/core/basesyntax/AreaCalculator.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface AreaCalculator {
+    double getArea();
```
petro-ushchuk, line 4: Better name `calculateArea` since you have calculations inside but not just getting field value.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private Random randomColor = new Random();
+
+    public Color getRandomColor() {
+        return Color.values()[randomColor.nextInt(Color.values().length)];
```
petro-ushchuk, line 9: Better to introduce `randomColor.nextInt(Color.values().length)` as local variable

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,72 @@
...
+
+public class FigureSupplier {
+    private static final double DEFAULT_PARAMETER = 10.0;
+    private static final Color DEFAULT_COLOR = Color.WHITE;
+
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private Random randomParameter = new Random();
+
+    public Figure getRandomFigure() {
+        switch (FigureName.values()[randomParameter.nextInt(FigureName.values().length)]) {
```
petro-ushchuk, line 13: Better to introduce `randomParameter.nextInt(FigureName.values().length)` as local variable

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,21 @@
+package core.basesyntax;
+
+public class Main {
+
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] figures = new Figure[6];
```
petro-ushchuk, line 7: Magic number `6` should be constant.

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,21 @@
+package core.basesyntax;
+
+public class Main {
+
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] figures = new Figure[6];
+
```
petro-ushchuk, line 8: Remove redundant empty lines

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,21 @@
...
+        Figure[] figures = new Figure[6];
+
+        for (int i = 0; i < figures.length; i++) {
+            if (i < figures.length / 2) {
+                figures[i] = figureSupplier.getRandomFigure();
+                continue;
+            }
+            figures[i] = figureSupplier.getDefaultFigure();
+        }
+
```
petro-ushchuk, line 16: And here

petro-ushchuk (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private Random randomColor = new Random();
+    private int randomIndex = randomColor.nextInt(Color.values().length);
```
Andriy-dz, line 7: the mentor meant to make this change in a method, not as a class field. Because in this case you will always have the same index

src/main/java/core/basesyntax/IsoscelesTrapezoid.java
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    @Override
+    public void draw() {
+        System.out.println("Figure: isosceles trapezoid, area: "
+                + calculateArea() + " sq.units, top side: "
+                + topSide + " units, bottom side; "
+                + bottomSide + " units, height; "
+                + height + " color: "
+                + getColor().toString().toLowerCase());
```
Andriy-dz, line 27:
```suggestion
                + getColor().name().toLowerCase());

```

Andriy-dz (changes requested):

Andriy-dz (approved):
Perfect!!

PR #947, opened 2023-02-12
Implemented solution for hw

RandomEastEcho:


src/main/java/core/basesyntax/AreaCalculator.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface AreaCalculator {
+    double getArea();
```
petro-ushchuk, line 4: Better name `calculateArea` since you are calculating area, not just getting field value

src/main/java/core/basesyntax/FigurePrinter.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface FigurePrinter {
+    void printFigure();
+}
```
petro-ushchuk, line 5: Don't give to interfaces concrete names, better to have something abstract like `Printer` or `Drawable` or `Paintable`

https://howtodoinjava.com/java/basics/java-naming-conventions/

petro-ushchuk (changes requested):

src/main/java/core/basesyntax/Printer.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Printer {
+    void printFigure();
+}
```
petro-ushchuk, line 5: `Printer` is abstract separate interface, so method naming should not be tied up to figures.

src/main/java/core/basesyntax/calculateArea.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface calculateArea {
```
petro-ushchuk, line 3: Interfaces should be in the title case with the first letter of each separate word capitalized.

Rename it to `AreaCalculator`

src/main/java/core/basesyntax/calculateArea.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface calculateArea {
+    double getArea();
```
petro-ushchuk, line 4: Rename it to `calculateArea` since you make calculations, not just getting value from some field.

petro-ushchuk (changes requested):

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public abstract class Figure implements Printer, AreaCalculator  {
+    protected String color;
```
karma-o, line 4: it can be protected **but** this modifier also makes this field accessible inside the same package which is not consistent, so I would recommend making it private.
And in any way, it should have getters and setters

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,61 @@
...
+                random.nextInt(MAX_SIDE_VALUE),
+                random.nextInt(MAX_SIDE_VALUE));
+    }
+
+    public IsoscelesTrapezoid getIsoscelesTrapezoid() {
+        return new IsoscelesTrapezoid(colorSupplier.getRandomColor(),
+                random.nextInt(MAX_SIDE_VALUE),
+                random.nextInt(MAX_SIDE_VALUE),
+                random.nextInt(MAX_SIDE_VALUE));
+    }
```
karma-o, line 60: these methods can be public, considering the context of the class, but then they need an appropriate naming,
like `getRandomRectangle()` same for all of the above
In general, since we are not using these methods anywhere else right now, I'd also recommend making them private, but that is up to you

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,61 @@
...
+                return getRectangle();
+            case 3:
+                return getRightTriangle();
+            default:
+                return getIsoscelesTrapezoid();
+        }
+    }
+
+    public Figure defaultFigure() {
+        return new Circle(Color.WHITE.toString(),
```
karma-o, line 29: it is recommended to use `name()` instead of `toString()`

karma-o (approved):
Approved with some minor comments :)

PR #958, opened 2023-02-13
Home work oop advanced

BakhmetIvan:


src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,38 @@
+package core.basesyntax;
+
+public abstract class Figure implements AreaCalculator {
+    private FigureType name;
+    private String color;
+
+    public Figure() {
+        this.color = "Unknown";
+    }
```
aleksandr-bevza, line 9: It would be better not to allow creating figures without name or without color
```suggestion
```

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,36 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+    private double radius;
+
+    public Circle() {
+        super();
+        setName(FigureType.CIRCLE);
+        this.radius = 1;
+    }
```
aleksandr-bevza, line 10: It would be better not to allow creating Circle without color or unknown radius
```suggestion
```

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,36 @@
...
+    private double radius;
+
+    public Circle() {
+        super();
+        setName(FigureType.CIRCLE);
+        this.radius = 1;
+    }
+
+    public Circle(FigureType name, String color, double radius) {
+        super(name, color);
```
aleksandr-bevza, line 13: We don't need 'figure name' argument here. It is always the same here
```suggestion
    public Circle(String color, double radius) {
        super(FigureType.CIRCLE, color);
```

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,36 @@
...
+        this.radius = radius;
+    }
+
+    public double getRadius() {
+        return radius;
+    }
+
+    public void setRadius(float radius) {
+        this.radius = (radius >= 0) ? radius : 0;
+    }
```
aleksandr-bevza, line 23: It's better to initialize radius in constructor when we create new Circle.
```suggestion
```

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,38 @@
+package core.basesyntax;
+
+public abstract class Figure implements AreaCalculator {
```
aleksandr-bevza, line 3: It's better to create Interface 'Drawable' and implement method 'draw()' in Circle and other figures. Don't implement 'draw()' in this abstract class
image
```suggestion
public abstract class Figure implements AreaCalculator, Drawable {
```

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,38 @@
...
+
+    public void setColor(String color) {
+        this.color = color;
+    }
+
+    public String toDraw() {
+        StringBuilder builder = new StringBuilder();
+        return builder.append("Figure: ").append(this.name)
+                .append(", color: ").append(this.color).toString();
+    }
```
aleksandr-bevza, line 36:
```suggestion
```

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @Override
+    public double getArea() {
+        return Math.PI * Math.pow(this.radius, 2);
+    }
+
+    @Override
+    public String toDraw() {
```
aleksandr-bevza, line 31: method 'draw()' should be void and it should have 'System.out.println'

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,19 @@
...
+    public static void main(String[] args) {
+        Figure []figures = new Figure[MAX_FIGURES_LENGTH];
+        for (int i = 0; i < MAX_FIGURES_LENGTH / 2; i++) {
+            figures[i] = figureSupplier.getRandomFigure();
+        }
+        for (int i = MAX_FIGURES_LENGTH / 2; i < MAX_FIGURES_LENGTH; i++) {
+            figures[i] = figureSupplier.getDefaultFigure();
+        }
+        for (Figure figure : figures) {
+            System.out.println(figure.toDraw());
```
aleksandr-bevza, line 16:
```suggestion
            figure.draw();
```

aleksandr-bevza (changes requested):
Make fixes for all figures. Use 'Circle' as an example

aleksandr-bevza (approved):

PR #959, opened 2023-02-13
added solution, all classes and methods implemented

SanykVolodymyr:


src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+
```
nacenik, line 4: Delete redundant empty line
```suggestion
```

src/main/java/core/basesyntax/AreaCalculatable.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface AreaCalculatable {
+    public abstract double calculateArea();
```
nacenik, line 4: it's not necessary to add `public` and `abstract` to  interface methods, cause they are `public` and `abstract` by default
```suggestion
    double calculateArea();
```

src/main/java/core/basesyntax/Drawable.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Drawable {
+    public abstract void draw();
```
nacenik, line 4: it's not necessary to add `public` and `abstract` to  interface methods, cause they are `public` and `abstract` by default
```suggestion
    void draw();
```

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -0,0 +1,24 @@
+package core.basesyntax;
+
+public class Rectangle extends Figure {
+    private double sideA;
+    private double sideB;
```
nacenik, line 5: Use meaningful names
```suggestion
    private double width;
    private double height;
```

nacenik (changes requested):

PR #971, opened 2023-02-19
added solution

Serhii-Nosulia:


src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum Colors {
```
aleksandr-bevza, line 3: enums should not be named in plural form 
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,23 @@
+package core.basesyntax;
+
+public class Figure implements AreaCalculator, Drawable {
```
aleksandr-bevza, line 3: We should not be able to create new Figure because we don't khow how it looks like. Figure is something abstract
```suggestion
public abstract class Figure implements AreaCalculator, Drawable {
```

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,23 @@
...
+
+    @Override
+    public double getArea() {
+        return 0;
+    }
+
+    @Override
+    public void draw() {
+
+    }
```
aleksandr-bevza, line 22:
```suggestion
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,59 @@
...
+            default:
+                return getDefaultFigure();
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(DEFAULT_COLOR, DEFAULT_RADIUS);
+    }
+
+    public Figure getRandomSquare() {
```
aleksandr-bevza, line 35: This method and other similar methods should be private

src/main/java/core/basesyntax/Figures.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum Figures {
```
aleksandr-bevza, line 3: enums should not be named in plural form
```suggestion
public enum FigureType {
```

aleksandr-bevza (changes requested):

aleksandr-bevza (approved):

PR #981, opened 2023-03-02
first solution

sshestaka:


src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+    private static final String name = "Circle";
```
uliana-andrushchuk, line 4: You should use UPPERCASE for constants

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,22 @@
...
+    }
+
+    @Override
+    public double getArea() {
+        return Math.PI * radius * radius;
+    }
+
+    @Override
+    public void drawFigure() {
+        System.out.println("Figure: " + name + ", area: "
```
uliana-andrushchuk, line 18: You can use here `getClass().getSimpleName()` instead of `name` variable

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private final Random random = new Random();
+    private final int index = random.nextInt(Color.values().length);
```
uliana-andrushchuk, line 7: It's better to move this logic to method, there is no need to define index on class level

src/main/java/core/basesyntax/FigureBehaviour.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface FigureBehaviour {
```
uliana-andrushchuk, line 3: try to find better name for this interface, which will describe the behaviour

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,51 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier extends Figure {
+    private static final int range = 100;
+    private static ColorSupplier colorSupplier;
+    private static Random random = new Random();
```
uliana-andrushchuk, line 8:
```suggestion
    private Random random = new Random();
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,51 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier extends Figure {
+    private static final int range = 100;
+    private static ColorSupplier colorSupplier;
```
uliana-andrushchuk, line 7:
```suggestion
    private ColorSupplier colorSupplier;
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,51 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier extends Figure {
+    private static final int range = 100;
```
uliana-andrushchuk, line 6: You should use UPPERCASE for constants

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,51 @@
...
+    private static ColorSupplier colorSupplier;
+    private static Random random = new Random();
+    private static int index;
+    private String figureType;
+    private Figure figure;
+    private String figureColor;
+
+    public Figure getRandomFigure() {
+        colorSupplier = new ColorSupplier();
```
uliana-andrushchuk, line 15: You can initialize colorSupplier on class level (row 7)

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,51 @@
...
+                figure.setColor(figureColor);
+                return figure;
+            default:
+                return null;
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        figure = new Circle(10);
+        figure.setColor("WHITE");
```
uliana-andrushchuk, line 48: `10` `"WHITE"` - please, make these constants

src/main/java/core/basesyntax/HelloWorld.java
```diff
@@ -4,5 +4,19 @@
...
 
+    public static void main(String[] args) {
+        Figure figure = new Figure();
+        FigureSupplier randomFigure = new FigureSupplier();
+        Figure[] figures = new Figure[count];
+        for (int i = 0; i < figures.length; i++) {
+            figures[i] = (i < figures.length / 2) ? randomFigure.getRandomFigure()
+                    : randomFigure.getDefaultFigure();
+        }
+        for (Figure figure1 : figures) {
```
uliana-andrushchuk, line 17: please, do not use numbers in variable names

src/main/java/core/basesyntax/FigureType.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum FigureType {
+    Square,
+    Rectangle,
+    RightTriangle,
+    Circle,
+    IsoscelesTrapezoid
```
uliana-andrushchuk, line 8: The fields in enum must be written in uppercase

uliana-andrushchuk (changes requested):

src/main/java/core/basesyntax/Square.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax;
+
+public class Square extends GetFigure {
+    private static final String NAME = "SQUARE";
```
okuzan, line 4: delete name from all figures, it doesn't add anything useful

src/main/java/core/basesyntax/GetFigure.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax;
+
+public class GetFigure implements GetFigureArea {
```
okuzan, line 3: this should be abstract class `Shape` or `Figure`. This is the base of your hierarchy - all your figures subclass it. 
Never name classes with verbs, only with nouns. No need to have all this logic, leave only color, and constructor for color.

src/main/java/core/basesyntax/GetFigureArea.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface GetFigureArea {
```
okuzan, line 3:
```suggestion
public interface AreaCalculator {
```

src/main/java/core/basesyntax/GetFigureSupplier.java
```diff
@@ -0,0 +1,51 @@
...
+
+public class GetFigureSupplier extends GetFigure {
+    private static final int RANGE = 100;
+    private static final String DEFAULT_COLOR = "WHITE";
+    private static int index;
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private Random random = new Random();
+    private String figureType;
+    private GetFigure figure;
+    private String figureColor;
```
okuzan, line 13: these have to be local variables

src/main/java/core/basesyntax/GetFigureSupplier.java
```diff
@@ -0,0 +1,51 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class GetFigureSupplier extends GetFigure {
+    private static final int RANGE = 100;
+    private static final String DEFAULT_COLOR = "WHITE";
+    private static int index;
```
okuzan, line 8: same for this

src/main/java/core/basesyntax/GetFigureSupplier.java
```diff
@@ -0,0 +1,51 @@
...
+    private Random random = new Random();
+    private String figureType;
+    private GetFigure figure;
+    private String figureColor;
+
+    public GetFigure getRandomFigure() {
+        index = random.nextInt(FigureType.values().length);
+        figureType = FigureType.values()[index].name();
+        figureColor = colorSupplier.getRandomColor();
+        switch (figureType) {
```
okuzan, line 19: that's the perfect opportunity to use enum, why would you willingly rely on strings instead

src/main/java/core/basesyntax/GetFigureSupplier.java
```diff
@@ -0,0 +1,51 @@
...
+            case "SQUARE":
+                figure = new Square(random.nextInt(RANGE));
+                figure.setColor(figureColor);
+                return figure;
```
okuzan, line 23: extract logic from each case to dedicated private method

src/main/java/core/basesyntax/GetFigureSupplier.java
```diff
@@ -0,0 +1,51 @@
...
+                figure = new Circle(random.nextInt(RANGE));
+                figure.setColor(figureColor);
+                return figure;
+            case "ISOSCELES_TRAPEZOID":
+                figure = new IsoscelesTrapezoid(random.nextInt(RANGE),
+                        random.nextInt(RANGE), random.nextInt(RANGE));
+                figure.setColor(figureColor);
+                return figure;
+            default:
+                return null;
```
okuzan, line 42: here return default figure

src/main/java/core/basesyntax/GetFigureSupplier.java
```diff
@@ -0,0 +1,51 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class GetFigureSupplier extends GetFigure {
```
okuzan, line 5:
```suggestion
public class FigureSupplier {
```

okuzan: that's not use case for hierarchy

src/main/java/core/basesyntax/GetFigure.java
okuzan: Create interface `Drawable` with void method `draw` and implement it here, then override in subclasses

okuzan (changes requested):

src/main/java/core/basesyntax/Square.java
sshestaka: Why? I disagree with it...

src/main/java/core/basesyntax/GetFigure.java
sshestaka: Except class name, I disagree with these comments. I don't understand what exactly is done wrong.

src/main/java/core/basesyntax/GetFigureArea.java
sshestaka: OK

src/main/java/core/basesyntax/GetFigureSupplier.java
sshestaka: Why?

src/main/java/core/basesyntax/GetFigureSupplier.java
sshestaka: Disagree...

src/main/java/core/basesyntax/GetFigureSupplier.java
sshestaka: I think it's a good idea to use ENUM and CASE for this task

src/main/java/core/basesyntax/GetFigureSupplier.java
sshestaka: Disagree... I don't  understandt why?

src/main/java/core/basesyntax/GetFigureSupplier.java
sshestaka: OK, maybe...

src/main/java/core/basesyntax/GetFigureSupplier.java
sshestaka: ok

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,22 @@
...
+    }
+
+    @Override
+    public double getArea() {
+        return 0;
+    }
+
+    public void drawFigure() {
+        System.out.println("Figure color: " + color);
+    }
```
kozhukhovsky, line 21:
```suggestion
    public void drawFigure() {
        System.out.println("Figure color: " + color);
    }
```

    move method draw to the separate interface Drawable, like AreaCalulator with the same logic for implementing

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,51 @@
...
+
+    public Figure getRandomFigure() {
+        index = random.nextInt(FigureType.values().length);
+        figureType = FigureType.values()[index].name();
+        figureColor = colorSupplier.getRandomColor();
+        switch (figureType) {
+            case "SQUARE":
+                figure = new Square(random.nextInt(RANGE));
+                figure.setColor(figureColor);
+                return figure;
```
kozhukhovsky, line 23:
```suggestion
                figure = new Square(random.nextInt(RANGE));
                figure.setColor(figureColor);
                return figure;
```

move creating each figures to the separate method, like getRandomSquare()

kozhukhovsky (changes requested):

src/main/java/core/basesyntax/Figure.java
sshestaka: Ok

src/main/java/core/basesyntax/FigureSupplier.java
sshestaka: I can't do this. Because in each CASE block I create a different figure! So, in this case I should made as many separate methods as CASE blocks I have. I don't think that it's good idea. I can put all CASE block in separate method...but I really don't see the point in this too...

src/main/java/core/basesyntax/Square.java
olekskov: Code should contain only what is required to solve the task. Random constant that is not used doesn't help to solve the task. In same way we may add some emojis or links to favourite memes

olekskov (changes requested):

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+    private static final String NAME = "CIRCLE";
```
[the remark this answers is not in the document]
sshestaka: Yes, you are right... but it was used...and then I have changed it to getClass().getSimpleName()

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax;
+
+public class Figure implements AreaCalculator, Drawable {
```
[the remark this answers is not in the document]
sshestaka: OK!

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,22 @@
...
+    public void setColor(String color) {
+        this.color = color;
+    }
+
+    @Override
+    public double getArea() {
+        return 0;
+    }
+
+    public void drawFigure() {
```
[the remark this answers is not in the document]
sshestaka: Ofcourse! I didn't notice this...

src/main/java/core/basesyntax/IsoscelesTrapezoid.java
```diff
@@ -0,0 +1,28 @@
+package core.basesyntax;
+
+public class IsoscelesTrapezoid extends Figure {
+    private static final String NAME = "ISOSCELES TRAPEZOID";
```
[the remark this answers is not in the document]
sshestaka: ok

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,43 @@
...
+            case "ISOSCELES_TRAPEZOID":
+                return new IsoscelesTrapezoid(getRandomSide(RANGE),
+                        getRandomSide(RANGE), getRandomSide(RANGE), figureColor);
+            default:
+                return new FigureSupplier().getDefaultFigure();
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(10, DEFAULT_COLOR);
```
uliana-andrushchuk, line 37: please, make `10` constant

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,43 @@
...
+    private Random random = new Random();
+    private String figureType;
+    private Figure figure;
+    private String figureColor;
+
+    public Figure getRandomFigure() {
+        index = random.nextInt(FigureType.values().length);
+        figureType = FigureType.values()[index].name();
+        figureColor = colorSupplier.getRandomColor();
+        switch (figureType) {
```
uliana-andrushchuk, line 19: Let's use FigureType enum for this switch case

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private final Random random = new Random();
+    private int index;
```
uliana-andrushchuk, line 7: Here we do not need to have `index` on class level.
This variable related only to the logic in `getRandomColor()` method.
Its common practice in Java to use such fields only in the method and not create them at class level.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,43 @@
...
+
+public class FigureSupplier {
+    private static final int RANGE = 100;
+    private static final String DEFAULT_COLOR = "WHITE";
+    private static int index;
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private Random random = new Random();
+    private String figureType;
+    private Figure figure;
+    private String figureColor;
```
uliana-andrushchuk, line 13: >     private static int index;
>     private String figureType;
>     private Figure figure;
>     private String figureColor;

There is no need to have these fields at class level.

uliana-andrushchuk (changes requested):

src/main/java/core/basesyntax/FigureSupplier.java
sshestaka: OK

src/main/java/core/basesyntax/FigureSupplier.java
sshestaka: I don't understand what do you mean... So, I put FigureType.values()[index].name() inside SWITCH condition without additional variable (figureType).

src/main/java/core/basesyntax/ColorSupplier.java
sshestaka: OK

src/main/java/core/basesyntax/FigureSupplier.java
sshestaka: OK!

olekskov (approved):
Good job!

PR #989, opened 2023-03-10
Hw 2 solution

Alex-Solod:


src/main/java/core/basesyntax/FigureColor.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum FigureColor {
```
okuzan, line 3:
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public interface Figure {
```
okuzan, line 3: make it an abstract class

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,20 @@
...
+            figure[i] = figureSupplier.getRandomFigure();
+        }
+
+        for (int i = 3; i < 6; i++) {
+            figure[i] = figureSupplier.getRandomFigure();
+        }
+
+        for (Figure figures : figure) {
+            figures.draw();
+        }
```
okuzan, line 18: you need only one loop here

okuzan (changes requested):
Don't request mentor's review until your build passes

PR #994, opened 2023-03-14
I worked on interfaces and abstract classes. I added 5 classes of fig…

vrnvv:
…ures that are inherited from an abstract class, which implements interfaces

### DIFF
--- a/src/main/java/core/basesyntax/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/FigureSupplier.java
@@ -4,9 +4,9 @@ import java.util.Random;
 
 public class FigureSupplier {
     private static final int MAX_SIDE = 25;
-    private static Random random = new Random();
     private static final int FIGURE_COUNT = 5;
-    private ColorSupplier colorSupplier = new ColorSupplier();
+    private final Random random = new Random();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
 
     public Figure getRandomFigure() {
         int figureNumber = random.nextInt(FIGURE_COUNT);
--- a/src/main/java/core/basesyntax/IsoscelesTrapezoid.java
+++ b/src/main/java/core/basesyntax/IsoscelesTrapezoid.java
@@ -22,7 +22,7 @@ public class IsoscelesTrapezoid extends Figure {
 
     @Override
     public Double getArea() {
-        return ((getTopLength() + getDownLength()) * getHeight()) / 2;
+        return (getTopLength() + getDownLength()) * getHeight() / 2;
     }
 
     private double getTopLength() {
--- a/src/main/java/core/basesyntax/Main.java
+++ b/src/main/java/core/basesyntax/Main.java
@@ -6,16 +6,13 @@ public class Main {
         int length = 10;
         Figure[] figures = new Figure[length];
 
-        for (int i = 0; i < length / 2; i++) {
-            figures[i] = figureSupplier.getRandomFigure(); 
-        }
-
-        for (int i = length / 2; i < length; i++) {
-            figures[i] = figureSupplier.getDefaultFigure();
-        }
-
-        for (Figure figure: figures) {
-            figure.draw();
+        for (int i = 0; i < length; i++) {
+            if (i < length / 2) {
+                figures[i] = figureSupplier.getRandomFigure();    
+            } else {
+                figures[i] = figureSupplier.getDefaultFigure();
+            }
+            figures[i].draw();
         }
     }
 
--- a/src/main/java/core/basesyntax/RightTriangle.java
+++ b/src/main/java/core/basesyntax/RightTriangle.java
@@ -12,7 +12,7 @@ public class RightTriangle extends Figure {
     
     @Override
     public Double getArea() {
-        return (getFirstLeg() * getSecondLeg()) / 2;
+        return getFirstLeg() * getSecondLeg() / 2;
     }
 
     @Override


src/main/java/core/basesyntax/Behaviour.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Behaviour {
```
uliana-andrushchuk, line 3: It's better to have more informative name for interface (like `Drawable`)

src/main/java/core/basesyntax/CalcArea.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface CalcArea {
```
uliana-andrushchuk, line 3: Please, do not use short word form in class and interface names

src/main/java/core/basesyntax/Color.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum Color {
+    RED,
+    Green,
```
uliana-andrushchuk, line 5: Please, use UPPERCASE for all fields

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,55 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static int PROPERTIES = 25;
```
uliana-andrushchuk, line 6:
```suggestion
    private static final int PROPERTIES = 25;
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,55 @@
...
+    private static Random random = new Random();
+    private static final int FIGURECOUNT = 5;
```
uliana-andrushchuk, line 8:
```suggestion
    private static final int FIGURE_COUNT = 5;
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,55 @@
...
+
+    public static Figure getRandomFigure() {
+        int figure = random.nextInt(FIGURECOUNT);
+        String color = new ColorSupplier().getRandomColor();
```
uliana-andrushchuk, line 12: Please, make ColorSupplier class-level field

src/main/java/core/basesyntax/FigureSupplier.java
uliana-andrushchuk: it would be better to change the variable name to more informative

src/main/java/core/basesyntax/RightTriangle.java
```diff
@@ -0,0 +1,32 @@
...
+    @Override
+    public Double getArea() {
+        return (getFirstLeg() * getSecondLeg()) / 2;
+    }
+
+    @Override
+    public void draw() {
+        System.out.println("Figure: right triangle, area: " + getArea() + " sq.units, firstLeg: "
+                + getFirstLeg() + ", secondLeg: " + getSecondLeg() + " units, color: " + color);
+
```
uliana-andrushchuk, line 22:
```suggestion
```
remove empty line

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+public abstract class Figure implements Behaviour, CalcArea {
+    // Why can't I write 2 behavior in 1 interface(draw,getArea)?
```
uliana-andrushchuk, line 4: Using multiple interfaces allows for more flexibility in the code. It allows different classes to implement only the methods they need, rather than implementing a large number of methods that may not be relevant to them. 

Here it would be better to have two separate interfaces with their own method, rather than combining them into a single interface. This would allow classes implementing these interfaces to only implement the specific methods they need, and it would make the code easier to read and maintain.

uliana-andrushchuk: Technically, you can write 2 methods in one interface, but it wouldn't be a good object-oriented programming practice.

uliana-andrushchuk (changes requested):

src/main/java/core/basesyntax/Figure.java
vrnvv: Thank you

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,56 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int MAX_SIDE = 25;
+    private static Random random = new Random();
+    private static final int FIGURE_COUNT = 5;
+    private ColorSupplier colorSupplier = new ColorSupplier();
```
olekskov, line 9: Overall regarding class properties we start with constants (private static final) and then proceed with other variables.
Also variables for services we rely on are usually private final as we do not plan to pass other object to them and will use same within all class methods
```suggestion
    private static final int MAX_SIDE = 25;
    private static final int FIGURE_COUNT = 5;
    private final Random random = new Random();
    private final ColorSupplier colorSupplier = new ColorSupplier();
```

src/main/java/core/basesyntax/IsoscelesTrapezoid.java
```diff
@@ -0,0 +1,39 @@
...
+    public void draw() {
+        System.out.println("Figure: isosceles trapezoid, area: " + getArea()
+                + " sq.units, topLength: " + getTopLength()
+                + " units, downLength: " + getDownLength()
+                + " units, height: " + getHeight() + " units, color: " + color);
+    }
+
+    @Override
+    public Double getArea() {
+        return ((getTopLength() + getDownLength()) * getHeight()) / 2;
```
olekskov, line 25:
```suggestion
        return (getTopLength() + getDownLength()) * getHeight() / 2;
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,22 @@
...
+            figures[i] = figureSupplier.getRandomFigure(); 
+        }
+
+        for (int i = length / 2; i < length; i++) {
+            figures[i] = figureSupplier.getDefaultFigure();
+        }
+
+        for (Figure figure: figures) {
+            figure.draw();
+        }
```
olekskov, line 19: could we make all three actions within one cycle?

src/main/java/core/basesyntax/RightTriangle.java
```diff
@@ -0,0 +1,31 @@
...
+    public RightTriangle(String color, double firstLeg, double secondLeg) {
+        super(color);
+        this.firstLeg = (int) firstLeg;
+        this.secondLeg = (int) secondLeg;
+    }
+    
+    @Override
+    public Double getArea() {
+        return (getFirstLeg() * getSecondLeg()) / 2;
```
olekskov, line 15:
```suggestion
        return getFirstLeg() * getSecondLeg() / 2;
```

olekskov (changes requested):

aaananas (approved):

PR #1000, opened 2023-03-19
Figure br

jakop51:


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,49 @@
...
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int AMOUNT_OF_FIGURES = 5;
+    private static final int MAX_RANDOM_VALUE = 100;
+    private Random random = new Random();
+    private int numOfFigure;
+    private String color;
+    private ColorSupplier colorSupplier;
```
olekskov, line 11: Services we rely on are better not to initialize with new in some method as they may be very heavy and we will spend lot of resources creating them each time we call method
At the same time numOfFigure and color are used in 1 method only so we may easily use them as method variables

```suggestion
    private final Random random = new Random();
    private final ColorSupplier colorSupplier = new ColorSupplier();
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,49 @@
...
+    public Figure getRandomFigure() {
+        Figure figure;
+        numOfFigure = random.nextInt(AMOUNT_OF_FIGURES);
+        colorSupplier = new ColorSupplier();
+        color = colorSupplier.getRandomColor();
+        switch (numOfFigure) {
+            case 0:
+                figure = new Rectangle("Rectangle", color, random.nextInt(MAX_RANDOM_VALUE) + 1,
+                        random.nextInt(MAX_RANDOM_VALUE) + 1);
+                break;
```
olekskov, line 22: If we make instant return of required figure we will not need figure variable and will not require breaks in our switch (same for all cases)
```suggestion
                return new Rectangle("Rectangle", color, random.nextInt(MAX_RANDOM_VALUE) + 1,
                        random.nextInt(MAX_RANDOM_VALUE) + 1);
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,49 @@
...
+    private int numOfFigure;
+    private String color;
+    private ColorSupplier colorSupplier;
+
+    public Figure getRandomFigure() {
+        Figure figure;
+        numOfFigure = random.nextInt(AMOUNT_OF_FIGURES);
+        colorSupplier = new ColorSupplier();
+        color = colorSupplier.getRandomColor();
+        switch (numOfFigure) {
```
olekskov, line 18: We should create additional variables when we really need them for logic or when introducing additional variable with some specific name will increase readability (like color here, it really makes further code more readable so it's fully ok). In other cases we may use expression itself:
```suggestion
        color = colorSupplier.getRandomColor();
        switch (random.nextInt(AMOUNT_OF_FIGURES)) {
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,49 @@
...
+                        random.nextInt(MAX_RANDOM_VALUE) + 1);
+                break;
+        }
+        return figure;
+    }
+
+    public Figure getDefaultFigure() {
+        int radius = 10;
+        Circle circle = new Circle("Circle", Color.WHITE.name(), radius);
+        return circle;
```
olekskov, line 47: don't create variables with obvious name to be used once:
```suggestion
        return new Circle("Circle", Color.WHITE.name(), radius);
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,49 @@
...
+                figure = new RightTriangle("Right Triangle", color,
+                        random.nextInt(MAX_RANDOM_VALUE) + 1,
+                        random.nextInt(MAX_RANDOM_VALUE) + 1);
+                break;
+        }
+        return figure;
+    }
+
+    public Figure getDefaultFigure() {
+        int radius = 10;
```
olekskov, line 45: magic values should be class constants

olekskov (changes requested):

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,24 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+    private double radius;
+
+    public Circle(String name, String color, double radius) {
+        super.name = name;
+        super.color = color;
```
uliana-andrushchuk, line 8: Please, remove `name` field in class Figure and use `super(color)` to set color

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,24 @@
...
+
+    @Override
+    public double getArea() {
+        double circleArea = Math.PI * radius * radius;
+        return circleArea;
+    }
+
+    @Override
+    public void printInformation() {
+        System.out.println("Figure: " + name + ", area: "
```
uliana-andrushchuk, line 20: You can use `getClass().getSimpleName()` instead of `name`

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,15 @@
...
+import java.util.Random;
+
+public class ColorSupplier {
+    private Random randomValue = new Random();
+    private int colorsCount;
+
+    public String getRandomColor() {
+        colorsCount = Color.values().length;
+        int colorPosition = randomValue.nextInt(colorsCount);
+        String randomColor = Color.values()[colorPosition].name();
```
uliana-andrushchuk, line 12:
```suggestion
        return Color.values()[colorPosition].name();

```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private Random randomValue = new Random();
+    private int colorsCount;
```
uliana-andrushchuk, line 7: You don't need this variable on class level

uliana-andrushchuk (changes requested):

aaananas (approved):

PR #1007, opened 2023-03-26
OOP solution

z1row:


src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,21 @@
...
+    private final int radius;
+
+    public Circle(int radius, String color) {
+        this.radius = radius;
+        this.color = color;
+    }
+
+    @Override
+    public double getArea() {
+        return Math.PI * radius;
```
olekskov, line 13: seems circle area has slightly another formula

src/main/java/core/basesyntax/Color.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public enum Color {
+    black, white, yellow, green, pink, red, brown, purple
```
olekskov, line 4: enum elements naming follows constants naming convention

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,30 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private final Random random = new Random();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        final int index = random.nextInt(5);
```
olekskov, line 10: this and all below magic values should be constants with proper naming

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+public class Main {
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] figures = new Figure[6];
```
olekskov, line 6: also magic numbers shouldn't be inlined, use constant

src/main/java/core/basesyntax/RightTriangle.java
```diff
@@ -0,0 +1,21 @@
+package core.basesyntax;
+
+public class RightTriangle extends Figure {
+    private final int side;
```
olekskov, line 4: I don't fully understand how we may describe right triangle with one side.
Seems we need 2 sides that create right corner

src/main/java/core/basesyntax/RightTriangle.java
```diff
@@ -0,0 +1,21 @@
...
+
+    @Override
+    public void draw() {
+        System.out.println("Figure: right triangle, area: " + String.format("%.2f", getArea())
+                + ", side: " + side + ", color: " + color);
+    }
+
+    @Override
+    public double getArea() {
+        return (Math.sqrt(3) / 4) * Math.pow(side, 2);
```
olekskov, line 19: after we have 2 sides area will be half of rectangle

olekskov (changes requested):

src/main/java/core/basesyntax/FigureSupplier.java
z1row: image
image
i made it constants and got this

PR #1008, opened 2023-03-27
Oop solution

z1row:


src/main/java/core/basesyntax/GetArea.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface GetArea {
```
uliana-andrushchuk, line 3: it would be better to name this interface `AreaCalculator`

src/main/java/core/basesyntax/Drawing.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Drawing {
```
uliana-andrushchuk, line 3: It would be better to name this interface Drawable

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,34 @@
...
+public class FigureSupplier {
+    private final Random random = new Random();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        final String color = colorSupplier.getRandomColor();
+        final int index = random.nextInt(5);
+        final int randomLength_1 = random.nextInt(100);
+        final int randomLength_2 = random.nextInt(100);
+        final int randomLength_3 = random.nextInt(100);
```
uliana-andrushchuk, line 14: It's not a good practice to use numbers in variable names, and also underscore is not used for variable names in java

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,34 @@
...
+                return new Rectangle(randomLength_1, randomLength_2, color);
+            case 3:
+                return new RightTriangle(randomLength_1, randomLength_2, color);
+            default:
+                return new Square(randomLength_1, color);
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(10, Color.WHITE.toString());
```
uliana-andrushchuk, line 32: Please, do not use magic numbers. Create a constant

uliana-andrushchuk (changes requested):

src/main/java/core/basesyntax/FigureSupplier.java
uliana-andrushchuk: not fixed
please create constants for these fields
`private static final int DEFAULT_RADIUS = 10;`
same for color

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+public class Main {
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] figures = new Figure[6];
+        for (int i = 0; i < 6; i++) {
+            if (i < 6 / 2) {
```
uliana-andrushchuk, line 8: Please make check if it is first or second half of array

uliana-andrushchuk (changes requested):

PR #1009, opened 2023-03-28
solved task with using abstract class and interface

OlehPryshchepa:


src/main/java/core/basesyntax/IsoscelesTrapezoid.java
```diff
@@ -0,0 +1,38 @@
...
+                + getArea()
+                + " sq.units"
+                + ", a side: "
+                + aside
+                + " units, b side: "
+                + bside
+                + " units, heigth: "
+                + heigth
+                + " units, color: "
+                + getColor();
```
rovein, line 36: From the requirements:

```
To 'draw' means to print out all information about a figure using System.out.println() (you shouldn't override toString() method for this)
```

So, create an interface with one abstract method `draw()` and implement it in your `Figure`.

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -0,0 +1,33 @@
...
+                + ", area: "
+                + getArea()
+                + " sq.units"
+                + ", a side: "
+                + aside
+                + " units, b side: "
+                + bside
+                + " units, color: "
+                + getColor();
+    }
```
rovein, line 32: From the requirements:

```
To 'draw' means to print out all information about a figure using System.out.println() (you shouldn't override toString() method for this)
```

So, create an interface with one abstract method `draw()` and implement it in your `Figure`.

src/main/java/core/basesyntax/RightTriangle.java
```diff
@@ -0,0 +1,34 @@
...
+                + getFigureType()
+                + ", area: "
+                + getArea()
+                + " sq.units"
+                + ", first leg: "
+                + firstLeg
+                + " units, second leg: "
+                + secondLeg
+                + " units, color: "
+                + getColor();
```
rovein, line 32: From the requirements:

```
To 'draw' means to print out all information about a figure using System.out.println() (you shouldn't override toString() method for this)
```

So, create an interface with one abstract method `draw()` and implement it in your `Figure`.

src/main/java/core/basesyntax/Square.java
```diff
@@ -0,0 +1,30 @@
...
+    public String toString() {
+        return super.toString()
+                + getFigureType()
+                + ", area: "
+                + getArea()
+                + " sq.units"
+                + ", side: "
+                + side
+                + " units, color: "
+                + getColor();
```
rovein, line 28: From the requirements:

```
To 'draw' means to print out all information about a figure using System.out.println() (you shouldn't override toString() method for this)
```

So, create an interface with one abstract method `draw()` and implement it in your `Figure`.

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,30 @@
...
+    public String toString() {
+        return super.toString()
+                + getFigureType()
+                + ", area: "
+                + getArea()
+                + " sq.units"
+                + ", radius: "
+                + radius
+                + " units, color: "
+                + getColor();
```
rovein, line 28: From the requirements:

```
To 'draw' means to print out all information about a figure using System.out.println() (you shouldn't override toString() method for this)
```

So, create an interface with one abstract method `draw()` and implement it in your `Figure`.

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,41 @@
+package core.basesyntax;
+
+public abstract class Figure implements AreaCalculator {
+    private String color;
+    private double area;
+    private String figureType;
```
rovein, line 6: You don't need it here.

If you change your `AreaCalculator` interface - you can retrieve the area from the `calculateArea()` method.

For the figure type - you can specify it directly in every method `draw()` (in your current case `toString()`).

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,41 @@
...
+        this.area = area;
+    }
+
+    public String getFigureType() {
+        return figureType;
+    }
+
+    public void setFigureType(String figureType) {
+        this.figureType = figureType;
+    }
```
rovein, line 30: If you remove unnecessary fields - don't forget to remove the setters/getters.

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Override
+    public String toString() {
+        return "Figure: ";
+    }
```
rovein, line 35: Don't override it, see requirements.

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Override
+    public void calculateArea() {
+
+    }
```
rovein, line 40: Don't override a method and provide it with empty behavior, it's a bad practice.

Instead, don't override it at all, we don't need to do it in the abstract class.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    public String getRandomColor() {
+        int index = new Random().nextInt(Colors.values().length);
+        return Colors.values()[index].toString().toLowerCase();
```
rovein, line 8: Same as from the previous assignment. Previous comments:

```
Do we really need to create a new instance of Random every time when this method is called?
It is better to create a new Random field as class property.
```

```
Your index has the same value on each getRandomColor() call. 
Let's better calculate index value within the method and make your Random instance a class field (initialized directly in the field or using a constructor)
```

Also, [Use name() for getting String representation of enum constants](https://mate-academy.github.io/jv-program-common-mistakes/java-core/abstract-class-interface/oop-advanced#use-name-for-getting-string-representation-of-enum-constants).

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,67 @@
...
+                                                colorSupplier.getRandomColor());
+                break;
+            default:
+                break;
+        }
+        return figure;
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(10, "white");
```
rovein, line 60: Magic numbers should be moved to constants. [Reference](https://mate-academy.github.io/jv-program-common-mistakes/java-core/abstract-class-interface/oop-advanced#all-magic-numbers-in-your-code-should-be-constants).

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,67 @@
...
+                figure = new IsoscelesTrapezoid(trapezoidRandomeASide,
+                                                trapezoidRandomeBSide,
+                                                trapezoidRandomeHeigth,
+                                                colorSupplier.getRandomColor());
+                break;
+            default:
+                break;
+        }
+        return figure;
+    }
```
rovein, line 57: All magic numbers within this method should be moved to constants. [Rererence](https://mate-academy.github.io/jv-program-common-mistakes/java-core/abstract-class-interface/oop-advanced#all-magic-numbers-in-your-code-should-be-constants).

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,67 @@
...
+            case "isoscelestrapezoid":
+                trapezoidRandomeASide = random.nextInt(100);
+                trapezoidRandomeBSide = random.nextInt(100);
+                trapezoidRandomeHeigth = random.nextInt(100);
+                figure = new IsoscelesTrapezoid(trapezoidRandomeASide,
+                                                trapezoidRandomeBSide,
+                                                trapezoidRandomeHeigth,
+                                                colorSupplier.getRandomColor());
+                break;
+            default:
```
rovein, line 53: Don't create the `default` block empty, instead transform your last `case` to default. See [here](https://mate-academy.github.io/jv-program-common-mistakes/java-core/abstract-class-interface/oop-advanced#dont-return-null-from-a-method).

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,67 @@
...
+    private Random random = new Random();
+    private int randomeRadius = 0;
+    private int squareRandomeSide = 0;
+    private int rectangleRandomASide = 0;
+    private int rectangleRandomBSide = 0;
+    private int firstRandomeLeg = 0;
+    private int secondRandomeLeg = 0;
+    private int trapezoidRandomeASide = 0;
+    private int trapezoidRandomeBSide = 0;
+    private int trapezoidRandomeHeigth = 0;
```
rovein, line 16: It's good that you don't pass expressions to constructors, instead using variables. 

But don't create them all on the class level. Instead move the logic of creating a random figure into the separate method and store those variables here.

[Reference](https://mate-academy.github.io/jv-program-common-mistakes/java-core/abstract-class-interface/oop-advanced#think-about-which-variables-should-be-local-in-the-method-and-which-should-be-class-level).

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,67 @@
...
+
+    public Figure getRandomFigure() {
+        String figureType = getRandomeType();
+        Figure figure = null;
+        switch (figureType) {
+            case "circle":
+                randomeRadius = random.nextInt(100);
+                figure = new Circle(randomeRadius, colorSupplier.getRandomColor());
+                break;
```
rovein, line 25: Move it and all other `case` blocks statements into the separate methods, which returns the random figure.

src/main/java/core/basesyntax/HelloWorld.java
```diff
@@ -3,6 +3,22 @@
...
+            if (i < figures.length / 2) {
+                figures[i] = figureSupplier.getRandomFigure();
+            } else {
+                figures[i] = figureSupplier.getDefaultFigure();
+            }
+        }
 
+        for (Figure figure : figures) {
+            System.out.println(figure);
+        }
```
rovein, line 22: Do we need an additional loop to draw all figures? Could we draw them using the previous loop?

src/main/java/core/basesyntax/HelloWorld.java
```diff
@@ -3,6 +3,22 @@
 /**
  * Feel free to remove this class and create your own.
  */
```
rovein, line 5: Remove unnecessary comments.

src/main/java/core/basesyntax/Figures.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+public enum Figures {
```
rovein, line 3: Check the previous comment, you don't need this enum at all.

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,41 @@
+package core.basesyntax;
+
+public abstract class Figure implements AreaCalculator {
+    private String color;
```
rovein, line 4: Maybe we can store the reference of your enum type `Color`?

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,30 @@
...
+
+public class Circle extends Figure {
+    private int radius;
+
+    public Circle(int radius, String color) {
+        this.radius = radius;
+        setFigureType("circle");
+        setColor(color);
+        calculateArea();
+    }
```
rovein, line 11: See my comments to the `Figure` class, firstly change them, and then you will need to change all the inheritors.

src/main/java/core/basesyntax/IsoscelesTrapezoid.java
```diff
@@ -0,0 +1,38 @@
...
+    private int heigth;
+
+    public IsoscelesTrapezoid(int aside, int bside, int heigth, String color) {
+        this.aside = aside;
+        this.bside = bside;
+        this.heigth = heigth;
+        setColor(color);
+        setFigureType("isosceles trapezoid");
+        calculateArea();
+    }
```
rovein, line 15: See my comments to the `Figure` class, firstly change them, and then you will need to change all the inheritors.

src/main/java/core/basesyntax/RightTriangle.java
```diff
@@ -0,0 +1,34 @@
...
+    private int firstLeg;
+    private int secondLeg;
+
+    public RightTriangle(int firstLeg, int secondLeg, String color) {
+        this.firstLeg = firstLeg;
+        this.secondLeg = secondLeg;
+        setColor(color);
+        setFigureType("right triangle");
+        calculateArea();
+    }
```
rovein, line 13: See my comments to the `Figure` class, firstly change them, and then you will need to change all the inheritors.

src/main/java/core/basesyntax/Square.java
```diff
@@ -0,0 +1,30 @@
...
+
+public class Square extends Figure {
+    private int side;
+
+    public Square(int side, String color) {
+        this.side = side;
+        setFigureType("square");
+        setColor(color);
+        calculateArea();
+    }
```
rovein, line 11: See my comments to the `Figure` class, firstly change them, and then you will need to change all the inheritors.

rovein (changes requested):
Good solution, but let's improve it ;)

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,81 @@
...
+            case 1:
+                figure = createSquare();
+                break;
+            case 2:
+
+                figure = createRectangle();
+                break;
+            case 3:
+
```
sarakhmen, line 28: same here

src/main/java/core/basesyntax/HelloWorld.java
sarakhmen: not fixed

src/main/java/core/basesyntax/IDraw.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface IDraw {
```
sarakhmen, line 3: Interface names could be adjectives and end with “able” or “ible” whenever the interface provides a capability; otherwise, they should be nouns.
Here I'd recommend you rename the interface to 'Drawable'

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -0,0 +1,31 @@
+package core.basesyntax;
+
+public class Rectangle extends Figure {
+    private int aside;
+    private int bside;
```
sarakhmen, line 5: I'd recommend renaming to 'width' and 'length'

sarakhmen (changes requested):
Overall good job, but let's improve your solution :)

rovein (approved):
Well done! Keep the rhythm :)

PR #1022, opened 2023-04-06
tasks completed

vojtcekhovski:


okuzan (changes requested):
You have typo in class name
image

src/main/java/core/basesyntax/Models.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum Models {
```
olekskov, line 3: It's recommended to use singular names for enum types. Rename `Models` to `Model` 🚀

src/main/java/core/basesyntax/AreaCalculator.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax;
+
+public interface AreaCalculator {
+    public abstract double getArea();
```
olekskov, line 4: Remove unnecessary 'public abstract' keywords, as they are implicit in interface methods. Simply use 'double getArea();'

src/main/java/core/basesyntax/Color.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
```
olekskov, line 2: You have an extra empty line after the package declaration. It is better to remove unnecessary empty lines for better code readability.

src/main/java/core/basesyntax/Draw.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax;
+
+public interface Draw {
+    public abstract void draw();
```
olekskov, line 4: Remove the redundant 'public' and 'abstract' keywords, as interface methods are implicitly public and abstract. Simply use 'void draw();' 🧹

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax;
+
+public class Main {
+    
+    public static void main(String[] args) {
+
+        Figure[] figures = new Figure[6];
```
olekskov, line 7: Consider using a constant for the size of the `figures` array to improve readability and maintainability. For example, `private static final int FIGURE_COUNT = 6;`

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.Random;
+
+public class FigureSupplier {
+    public static final int MAX_VALUE = 100;
+    public static final int DEFAULT_RADIUS = 10;
+    public static final Color DEFAULT_COLOR = Color.WHITE;
+
+    private final Random random = new Random();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private Models models;
```
olekskov, line 12: The `models` field should be a local variable inside `getRandomFigure()` method, as it's not used elsewhere in the class. 🛠️

src/main/java/core/basesyntax/IsoscelesTrapezoid.java
```diff
@@ -0,0 +1,27 @@
+package core.basesyntax;
+
+public class IsoscelesTrapezoid extends Figure {
+    private int top;
+    private int bottom;
+    private int high;
```
olekskov, line 6: Consider renaming the variable 'high' to 'height'. This would improve readability and better represent its purpose in the context of an isosceles trapezoid. 📏

src/main/java/core/basesyntax/IsoscelesTrapezoid.java
```diff
@@ -0,0 +1,27 @@
...
+    @Override
+    public double getArea() {
+        return ((top + bottom) / 2) * high;
+    }
+
+    @Override
+    public void draw() {
+        System.out.println("Figure: IsoscelesTrapezoid, Area: " + getArea()
+                + " sq. units, Top: " + top + " Bottom: " + bottom
+                + " High: " + high
```
olekskov, line 24: Replace 'High' with 'Height' in the output string to match the recommended variable name change. 🚀

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,69 @@
...
+            case ISOSCELES_TRAPEZOID:
+                return getRandomIsoscelesTrapezoid();
+            case RECTANGLE:
+                return getRandomRectangle();
+            case RIGHT_TRIANGLE:
+                return getRandomRightTriangle();
+            case SQUARE:
+                return getRandomSquare();
+            default:
+                return getRandomFigure();
```
olekskov, line 28: method getRandomFigure shouldn't return default figure, if you checked all figures except square then we know that square should be returned
```suggestion
            default:
                return getRandomSquare();
```

src/main/java/core/basesyntax/Draw.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax;
+
+public interface Draw {
```
olekskov, line 3: Action names are suitable for methods, interfaces naming is better as noun or adjective, e.g. Drawable

olekskov (changes requested):

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,67 @@
...
+            case CIRCLE:
+                return getRandomCircle();
+            case ISOSCELES_TRAPEZOID:
+                return getRandomIsoscelesTrapezoid();
+            case RECTANGLE:
+                return getRandomRectangle();
+            case RIGHT_TRIANGLE:
+                return getRandomRightTriangle();
+            default:
+                return getRandomSquare();
```
Ivan95kos, line 26:
```suggestion
                return getDefaultFigure();
```

Ivan95kos (approved):
Great job, with minor comments

PR #1037, opened 2023-04-19
Added the task solution:

igaryaushskaus:
1. Made an abstract class Figure with color and area fields.
2. Made an interface FigureAreaCalculator for calculating figure's area and printing info about the figure.
3. Made figure classes: Square, Circle, Rectangle, RightTriangle, IsoscelesTrapezoid which are heirs of the class Figure.
4. Made Color enum with figures colors.
5. Made class ColorSupplier with method getRandomColor() to get random colors for figures.
6. Made class FigureSupplier with methods getRandomFigure() and getDefaultFigure() to get figures.
7. In main class Main made an array of figures and fill them with random/default figures using methods in FigureSupplier class.

### DIFF
--- a/src/main/java/core/basesyntax/Figure.java
+++ b/src/main/java/core/basesyntax/Figure.java
@@ -1,0 +1,26 @@
+package core.basesyntax;
+
+public abstract class Figure implements FigureAreaCalculator {
+    private String color;
+    private double area;
+
+    public Figure(String color) {
+        this.color = color;
+    }
+
+    public String getColor() {
+        return color;
+    }
+
+    public void setColor(String color) {
+        this.color = color;
+    }
+
+    public double getArea() {
+        return area;
+    }
+
+    public void setArea(double area) {
+        this.area = area;
+    }
+}
--- a/src/main/java/core/basesyntax/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/FigureSupplier.java
@@ -1,0 +1,38 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int FIGURES_AMOUNT = 5;
+    private static final double MAGIC_NUMBER = 25;
+    private final Random random = new Random();
+    private final ColorSupplier supplier = new ColorSupplier();
+    private Figure figure;
+
+    public Figure getRandomFigure() {
+        switch (random.nextInt(FIGURES_AMOUNT)) {
+            case 1:
+                figure = new Square(supplier.getRandomColor(), MAGIC_NUMBER);
+                break;
+            case 2:
+                figure = new Circle(supplier.getRandomColor(), MAGIC_NUMBER);
+                break;
+            case 3:
+                figure = new RightTriangle(supplier.getRandomColor(), MAGIC_NUMBER, MAGIC_NUMBER);
+                break;
+            case 4:
+                figure = new IsoscelesTrapezoid(supplier.getRandomColor(), MAGIC_NUMBER, MAGIC_NUMBER, MAGIC_NUMBER);
+                break;
+            case 5:
+                figure = new Rectangle(supplier.getRandomColor(), MAGIC_NUMBER, MAGIC_NUMBER);
+                break;
+            default:
+                break;
+        }
+        return figure;
+    }
+
+    public Figure getDefaultFigure() {
+        return figure = new Circle("white", 10.0);
+    }
+}
--- a/src/main/java/core/basesyntax/Main.java
+++ b/src/main/java/core/basesyntax/Main.java
@@ -1,0 +1,20 @@
+package core.basesyntax;
+
+public class Main {
+    private static final int MAGIC_NUMBER = 6;
+
+    public static void main(String[] args) {
+        Figure[] figuresArray = new Figure[MAGIC_NUMBER];
+        FigureSupplier supplier = new FigureSupplier();
+
+        for (int i = 0; i < 2; i++) {
+            figuresArray[i] = supplier.getRandomFigure();
+        }
+        for (int i = 3; i < 5; i++) {
+            figuresArray[i] = supplier.getDefaultFigure();
+        }
+        for (Figure figure : figuresArray) {
+            figure.printInfo();
+        }
+    }
+}
--- a/src/main/java/core/basesyntax/Circle.java
+++ b/src/main/java/core/basesyntax/Circle.java
@@ -1,0 +1,16 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+    private double radius;
+
+    public Circle(String color, double radius) {
+        super(color);
+        this.radius = radius;
+    }
+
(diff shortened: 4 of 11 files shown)


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,39 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int FIGURES_AMOUNT = 5;
+    private static final double MAGIC_NUMBER = 25;
```
uliana-andrushchuk, line 7: Please do not use MAGIC_NUMBER for a constant. 
Variable names should describe what the variable is used for.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,39 @@
...
+                figure = new Rectangle(supplier.getRandomColor(), MAGIC_NUMBER, MAGIC_NUMBER);
+                break;
+            default:
+                break;
+        }
+        return figure;
+    }
+
+    public Figure getDefaultFigure() {
+        return figure = new Circle("white", 10.0);
```
uliana-andrushchuk, line 37: Please, make "white" and 10.0 constants
Use enum for "white" instead of String

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax;
+
+public class Main {
+    private static final int MAGIC_NUMBER = 6;
+
+    public static void main(String[] args) {
+        Figure[] figuresArray = new Figure[MAGIC_NUMBER];
+        FigureSupplier supplier = new FigureSupplier();
+
+        for (int i = 0; i < 2; i++) {
```
uliana-andrushchuk, line 10: You need to iterate through the entire array, but with an additional check (if this first half of array) inside the loop.

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,26 @@
+package core.basesyntax;
+
+public abstract class Figure implements FigureAreaCalculator {
+    private String color;
+    private double area;
```
uliana-andrushchuk, line 5: you don't need to have field `area` in Figure class

uliana-andrushchuk (changes requested):

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,39 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int FIGURES_AMOUNT = 5;
+    private static final double MAGIC_NUMBER = 25;
+    private final Random random = new Random();
+    private final ColorSupplier supplier = new ColorSupplier();
+    private Figure figure;
```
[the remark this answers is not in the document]
igaryaushskaus: which field?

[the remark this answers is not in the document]
kozhukhovsky: figure

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,43 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int FIGURES_AMOUNT = 5;
+    private static final double RANDOM_PROPERTIES = 25;
+    private static final double DEFAULT_CIRCLE_RADIUS = 10;
```
kozhukhovsky, line 8: You won't get a random value, because we have it fixed and goes to all figures.
The value with the sides of the figure should be obtained randomly through Random

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,20 @@
...
+
+    public static void main(String[] args) {
+        Figure[] figuresArray = new Figure[ARRAY_LENGTH];
+        FigureSupplier supplier = new FigureSupplier();
+
+        for (int i = 0; i < figuresArray.length; i++) {
+            if (i <= figuresArray.length / 2) {
+                figuresArray[i] = supplier.getRandomFigure();
+            }
+            figuresArray[i] = supplier.getDefaultFigure();
```
kozhukhovsky, line 14:
```suggestion
            if (i <= figuresArray.length / 2) {
                figuresArray[i] = supplier.getRandomFigure();
            } else {
                figuresArray[i] = supplier.getDefaultFigure();
            }

```

Otherwise, the figure will be overwritten each time, since it will reach here anyway. And we need either or

src/main/java/core/basesyntax/AreaCalculator.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface AreaCalculator {
+    double areaCalculator();
```
kozhukhovsky, line 4:
```suggestion
    double calculateArea();
```

Methods should answer the question "what to do", i.e. must be a verb

kozhukhovsky (changes requested):

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,21 @@
...
+        for (int i = 0; i < figuresArray.length; i++) {
+            if (i <= figuresArray.length / 2) {
+                figuresArray[i] = supplier.getRandomFigure();
+            } else {
+                figuresArray[i] = supplier.getDefaultFigure();
+            }
+        }
+        for (Figure figure : figuresArray) {
+            figure.printInfo();
+        }
```
okuzan, line 19: move this to previos for loop

src/main/java/core/basesyntax/FigurePrintInfo.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface FigurePrintInfo {
+    void printInfo();
+}
```
okuzan, line 5:
```suggestion
public interface Drawable {
    void draw();
}
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,42 @@
...
+                        random.nextDouble() * RANDOM_PROPERTIES);
+            case 3:
+                return new IsoscelesTrapezoid(supplier.getRandomColor(),
+                        random.nextDouble() * RANDOM_PROPERTIES,
+                        random.nextDouble() * RANDOM_PROPERTIES,
+                        random.nextDouble() * RANDOM_PROPERTIES);
+            case 4:
+                return new Rectangle(supplier.getRandomColor(),
+                        random.nextDouble() * RANDOM_PROPERTIES,
+                        random.nextDouble() * RANDOM_PROPERTIES);
```
okuzan, line 33: Could generate 0, and figures can have 0 property

okuzan (changes requested):

PR #1039, opened 2023-04-23
implemented task

chrisnyrka:


src/main/java/core/basesyntax/FiguresName.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum FiguresName {
+    CIRCLE,
+    SQUARE,
+    RIGHT_TRIANGLE,
+    RECTANGLE,
+    ISOSCELES_TRAPEZOID
+}
```
rovein, line 9: What if we will create another `Figure` subclass? Will we always update this enum accordingly?

Let's try not to use it. See comments tips in `FigureSupplier`

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private final Random randomColor = new Random();
```
rovein, line 6: Is it `randomColor`, or just `random`?

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
...
+
+    public String getRandomColor() {
+        int randomNumberColor = randomColor.nextInt(Color.values().length);
```
rovein, line 9: Let it be just 
```suggestion
        int randomColorIndex = randomColor.nextInt(Color.values().length);
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,46 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final String DEFAULT_COLOR = "WHITE";
+    private static final int MAX_RANGE = 20;
+    private static final int DEFAULT_RADIUS = 10;
+    private Random random = new Random();
```
rovein, line 9: Let it also be `final`

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,46 @@
...
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        int randomNameFigure = new Random().nextInt(FiguresName.values().length);
```
rovein, line 13: Don't create `new Random()` here, you already have a `random` field.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,46 @@
...
+        switch (FiguresName.values()[randomNameFigure]) {
+            case CIRCLE:
```
rovein, line 15: Instead of using `FigureName.values()`, you can just generate a random number from 0 to figures count. 

```suggestion
        switch (randomFigureNumber) {
            case 0:
             ....
            case 1:
             ....
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,46 @@
...
+                Figure circle = new Circle(colorSupplier.getRandomColor(),
+                        random.nextInt(MAX_RANGE) + 1);
+                return circle;
```
rovein, line 18: Don't pass expressions to the constructor. Also, inline the variable.
```suggestion
                int radius = ...;
                return new Circle(color, radius);
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,46 @@
...
+            case SQUARE:
+                Figure square = new Square(colorSupplier.getRandomColor(),
```
rovein, line 20: `colorSupplier.getRandomColor()` can be moved to a variable before the `switch`

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,46 @@
...
+                        random.nextInt(MAX_RANGE) + 1,
+                        random.nextInt(MAX_RANGE) + 1);
+                return rectangle;
+            default:
+                Figure isoscelesTrapezoid = new IsoscelesTrapezoid(colorSupplier.getRandomColor(),
+                        random.nextInt(MAX_RANGE) + 1,
+                        random.nextInt(MAX_RANGE) + 1,
+                        random.nextInt(MAX_RANGE) + 1);
+                return isoscelesTrapezoid;
+        }
```
rovein, line 39: The same for all cases as in previous comments.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    public Figure getDefaultFigure() {
+        Figure circle = new Circle(DEFAULT_COLOR,DEFAULT_RADIUS);
+        return circle;
```
rovein, line 44:
```suggestion
        return new Circle(DEFAULT_COLOR,DEFAULT_RADIUS);
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,19 @@
...
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] figures = new Figure[SIZE_ARRAY];
+        for (int i = 0; i < SIZE_ARRAY; i++) {
+            if (i < SIZE_ARRAY / 2) {
+                figures[i] = figureSupplier.getRandomFigure();
+                System.out.println(figures[i].printFigure());
+            } else {
+                figures[i] = figureSupplier.getDefaultFigure();
+                System.out.println(figures[i].printFigure());
```
rovein, line 15: It can be moved after `if/else` with removed duplicate at line 12.

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -0,0 +1,25 @@
+package core.basesyntax;
+
+public class Rectangle extends Figure {
+    private int firstSide;
+    private int secondSide;
```
rovein, line 5: Let it be `width` and `length`.

src/main/java/core/basesyntax/FigureBehaviour.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface FigureBehaviour {
```
rovein, line 3: Let's rename it to `AreaCalculator`

src/main/java/core/basesyntax/FigureBehaviour.java
```diff
@@ -0,0 +1,5 @@
...
+    double getArea();
```
rovein, line 4: And it's to `calculateArea`

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,15 @@
...
+
+    public String getColor() {
+        return color;
+    }
+
+    public void setColor(String color) {
+        this.color = color;
+    }
+
+    public abstract String printFigure();
```
rovein, line 14: Create this method also in separate interface `Drawable`, and add  `implements Drawable` in Figure.

Also, let it be just draw()
```suggestion
    public abstract String draw();
```

rovein (changes requested):
Good job! But let's improve your solution ;)

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,25 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+    private int radius;
+
+    public Circle(String color, int radius) {
+        super.setColor(color);
```
maxline, line 7: use constructor `super(color);` instead of setter here. fix the same in all figures classes.

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,25 @@
...
+        this.radius = radius;
+    }
+
+    public Circle() {
```
maxline, line 11: do we need a constructor without parameters? fix in all figures classes please.

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public abstract class Figure implements AreaCalculator, Drawable {
+    private String color;
```
maxline, line 4: lets use existing `enum Color` for this field instead of the `String` type.

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,13 @@
...
+
+    public String getColor() {
```
maxline, line 6: lets use existing `enum Color` for this getter instead of the `String` type.

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,13 @@
...
+        return color;
+    }
+
+    public void setColor(String color) {
```
maxline, line 10: the same - use `enum Color`

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final String DEFAULT_COLOR = "WHITE";
+    private static final int DEFAULT_RADIUS = 10;
+    private static final int MAX_RANGE = 20;
+    private final Random random = new Random();
+    private final ColorSupplier colorSupplier = new ColorSupplier(random);
+    private final Figure[] figures = {new Circle(), new Square(), new RightTriangle(),
```
maxline, line 11: this array `figures`  is redundant

src/main/java/core/basesyntax/FigureSupplier.java
maxline: not fixed

src/main/java/core/basesyntax/FigureSupplier.java
maxline: not fixed

maxline (changes requested):
good job! let's improve the solution

src/main/java/core/basesyntax/Circle.java
chrisnyrka: In the new solution I will need empty constructors with no parameters, since you said to declare new objects in class fields and not in methods

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,53 @@
...
+public class FigureSupplier {
+    private static final Color DEFAULT_COLOR = Color.WHITE;
+    private static final int DEFAULT_RADIUS = 10;
+    private static final int MAX_RANGE = 20;
+    private final Random random = new Random();
+    private final Circle circle = new Circle();
+    private final Square square = new Square();
+    private final RightTriangle rightTriangle = new RightTriangle();
+    private final Rectangle rectangle = new Rectangle();
+    private final IsoscelesTrapezoid isoscelesTrapezoid = new IsoscelesTrapezoid();
```
rovein, line 14: Don't create them here. See the example below.

You will always return a reference to one single object. Instead, you need a new object every time you call "getRandomFigure".

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,53 @@
...
+    private final ColorSupplier colorSupplier = new ColorSupplier(random);
+    private final Figure[] figures = {circle, square, rightTriangle, rectangle, isoscelesTrapezoid};
```
rovein, line 16: Let's remove it, see the next comments.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public Figure getRandomFigure() {
+        int randomNameFigure = random.nextInt(figures.length);
```
rovein, line 19: Here is the only place where you need your `figures`. Instead, you can do next:
```suggestion
        int randomNameFigure = random.nextInt(FIGURES_COUNT);
```

Declate `FIGURES_COUNT` as a constant an assign there a corresponding value.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,53 @@
...
+        switch (randomNameFigure) {
+            case 0:
+                circle.setColor(colorSupplier.getRandomColor());
+                circle.setRadius(random.nextInt(MAX_RANGE) + 1);
+                return circle;
```
rovein, line 24:
```suggestion
                Circle circle = new Circle();
                circle.setColor(colorSupplier.getRandomColor());
                circle.setRadius(random.nextInt(MAX_RANGE) + 1);
                return circle;
```

Applicable for all cases. So move from class-level declaration here.

rovein (changes requested):
Great! But let's make final improvements in FigureSupplier.

rovein (approved):
Great! Thanks for the patience :)

PR #1043, opened 2023-04-23
added classes, interface and enum to create random and default figures

kvjohnny:


src/main/java/core/basesyntax/RightTriangle.java
```diff
@@ -0,0 +1,26 @@
+package core.basesyntax;
+
+public class RightTriangle extends Figure {
+    private final int firstLeg;
+    private final double secondLeg;
```
olekskov, line 5: It would be better to use the same data type for both legs (either both int or both double). 🧐

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    public String getRandomColor() {
```
olekskov, line 6: It's better to use an instance of `Random` instead of creating a new one every time. Consider creating a private final field and using it in the `getRandomColor()` method. 🌈

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,18 @@
...
+
+public class Main {
+    private static final int SIZE_OF_ARRAY = 6;
+
+    public static void main(String[] args) {
+        Figure[] figures = new Figure[SIZE_OF_ARRAY];
+        for (int i = 0; i < SIZE_OF_ARRAY; i++) {
+            if (i < 3) {
+                figures[i] = new FigureSupplier().getRandomFigure();
+                System.out.println();
```
olekskov, line 11: Remove the empty `System.out.println();` since it is not printing any information. 🧹

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,18 @@
...
+            } else {
+                figures[i] = new FigureSupplier().getDefaultFigure();
+            }
+        }
```
olekskov, line 15: Use `System.out.println(figures[i]);` to print each figure. 🖨️

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,24 @@
...
+
+    public String getColor() {
+        return color;
+    }
+
+    public void setColor(String color) {
+        this.color = color;
+    }
+
+    public double getArea() {
```
olekskov, line 18: getArea() method should be abstract, as it should be implemented by each figure type.

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,24 @@
+package core.basesyntax;
+
+public class Figure implements Figureable {
```
olekskov, line 3: Figure class should be an abstract class since it's not meant to be instantiated directly.

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,24 @@
...
+
+    public void setColor(String color) {
+        this.color = color;
+    }
+
+    public double getArea() {
+        return 0;
+    }
+
+    public void draw() {
```
olekskov, line 22: draw() method should be abstract, as it should be implemented by each figure type.

src/main/java/core/basesyntax/IsoscelesTrapezoid.java
```diff
@@ -0,0 +1,30 @@
+package core.basesyntax;
+
+public class IsoscelesTrapezoid extends Figure {
+    private final double firstParallel;
+    private final double secondParallel;
+    private final int height;
```
olekskov, line 6: For consistency, consider using 'double' instead of 'int' for the height, since the other dimensions are represented as doubles.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,36 @@
...
+    private final Figure square = new Square(colorSupplier.getRandomColor(),
+            random.nextInt(MAX_NUM));
+    private final Figure rectangle = new Rectangle(colorSupplier.getRandomColor(),
+            random.nextInt(MAX_NUM), random.nextInt(MAX_NUM));
+    private final Figure rightTriangle = new RightTriangle(colorSupplier.getRandomColor(),
+            random.nextInt(MAX_NUM), random.nextDouble());
+    private final Figure isoscelesTrapezoid = new IsoscelesTrapezoid(colorSupplier.getRandomColor(),
+            random.nextDouble(), random.nextDouble(), random.nextInt(MAX_NUM));
+
+    public Figure getRandomFigure() {
```
olekskov, line 23: The `getRandomFigure` method should return a random figure, not just draw them. You can use a switch statement to randomly choose a figure type and then return the corresponding instance.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,36 @@
...
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int RADIUS = 10;
+    private static final int MAX_NUM = 20;
+
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final Random random = new Random();
+    private final Figure circle = new Circle(colorSupplier.getRandomColor(),
```
olekskov, line 11: It's better to create random figures in the `getRandomFigure` method rather than initializing them as instance variables. This will allow you to generate different figures each time the method is called.

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,23 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+    private final double radius;
+
+    public Circle(String color, int radius) {
```
olekskov, line 6: The type of `radius` parameter in the constructor should be `double` instead of `int`. Change the parameter type to `double` for correct calculations. 🧮

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,36 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int RADIUS = 10;
+    private static final int MAX_NUM = 20;
```
olekskov, line 7:
```suggestion
    private static final int DEFAULT_RADIUS = 10;
    private static final int MAX_BALL_NUMBER = 20;
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,18 @@
+package core.basesyntax;
+
+public class Main {
+    private static final int SIZE_OF_ARRAY = 6;
+
+    public static void main(String[] args) {
+        Figure[] figures = new Figure[SIZE_OF_ARRAY];
+        for (int i = 0; i < SIZE_OF_ARRAY; i++) {
+            if (i < 3) {
+                figures[i] = new FigureSupplier().getRandomFigure();
```
olekskov, line 10: Create an instance of FigureSupplier class outside the loop to avoid unnecessary object creation. 🛠️

olekskov (changes requested):

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,19 @@
...
+    public static void main(String[] args) {
+        Figure[] figures = new Figure[ARRAY_LENGTH];
+        FigureSupplier figureSupplier = new FigureSupplier();
+        for (int i = 0; i < ARRAY_LENGTH; i++) {
+            if (i < 3) {
+                figures[i] = figureSupplier.getRandomFigure();
+                System.out.println(figures[i]);
+            } else {
+                figures[i] = figureSupplier.getDefaultFigure();
+                System.out.println(figures[i]);
```
okuzan, line 15: extract common part from if-else

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,55 @@
...
+                rectangle.draw();
+                break;
+            case 3:
+                rightTriangle.draw();
+                break;
+            case 4:
+                isoscelesTrapezoid.draw();
+                break;
+            default:
+                break;
```
okuzan, line 41: in switch don't draw, return directly. This method isn't about drawing. it just gets random figure, as stated in its name. The user of the class then does what they want with it - draw or something else. To make it more readable, move the logic of creating figures right into cases, or extract it to dedicated private methods

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,55 @@
...
+        }
+        return new Figure(colorSupplier.getRandomColor()) {
+            @Override
+            public double getArea() {
+                return 0;
+            }
+        };
```
okuzan, line 48: do you understand what you are doing here? :) You are returning the dummy anonymous class, which extends the abstract class Figure. It has color, but doesn't have area, and it's not Square, circle or anything else

src/main/java/core/basesyntax/Figureable.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax;
+
+public interface Figureable {
```
okuzan, line 3:
```suggestion
public interface AreaCalculator {
```

src/main/java/core/basesyntax/Figureable.java
```diff
@@ -0,0 +1,6 @@
...
+    double getArea();
+
```
okuzan, line 5:
```suggestion
```

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax;
+
+public abstract class Figure implements Figureable {
```
okuzan, line 3: create interface Drawable and make Figure implement it

okuzan, line 3: all you need for Figure class if a field, constructor and getter. We don't need to explicitly add here methods of the interfaces, it's enough to just `implement` them. Our figures will override them

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private final Random random = new Random();
+
+    public String getRandomColor() {
+        return Color.values()[random.nextInt(Color.values().length)].toString();
```
okuzan, line 9:
```suggestion
        return Color.values()[random.nextInt(Color.values().length)].name();
```

okuzan (changes requested):

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,41 @@
...
+
+public class FigureSupplier {
+    private static final int DEFAULT_RADIUS = 10;
+    private static final int MAX_LENGTH_OF_SEGMENT = 20;
+    private static final int FIGURE_COUNT = 5;
+
+    public Figure getRandomFigure() {
+        final Random random = new Random();
+        final int figureNumber = random.nextInt(FIGURE_COUNT);
+        final ColorSupplier colorSupplier = new ColorSupplier();
```
uliana-andrushchuk, line 13: Please, make `final Random random = new Random();` and `final ColorSupplier colorSupplier = new ColorSupplier();` class level fields
Don't forget to add `private`

src/main/java/core/basesyntax/Square.java
```diff
@@ -0,0 +1,24 @@
...
+        return side * side;
+    }
+
+    @Override
+    public void draw() {
+        System.out.println("Figure: square"
+                + ", area: " + getArea()
+                + " sq.units, side: " + side
+                + " units, color: " + getColor());
+
```
uliana-andrushchuk, line 22:
```suggestion
```

uliana-andrushchuk (changes requested):

olekskov (approved):
Good job

PR #1046, opened 2023-04-24
Hw 4 solution

AnastasiiaVariuchuk:


src/main/java/core/basesyntax/IDraw.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface IDraw {
```
sarakhmen, line 3: Interface names shouldn't be verbs. I'd recommend renaming it to Drawable or something like this

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,18 @@
+package core.basesyntax;
+
+public class Main {
+
```
sarakhmen, line 4: redundant empty line

sarakhmen (approved):
Overall very good job, you can improve your solution if you want :)

PR #1050, opened 2023-04-24
created classes for all geometric figures, abstract class Figure, int…

ViktoriiaSemenchenko:
…erfaces Printable and AreaCalculator

implemented methods for random color and geometric figures generation

### DIFF
--- a/src/main/java/core/basesyntax/Circle.java
+++ b/src/main/java/core/basesyntax/Circle.java
@@ -1,0 +1,32 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+    public static final int MAX_RADIUS = 20;
+    public static final int DEFAULT_RADIUS = 10;
+    private int radius;
+
+    public Circle(Color color, int radius) {
+        super(color);
+        this.radius = radius;
+    }
+
+    public int getRadius() {
+        return radius;
+    }
+
+    public void setRadius(int radius) {
+        this.radius = radius;
+    }
+
+    @Override
+    public double getArea() {
+        return Math.PI * radius * radius;
+    }
+
+    @Override
+    public void printFigureInfo() {
+        System.out.println("Figure: circle, area: " + getArea()
+                + " sq.units, radius: " + getRadius()
+                + " units, color: " + getColor());
+    }
+}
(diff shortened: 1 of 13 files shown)


src/main/java/core/basesyntax/Printable.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Printable {
+    void printFigureInfo();
+}
```
rovein, line 5: Let it be `Drawable` and `draw` correspondingly as mentioned in the requirements. (... able to draw it. To 'draw' means ...).

src/main/java/core/basesyntax/Color.java
```diff
@@ -0,0 +1,12 @@
...
+
+public enum Color {
+    RED,
+    ORANGE,
+    YELLOW,
+    GREEN,
+    BLUE,
+    BLACK,
+    WHITE
+
```
rovein, line 11: Let's remove redundant empty line.

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,32 @@
...
+
+    @Override
+    public double getArea() {
+        return Math.PI * radius * radius;
+    }
+
+    @Override
+    public void printFigureInfo() {
+        System.out.println("Figure: circle, area: " + getArea()
+                + " sq.units, radius: " + getRadius()
```
rovein, line 29: Don't use getters when you have direct access to the fields. See all your figures.

```suggestion
                + " sq.units, radius: " + radius
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,19 @@
...
+        Figure[] figures = new Figure[NUMBER_OF_FIGURES];
+        for (int i = 0; i < figures.length / 2; i++) {
+            figures[i] = figureSupplier.getRandomFigure();
+        }
+        for (int i = figures.length / 2; i < NUMBER_OF_FIGURES; i++) {
+            figures[i] = figureSupplier.getDefaultFigure();
+        }
+        for (Figure figure : figures) {
+            figure.printFigureInfo();
+        }
```
rovein, line 17: Do we actually need a separate loop? Could it affect the performance, if we had an array of billions of elements?

Let's print figures in the previous loops after their creation.

rovein (changes requested):
Great job! But let's remove those minor comments ;)

rovein (approved):
Great job :)

PR #1066, opened 2023-04-25
implemented task

mpankrad:
implemented task

### DIFF
--- a/src/main/java/core/basesyntax/Circle.java
+++ b/src/main/java/core/basesyntax/Circle.java
@@ -1,0 +1,31 @@
+package core.basesyntax;
+
+public class Circle extends Shape {
+    private final double radius;
+
+    public Circle(double radius, String color) {
+        super(color);
+        this.radius = radius;
+    }
+
+    @Override
+    public double getArea() {
+        return Math.PI * radius * radius;
+    }
+
+    @Override
+    public String toString() {
+        return "Circle{"
+                + "radius="
+                + radius + ", color='"
+                + getColor() + '\''
+                + '}';
+    }
+
+    @Override
+    public void draw() {
+        System.out.println("Figure: circle, area: " + getArea()
+                + " radius: " + radius
+                + " color: " + getColor());
+    }
+}
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private Random random;
+
+    public ColorSupplier() {
+        this.random = new Random();
+    }
+
+    public String getRandomColor() {
+        return String.valueOf(Color.values()[random.nextInt(Color.values().length)]);
+    }
+}
--- a/src/main/java/core/basesyntax/Figure.java
+++ b/src/main/java/core/basesyntax/Figure.java
@@ -1,0 +1,5 @@
+package core.basesyntax;
+
+interface Figure {
+    void draw();
+}
--- a/src/main/java/core/basesyntax/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/FigureSupplier.java
@@ -1,0 +1,66 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static Random random;
+    private ColorSupplier colorSupplier;
+
+    public FigureSupplier(Random random, ColorSupplier colorSupplier) {
+        this.random = random;
+        this.colorSupplier = colorSupplier;
+    }
+
+    public Figure getRandomFigure() {
+        String color = colorSupplier.getRandomColor();
+        String[] figureTypes = {"Square", "Rectangle",
+                "Right Rectangle", "Circle", "Isosceles Trapezoid"};
+        int figureTypeIndex = random.nextInt(figureTypes.length);
+        String figureType = figureTypes[figureTypeIndex];
+        return createFigure(figureType, random, colorSupplier);
+    }
+
+    private static Figure createFigure(String figureType,
+                                       Random random,
+                                       ColorSupplier colorSupplier) {
+        Figure figure;
+        switch (figureType) {
+            case "Square":
+                double side = random.nextDouble() * 10 + 1;
+                figure = new Square(side, colorSupplier.getRandomColor());
+                break;
+            case "Rectangle":
+                double length = random.nextDouble() * 10;
+                double width = random.nextDouble() * 10;
+                figure = new Rectangle(length, width, colorSupplier.getRandomColor());
+                break;
+            case "Right Rectangle":
+                double firstleg = random.nextDouble() * 10;
+                double secondleg = random.nextDouble() * 10;
+                figure = new RightTriangle(firstleg, secondleg, colorSupplier.getRandomColor());
+                break;
+            case "Circle":
+                double radius = random.nextDouble() * 10;
+                figure = new Circle(radius, colorSupplier.getRandomColor());
+                break;
+            case "Isosceles Trapezoid":
+                double upperSide =
+                        random.nextDouble() * 10;
+                double lowerSide =
+                        random.nextDouble() * 10;
+                double height =
+                        random.nextDouble() * 10;
+                figure = new IsoscelesTrapezoid(upperSide,
+                        lowerSide, height,
+                        colorSupplier.getRandomColor());
+                break;
+            default:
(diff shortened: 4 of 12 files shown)


src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+interface Figure {
```
rovein, line 3: Let's rename it to `Drawable`, because it can draw itself.

src/main/java/core/basesyntax/Shape.java
```diff
@@ -0,0 +1,21 @@
+package core.basesyntax;
+
+public abstract class Shape implements Figure {
```
rovein, line 3: Let's rename `Shape` to `Figure`, and also see comment in `Figure` file.

src/main/java/core/basesyntax/Shape.java
```diff
@@ -0,0 +1,21 @@
...
+
+    public String getColor() {
+        return color;
+    }
+
+    public void setColor(String color) {
+        this.color = color;
+    }
+
+    public abstract double getArea();
```
rovein, line 18: Don’t use abstract classes to set behaviour for classes. Also, don’t put all behaviour into a single interface if the methods are conceptually different from each other. 

So, let's create another interface `AreaCalculator`, where you will define the method `getArea()`. Don't forget to add this new interface to `implements` statement in this class.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,15 @@
...
+
+public class ColorSupplier {
+    private Random random;
+
+    public ColorSupplier() {
+        this.random = new Random();
+    }
+
+    public String getRandomColor() {
+        return String.valueOf(Color.values()[random.nextInt(Color.values().length)]);
```
rovein, line 13: Use name() for getting String representation of enum constants. Don't use `toString()` or `valueOf()`.
```suggestion
        int index = random.nextInt(Color.values().length);
        return Color.values()[index].name();
```

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return "Circle{"
+                + "radius="
+                + radius + ", color='"
+                + getColor() + '\''
+                + '}';
+    }
```
rovein, line 23: From requirements: "you shouldn't override toString() method for this". Let's completely remove it here and in all other figures. You already have the method `draw()`.

src/main/java/core/basesyntax/IsoscelesTrapezoid.java
```diff
@@ -0,0 +1,39 @@
+package core.basesyntax;
+
+public class IsoscelesTrapezoid extends Shape {
+
```
rovein, line 4: Let's remove redundant empty line.

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -0,0 +1,35 @@
+package core.basesyntax;
+
+public class Rectangle extends Shape {
+
```
rovein, line 4: Let's remove redundant empty line. Applicable for all other figures.

src/main/java/core/basesyntax/Square.java
```diff
@@ -0,0 +1,31 @@
+package core.basesyntax;
+
+public class Square extends Shape {
+    private double side;
+
+    public Square(double side, String color) {
+        super(color);
+        this.side = side;
+
```
rovein, line 9: Let's remove redundant empty line.

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,21 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Main {
+    public static void main(String[] args) {
+        Figure[] figures = new Figure[6];
```
rovein, line 7: Move magic number `6` to the constant.

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,21 @@
...
+        Random random = new Random();
+        ColorSupplier colorSupplier = new ColorSupplier();
+        FigureSupplier figureSupplier = new FigureSupplier(random, colorSupplier);
+        for (int i = 0; i < 5; i++) {
+            figures[i] = figureSupplier.getRandomFigure();
+        }
+        for (int i = 5; i < 6; i++) {
+            figures[i] = figureSupplier.getDefaultFigure();
+        }
```
rovein, line 16: The first half of figures in this array should be generated with random parameters. From requirements. So change a little bit the conditions of your loops.

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,21 @@
...
+        for (Figure figure : figures) {
+            figure.draw();
+        }
```
rovein, line 19: Do we actually need a separate loop? Could it affect the performance, if we had an array of billions of elements?

Let's print figures in the previous loops after their creation.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,66 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static Random random;
```
rovein, line 6: Remove `static` here.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,66 @@
...
+                        colorSupplier.getRandomColor());
+                break;
+            default:
+                throw new RuntimeException("Invalid figure type");
+        }
+        return figure;
+    }
+
+    public static Figure getDefaultFigure() {
+        return (Figure) new Circle(10, "white");
```
rovein, line 64: You don't need to cast here. Also, move magic number `10` to the constant. And "white" string also.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    private static Figure createFigure(String figureType,
+                                       Random random,
+                                       ColorSupplier colorSupplier) {
+        Figure figure;
+        switch (figureType) {
+            case "Square":
+                double side = random.nextDouble() * 10 + 1;
+                figure = new Square(side, colorSupplier.getRandomColor());
```
rovein, line 30: Let's create a `randomColor` (colorSupplier.getRandomColor()) variable before the `switch` and use it in all cases.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,66 @@
...
+
+    public FigureSupplier(Random random, ColorSupplier colorSupplier) {
+        this.random = random;
+        this.colorSupplier = colorSupplier;
+    }
+
+    public Figure getRandomFigure() {
+        String color = colorSupplier.getRandomColor();
+        String[] figureTypes = {"Square", "Rectangle",
+                "Right Rectangle", "Circle", "Isosceles Trapezoid"};
```
rovein, line 17: Let's instead create a `FIGURES_COUNT` constant, where we would store value 5, and then we will generate a random number from 0 to 5 and check it in case statements. See next comment.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,66 @@
...
+            case "Rectangle":
+                double length = random.nextDouble() * 10;
+                double width = random.nextDouble() * 10;
+                figure = new Rectangle(length, width, colorSupplier.getRandomColor());
+                break;
+            case "Right Rectangle":
+                double firstleg = random.nextDouble() * 10;
+                double secondleg = random.nextDouble() * 10;
+                figure = new RightTriangle(firstleg, secondleg, colorSupplier.getRandomColor());
+                break;
```
rovein, line 41:
```suggestion
            case 1:
                double length = random.nextDouble() * 10;
                double width = random.nextDouble() * 10;
                figure = new Rectangle(length, width, colorSupplier.getRandomColor());
                break;
            case 2:
                double firstleg = random.nextDouble() * 10;
                double secondleg = random.nextDouble() * 10;
                figure = new RightTriangle(firstleg, secondleg, colorSupplier.getRandomColor());
                break;
```

rovein (changes requested):
Good job! But let's improve your solution ;)

rovein (approved):
Good job! Approved.

PR #1104, opened 2023-06-07
Hw05 oop advance

Stelmaschyk:


olekskov (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    public String getRandomColor() {
+        int index = new Random().nextInt(Color.values().length);
```
overpathz, line 7: When you write new ..() -- memory is allocated.
Each method invocation you will be creating a new object, that performs same operations.
This object should be final field.

src/main/java/core/basesyntax/CountSquare.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax;
+
+public interface CountSquare {
+
```
overpathz, line 4: Extra empty line

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,59 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier implements NumberSupplier {
+    private static final int RADIUS_CIRCLE = 10;
+    private static final String DEFAULT_COLOR = Color.BLACK.toString();
+    private static final int AMOUNT_OF_FIGURE_TYPE = 5;
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
```
overpathz, line 10: Make this fields final since you won't change state of this objects

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,59 @@
...
+            case 3:
+                return getRandomRightTriangle();
+            case 4:
+            default:
+                return getRandomSquare();
+        }
+    }
+    
+    public int getRandomIntNumber() {
+        return random.nextInt(20);
```
overpathz, line 29: What is 20? Magic number :)
Make it constant field.

Also, you have special separate method for getting random number.
Why wouldn't to create the same, but for the color?

overpathz (changes requested):
Good work. A little fixes required

src/main/java/core/basesyntax/Drawable.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Drawable {
+    void print();
```
```suggestion
    void draw();
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+public class Main {
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] arrayOfFigure = new Figure[6];
+        for (int i = 0; i < arrayOfFigure.length; i++) {
+            if (i < arrayOfFigure.length / 2) {
+                arrayOfFigure[i] = figureSupplier.getRandomFigure();
+                arrayOfFigure[i].print();
```
okuzan, line 10: common part in if-else, extract

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,66 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier implements NumberSupplier, ColorSupplier {
+    private static final int RADIUS_CIRCLE = 10;
+    private static final String DEFAULT_COLOR = Color.BLACK.toString();
```
```suggestion
    private static final String DEFAULT_COLOR = Color.BLACK.name();
```

src/main/java/core/basesyntax/Drawable.java
okuzan: not explicitly required in task description, but would make sense to name the only method of Drawable - draw

okuzan (changes requested):
You mustn't make changes to test files. Please undo them before requesting a review

aaananas (approved):

PR #1112, opened 2023-06-14
I created an array of figures and "painted" them

mhorianin:


src/main/java/core/basesyntax/Drawable.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Drawable {
+    String getDraw();
```
```suggestion
    String draw();

```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int LIMIT = 70;
+    private static final int DEFAULT_RADIUS = 10;
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final Random random = new Random();
+
+    public Figure getRandomFigure() {
+        int numberFigure = random.nextInt(7);
```
okuzan, line 12: this returns numbers from 0 to 6. Numerate your cases from 0 to 4, and randomize with limit of 5 instead

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,62 @@
...
+            case 2:
+                return prepareRectangle();
+            case 3:
+                return prepareCircle();
+            case 4:
+                return prepareIsoscelesTrapezoid();
+            case 5:
+                return prepareRightTriangle();
+            default:
+                return getDefaultFigure();
```
okuzan, line 25: better throw an exception, here it's not intended behavior, we should always be in the valid range

okuzan (changes requested):
Good job on your first try!

aaananas (approved):

PR #1116, opened 2023-06-17
work done

mate-academy-student-2023:
Please review.
Thank you!

### DIFF
--- a/src/main/java/core/basesyntax/Figure.java
+++ b/src/main/java/core/basesyntax/Figure.java
@@ -1,0 +1,31 @@
+package core.basesyntax;
+
+public abstract class Figure implements Drawable {
+    public static final int MAX_SIZE = 100;
+    private Color color;
+
+    public Figure(Color color) {
+        this.color = color;
+    }
+
+    public Color getColor() {
+        return color;
+    }
+
+    public void setColor(Color color) {
+        this.color = color;
+    }
+
+    @Override
+    public void draw() {
+        System.out.println("Figure: " + getShape().getDescription().toLowerCase()
+                + ", area:" + String.format("%.2f", getArea()) + " sq.units, "
+                + getSizeDescription() + ", string: " + color.name().toLowerCase());
+    }
+
+    public abstract double getArea();
+
+    public abstract Shape getShape();
+
+    protected abstract String getSizeDescription();
+}
--- a/src/main/java/core/basesyntax/Figure.java
+++ b/src/main/java/core/basesyntax/Figure.java
@@ -20,7 +20,7 @@ public abstract class Figure implements Drawable {
     public void draw() {
         System.out.println("Figure: " + getShape().getDescription().toLowerCase()
                 + ", area:" + String.format("%.2f", getArea()) + " sq.units, "
-                + getSizeDescription() + ", string: " + color.name().toLowerCase());
+                + getSizeDescription() + ", color: " + color.name().toLowerCase());
     }
 
     public abstract double getArea();
--- a/src/main/java/core/basesyntax/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/FigureSupplier.java
@@ -1,0 +1,50 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        Shape randomShape = Shape.values()[random.nextInt(Shape.values().length)];
+        // This requirement looks strange, but as you wish :-)
+        // ColorSupplier with public String getRandomColor() method - for generating random string,
+        Color randomColor = Color.valueOf(colorSupplier.getRandomColor());
+
+        switch (randomShape) {
+            case CIRCLE:
+                return new Circle(randomColor, random.nextInt(Figure.MAX_SIZE));
+            case ISOSCELES_TRAPEZOID:
+                return new IsoscelesTrapezoid(randomColor,
+                        random.nextInt(Figure.MAX_SIZE),
+                        random.nextInt(Figure.MAX_SIZE),
+                        random.nextInt(Figure.MAX_SIZE));
+            case RECTANGLE:
+                return new Rectangle(randomColor,
+                        random.nextInt(Figure.MAX_SIZE),
+                        random.nextInt(Figure.MAX_SIZE));
+            case RIGHT_TRIANGLE:
+                return new RightTriangle(randomColor,
+                        random.nextInt(Figure.MAX_SIZE),
+                        random.nextInt(Figure.MAX_SIZE));
+            case SQUARE:
+                return new Square(randomColor,
+                        random.nextInt(Figure.MAX_SIZE));
+            default:
+                System.out.println("Something went wrong!");
+                break;
+        }
+        return null;
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(Color.WHITE, 10);
+    }
+
+    private class ColorSupplier {
+        public String getRandomColor() {
+            return Color.values()[random.nextInt(Color.values().length)].name();
+        }
+    }
+}
--- a/src/main/java/core/basesyntax/HelloWorld.java
+++ b/src/main/java/core/basesyntax/HelloWorld.java
@@ -4,5 +4,22 @@ package core.basesyntax;
  * Feel free to remove this class and create your own.
  */
 public class HelloWorld {
+    private static final int ARRAY_SIZE = 6;
 
+    public static void main(String[] args) {
+        Figure[] figures = new Figure[ARRAY_SIZE];
+        FigureSupplier figureSupplier = new FigureSupplier();
+
+        for (int i = 0; i < ARRAY_SIZE; i++) {
+            if (i < 3) {
+                figures[i] = figureSupplier.getRandomFigure();
+            } else {
+                figures[i] = figureSupplier.getDefaultFigure();
+            }
+        }
+
+        for (Figure figure: figures) {
+            figure.draw();
+        }
+    }
 }
--- a/src/main/java/core/basesyntax/RightTriangle.java
+++ b/src/main/java/core/basesyntax/RightTriangle.java
@@ -1,0 +1,44 @@
+package core.basesyntax;
+
+public class RightTriangle extends Figure {
+
+    private int firstLeg;
+    private int secondLeg;
+
+    public RightTriangle(Color color, int firstLeg, int secondLeg) {
+        super(color);
+        this.firstLeg = firstLeg;
+        this.secondLeg = secondLeg;
+    }
+
+    public int getFirstLeg() {
+        return firstLeg;
+    }
+
+    public void setFirstLeg(int firstLeg) {
+        this.firstLeg = firstLeg;
+    }
+
+    public int getSecondLeg() {
+        return secondLeg;
+    }
+
+    public void setSecondLeg(int secondLeg) {
+        this.secondLeg = secondLeg;
+    }
+
+    @Override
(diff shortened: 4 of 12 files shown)


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,50 @@
...
+                        random.nextInt(Figure.MAX_SIZE));
+            default:
+                System.out.println("Something went wrong!");
+                break;
+        }
+        return null;
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(Color.WHITE, 10);
```
Ivan95kos, line 42: All magic numbers, strings, symbols in your code should be constants.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    private class ColorSupplier {
```
Ivan95kos, line 45: ColorSupplier should not be inner class

src/main/java/core/basesyntax/HelloWorld.java
```diff
@@ -4,5 +4,22 @@
  * Feel free to remove this class and create your own.
  */
 public class HelloWorld {
+    private static final int ARRAY_SIZE = 6;
 
+    public static void main(String[] args) {
+        Figure[] figures = new Figure[ARRAY_SIZE];
+        FigureSupplier figureSupplier = new FigureSupplier();
+
+        for (int i = 0; i < ARRAY_SIZE; i++) {
```
Ivan95kos, line 13: Use one loop `for`

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.Random;
+
+public class FigureSupplier {
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        Shape randomShape = Shape.values()[random.nextInt(Shape.values().length)];
+        // This requirement looks strange, but as you wish :-)
+        // ColorSupplier with public String getRandomColor() method - for generating random string,
```
Ivan95kos, line 12: Thanks for the comments,
but it is better to remove them

Ivan95kos (changes requested):
Good job overall!

src/main/java/core/basesyntax/HelloWorld.java
```diff
@@ -4,5 +4,19 @@
...
+            if (i < 3) {
```
okuzan, line 14: rely on your constant. If you decide to change it, you may forget to change it here

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,31 @@
+package core.basesyntax;
+
+public abstract class Figure implements Drawable {
+    public static final int MAX_SIZE = 100;
```
okuzan, line 4: why is this constant here

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,44 @@
...
+                        random.nextInt(Figure.MAX_SIZE),
+                        random.nextInt(Figure.MAX_SIZE));
+            case SQUARE:
+                return new Square(randomColor,
+                        random.nextInt(Figure.MAX_SIZE));
+            default:
+                System.out.println("Something went wrong!");
+                break;
+        }
+        return null;
```
okuzan, line 38:
```suggestion
         throw new RuntimeException("Couldn't generate random figure");
```
better throw an exception if the state is unexpected then  return  null silently

src/main/java/core/basesyntax/RightTriangle.java
```diff
@@ -0,0 +1,44 @@
+package core.basesyntax;
+
+public class RightTriangle extends Figure {
+
```
okuzan, line 4: remove the line

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,31 @@
...
+        System.out.println("Figure: " + getShape().getDescription().toLowerCase()
+                + ", area: " + String.format("%.2f", getArea()) + " sq.units, "
+                + getSizeDescription() + ", color: " + color.name().toLowerCase());
+    }
+
+    public abstract double getArea();
+
+    public abstract Shape getShape();
+
+    protected abstract String getSizeDescription();
```
okuzan, line 30: why access modifer different here?

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,31 @@
...
+    public void setColor(Color color) {
+        this.color = color;
+    }
+
+    @Override
+    public void draw() {
+        System.out.println("Figure: " + getShape().getDescription().toLowerCase()
+                + ", area: " + String.format("%.2f", getArea()) + " sq.units, "
+                + getSizeDescription() + ", color: " + color.name().toLowerCase());
+    }
```
okuzan, line 24: unused

okuzan (changes requested):

src/main/java/core/basesyntax/Figure.java
mate-academy-student-2023: getShape() represents a type of figure and could be considered the Figure class property.
getSizeDescription() was added as an utility method for draw() implementation. It wasn't supposed to be used independently. It might be called like describeSizes() to prevent confusing it with a getter.
Well, it could be both public, if we treat getSizeDescription() as a class property following getter's convention.
Let's go with the public. 
Thanks!

src/main/java/core/basesyntax/RightTriangle.java
mate-academy-student-2023: +

src/main/java/core/basesyntax/FigureSupplier.java
mate-academy-student-2023: Added. Thanks!

src/main/java/core/basesyntax/FigureSupplier.java
mate-academy-student-2023: Ok, Moved. Thanks!

src/main/java/core/basesyntax/HelloWorld.java
mate-academy-student-2023: Changed. Thanks!

src/main/java/core/basesyntax/FigureSupplier.java
mate-academy-student-2023: Remove. Thanks!

src/main/java/core/basesyntax/Figure.java
mate-academy-student-2023: I suppose you mean getColor().
Ok. added. Thank you!

mate-academy-student-2023: setColor() removed as well. Thanks!

src/main/java/core/basesyntax/Figure.java
mate-academy-student-2023: I thought about it as a maximum dimension for a figure and so to be something related to Figure class.
Ok, moved to the supplier. Thank you!

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,26 @@
...
+
+    @Override
+    public void draw() {
+        System.out.println("Figure: " + getShape().getDescription().toLowerCase()
+                + ", area: " + String.format("%.2f", getArea()) + " sq.units, "
+                + getSizeDescription() + ", color: " + getColor().name().toLowerCase());
+    }
+
+    public abstract double getArea();
```
olekskov, line 21: Please check common mistake regarding behaviour and introduce interface for calculating are method
https://mate-academy.github.io/jv-program-common-mistakes/java-core/abstract-class-interface/oop-advanced#dont-use-abstract-classes-to-set-behaviour-for-classes

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,26 @@
...
+
+    public abstract Shape getShape();
```
olekskov, line 23: I can't understand why name of the figure should be done via separate method.
Make it class property if you feel it necessary.
But for purposes it used here you may just call getClass().getSimpleName() method and you will get rather fine name of figure in draw() method without additional enum

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,26 @@
...
+
+    public abstract String getSizeDescription();
```
olekskov, line 25: This method is just used to extract part of text from draw() method.
This method doesn't provide any functionality that is aimed to be used with public access.
Also it's not good that we're forcing to follow our code-style pattern in such way. Other programmers may want to use their own code-style, probably without extracting text from draw() method, when creating new child class for our Figure. 
Please just use private methods in child classes if you really want to extract something from public method

src/main/java/core/basesyntax/HelloWorld.java
```diff
@@ -4,5 +4,19 @@
  * Feel free to remove this class and create your own.
  */
 public class HelloWorld {
```
olekskov, line 6: Please use more appropriate class name

olekskov (changes requested):

src/main/java/core/basesyntax/Figure.java
mate-academy-student-2023: Agree. 
AreaCalculator with calculateArea() method added.
Thank you!

src/main/java/core/basesyntax/Figure.java
mate-academy-student-2023: Agree, these methods may not be public if we consider their usage.
Protected may look better.

getShape() and getSizeDescription() are aimed to prevent duplicating code of draw() in every Figure child class.

If we follow the idea of not forcing other programmers to write these methods and ask them to write their own version of the draw() method, we could place these methods into a dedicated interface. 
Please take a look at fielddescriptor branch. 
Do you think that version is better? 
In my opinion, it looks awkward and overdesigned for this task purpose.

src/main/java/core/basesyntax/HelloWorld.java
mate-academy-student-2023: Changed. Thanks!

src/main/java/core/basesyntax/RightTriangle.java
```diff
@@ -0,0 +1,44 @@
+package core.basesyntax;
+
+public class RightTriangle extends Figure {
+    public static final String SHAPE_NAME = "Right Triangle";
```
overpathz, line 4: Why this constant are public if it's usage limited by scope of this class? (other triable as well)

mate-academy-student-2023: Agree. Thanks to IDEA :-)
Fixed.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,44 @@
...
+        Shape randomShape = Shape.values()[random.nextInt(Shape.values().length)];
+        Color randomColor = Color.valueOf(colorSupplier.getRandomColor());
+
+        switch (randomShape) {
+            case CIRCLE:
+                return new Circle(randomColor, random.nextInt(MAX_SIZE));
+            case ISOSCELES_TRAPEZOID:
+                return new IsoscelesTrapezoid(randomColor,
+                        random.nextInt(MAX_SIZE),
+                        random.nextInt(MAX_SIZE),
```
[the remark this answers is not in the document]
mate-academy-student-2023: Agree. Moved.

PR #1117, opened 2023-06-19
created all

LiudmylaPechevysta:


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,40 @@
...
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        int index = random.nextInt(FIGURE_COUNT);
+        switch (index) {
+            case 0:
+                return new Circle(colorSupplier.getRandomColor(), random.nextInt(MAX_VALUE));
+            case 1:
+                return new Rectangle(colorSupplier.getRandomColor(), random.nextInt(MAX_VALUE),
+                        random.nextInt(MAX_VALUE));
```
overpathz, line 21: Consider creating utility method like "getRandomNumber()", so as not to duplicate this "random.nextInt(..)"

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,40 @@
...
+            case 2:
+                return new RightTriangle(colorSupplier.getRandomColor(), random.nextInt(MAX_VALUE),
+                        random.nextInt(MAX_VALUE));
+            case 3:
+                return new Square(colorSupplier.getRandomColor(), random.nextInt(MAX_VALUE));
```
overpathz, line 26: I would move this "new <SomeFigure>(..)" to separate methods (and you will be have smth like: getCircle(), getTriangle(), etc).

Also, you have default option in switch-case, you can use your getDefaultFigure() method instead.

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,18 @@
+package core.basesyntax;
+
+public class Main {
+    public static void main(String[] args) {
+        Figure[] figures = new Figure[6];
```
overpathz, line 5: This, you are not using your figures in this array
(you need to initialize(fill) it)

overpathz (changes requested):
Other places are ok.
Update PR with specified fixes and it'll OK to approve. Thanks

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,18 @@
...
+        FigureSupplier figureSupplier = new FigureSupplier();
+
+        for (int i = 0; i < figures.length; i++) {
+            if (i < figures.length / 2) {
+                figures[i] = figureSupplier.getRandomFigure();
+                figures[i].draw();
+            } else {
+                figures[i] = figureSupplier.getDefaultFigure();
+                figures[i].draw();
+            }
```
olekskov, line 15: It's better to extract common logic:
```suggestion
        for (int i = 0; i < figures.length; i++) {
            if (i < figures.length / 2) {
                figures[i] = figureSupplier.getRandomFigure();
            } else {
                figures[i] = figureSupplier.getDefaultFigure();
            }
            figures[i].draw();
```

olekskov (approved):
Good job

PR #1134, opened 2023-06-30
added new classes and interfaces

OlegTuryk:


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,42 @@
...
+            case 4:
+                return new Circle(colorSupplier.getRandomColor(),
+                        random.nextInt(figureSideBounds) + 1);
+            case 5:
+                return new IsoscelesTrapezoid(colorSupplier.getRandomColor(),
+                        random.nextInt(figureSideBounds) + 1,
+                        random.nextInt(figureSideBounds) + 1,
+                        random.nextInt(figureSideBounds) + 1);
+            default:
+                throw new IllegalArgumentException("Illegal figure");
```
nacenik, line 35: Bad exception message

nacenik (changes requested):
Good solution! Let’s improve it 😉

src/main/java/core/basesyntax/RightTriangle.java
```diff
@@ -4,21 +4,22 @@ public class RightTriangle extends Figure {
...
+        System.out.println("Figure: triangle, area: " + calculateArea()
                 + " sq.units, side: " + side + " units, height: "
                 + height + " units, color: " + getColor());
     }
 
     @Override
-    public double getArea() {
-        return 0.5 * side * height;
+    public double calculateArea() {
+        final double coefficient = 0.5;
```
Vasul-Pankiv, line 22: This is not a constant
Constant should be class level, static and final. Constant name should be in uppercase
For example private static final double FORMULA_COEFFICIENT = 0.5;

Vasul-Pankiv (approved):

PR #1139, opened 2023-07-01
implemented figure classes

DmytroV95:


src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public abstract class Figure implements AreaCalculator, Drawable {
+    private final String color;
```
nacenik, line 4: This variable should be Color type, not String

src/main/java/core/basesyntax/FigureEnum.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum FigureEnum {
```
nacenik: Bad name. 
```suggestion
public enum FigureType {
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,48 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    public static final FigureEnum DEFAULT_FIGURE_NAME = FigureEnum.CIRCLE;
+    public static final double DEFAULT_CIRCLE_RADIUS = 10;
+    public static final String DEFAULT_CIRCLE_COLOR = Color.WHITE.name();
+    public static final int MAX_FIGURES = 6;
```
nacenik, line 9: Bad name. `MAX_FIGURES` means the maximum number of figures, not the maximum size or sth like this

src/main/java/core/basesyntax/IsoscelesTrapezoid.java
```diff
@@ -0,0 +1,33 @@
...
+        super(color);
+        this.upperBase = upperBase;
+        this.bottomBase = bottomBase;
+        this.height = height;
+        this.figureName = FigureEnum.ISOSCELES_TRAPEZOID;
+    }
+
+    @Override
+    public double getFigureArea() {
+        return 2 * height * (upperBase + bottomBase);
```
nacenik, line 21: All magic numbers should be constant 
fix it for all figures

src/main/java/core/basesyntax/AreaCalculator.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface AreaCalculator {
+    double getFigureArea();
```
nacenik, line 4: Better to rename this method on `calculateFigureArea` to avoid misunderstanding that it is a getter for field

nacenik (changes requested):
Good solution! Let’s improve it 😉

Vasul-Pankiv (approved):

PR #1145, opened 2023-07-02
Add new classes and their implementations. Also add new interfaces.

OlegStashkiv:


src/main/java/core/basesyntax/enums/Color.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax.enums;
+
+public enum Color {
+    BLUE,
+    RED,
+    BlACK,
```
nacenik, line 6: Typo mistake 
```suggestion
    BLACK,
```

src/main/java/core/basesyntax/models/RightTriangle.java
```diff
@@ -0,0 +1,26 @@
...
+
+    public RightTriangle(Color color, double firstLeg, double secondLeg) {
+        super(color);
+        this.firstLeg = firstLeg;
+        this.secondLeg = secondLeg;
+    }
+
+    @Override
+    public double getArea() {
+        return firstLeg * secondLeg / 2;
```
nacenik, line 17: All magic numbers should be a constant
fix it for all figures

nacenik (approved):
Great job! You can improve your solution if you want 😉

PR #1152, opened 2023-07-03
added new classes, interfaces and implemented main application logic

ArtemReizer:


src/main/java/core/basesyntax/FigureDrawing.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface FigureDrawing {
```
KseniiaMakarova, line 3:
```suggestion
public interface Drawer {
```

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #1169, opened 2023-07-10
jv-oop-advanced-solution

Valdemar3X:


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,62 @@
...
+                break;
+
+            }
+            default: {
+                System.out.println("Switch default");
+                break;
+            }
+
+        }
+        return figure;
```
olekskov, line 56: 1. unnecessary empty lines
2. unnecessary breaks
3. returning null is a very bad practice, it's better to remove case 4 and make last figure to be returned in default case
```suggestion
            case 0: {
                return new Square(randomPart1, randomColor);
            }
            case 1: {
                return new Rectangle(randomPart1, randomPart2, randomColor);
            }
            case 2: {
                return new Circle(randomPart1, randomColor);
            }
            case 3: {
                return new RightTriangle(randomPart1, randomPart2, randomColor);
            }
            default: {
                return new IsoscelesTrapezoid(randomPart1, randomPart2, randomPart3, randomColor);
            }
        }
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Feel free to remove this class and create your own.
+ */
+public class Main {
+    public static void main(String[] args) {
+
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] figures = new Figure[6];
+        for (int i = 0; i < figures.length; i++) {
+            if (i < 3) {
```
olekskov, line 14: it's better to use expression instead of hardcoded value where possible to be able to make change in place in case number of figures is changed
```suggestion
            if (i < FIGURES_AMOUNT / 2) {
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,23 @@
...
+import core.basesyntax.model.Figure;
+
+/**
+ * Feel free to remove this class and create your own.
+ */
+public class Main {
+    public static void main(String[] args) {
+
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] figures = new Figure[6];
```
olekskov, line 12: Please use constant for magic value
```suggestion
        Figure[] figures = new Figure[FIGURES_AMOUNT];
```

olekskov (changes requested):
Good job regarding structure and logic.
Need to fix some code-style issues

olekskov (approved):
Good job

PR #1171, opened 2023-07-11
done

bas0101:


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,37 @@
...
+    private final Random random = new Random();
+
+    public Figure getDefaultFigure() {
+        return new Circle(DEFAULT_COLOR.name(), DEFAULT_RADIUS);
+    }
+
+    public Figure getRandomFigure() {
+        String color = colorSupplier.getRandomColor();
+
+        Figure[] figures = new Figure[]{
```
Ivan95kos, line 21: You always generate all Figures, but you only need one
try to use the next example:
```
        int index = random.nextInt(NUMBER_OF_FIGURES);
        switch (index) {
            case 1:
                return prepareCircle();
            case 2:
                return prepareIsoscelesTrapezoid();
            case 3:
                return prepareRightTriangle();
            case 4:
                return prepareRectangle();
            case 5:
                return prepareSquare();
            default:
                return getDefaultFigure();
        }
    }
```

Ivan95kos (changes requested):
Good job overall!

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,22 @@
...
+            if (i < FIGURES_QUANTITY / 2) {
+                figures[i] = figureSupplier.getRandomFigure();
+            } else {
+                figures[i] = figureSupplier.getDefaultFigure();
+            }
+        }
+
+        for (Figure figure : figures) {
+            figure.draw();
+        }
```
okuzan, line 20:
```suggestion
            }
            figures[i].draw();
        }
```

okuzan (approved):
Good job!

PR #1177, opened 2023-07-14
Solution

breakingtumble:
I don't understand only one thing, why we need to create some interfaces to calculate area or draw a figure, instead we can create these abstract methods in Figure class, nothing will conceptually change?

### DIFF
--- a/src/main/java/core/basesyntax/AreaCalculable.java
+++ b/src/main/java/core/basesyntax/AreaCalculable.java
@@ -1,0 +1,5 @@
+package core.basesyntax;
+
+public interface AreaCalculable {
+    double calculateArea();
+}
--- a/src/main/java/core/basesyntax/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/FigureSupplier.java
@@ -1,0 +1,35 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private final static int FIGURE_COUNT = 5;
+    private final static int UPPER_INTEGER_BOUND = 10;
+    private final Random random = new Random();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        switch (random.nextInt(FIGURE_COUNT)) {
+            case 0:
+                return new Circle(colorSupplier.getRandomColor(), getRandomIntValue());
+            case 1:
+                return new IsoscelesTrapezoid(colorSupplier.getRandomColor(), getRandomIntValue(), getRandomIntValue(), getRandomIntValue());
+            case 2:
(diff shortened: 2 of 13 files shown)


src/main/java/core/basesyntax/AreaCalculable.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface AreaCalculable {
```
aaananas, line 3:
```suggestion
public interface AreaCalculator {
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,38 @@
...
+                        getRandomIntValue());
+            case 4:
+                return new Square(colorSupplier.getRandomColor(), getRandomIntValue());
+            default:
+                return new Circle(Color.WHITE.name(), getRandomIntValue());
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(Color.WHITE.name().toLowerCase(), 10);
```
aaananas, line 32: lets add constants for default values

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,38 @@
...
+public class FigureSupplier {
+    private static final int FIGURE_COUNT = 5;
+    private static final int UPPER_INTEGER_BOUND = 10;
+    private final Random random = new Random();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        switch (random.nextInt(FIGURE_COUNT)) {
+            case 0:
+                return new Circle(colorSupplier.getRandomColor(), getRandomIntValue());
```
aaananas, line 14: lets add private methods to get each random figure - this would make your code easier to read

```suggestion
            case 0:
                return getRandomCircle();
```

aaananas (changes requested):

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax;
+
+public class MainApp {
+    private static final int SIZE_FIGURE_ARRAY = 6;
+    private static final FigureSupplier figureSupplier = new FigureSupplier();
+
+    public static void main(String[] args) {
+        Figure[] figures = new Figure[SIZE_FIGURE_ARRAY];
+        for (int i = 0; i < figures.length; i++) {
+            if (i < 3) {
```
olekskov, line 10: It's better to reuse constants in expressions in order to make only one change if you require another number of figures:
```suggestion
            if (i < SIZE_FIGURE_ARRAY / 2) {
```

olekskov (approved):
Good job, approved

PR #1178, opened 2023-07-15
oop-advanced homework is done

GiantBear321:
I decided to make array in method, but it`s also working as a class constant. Could i make array in class? Is it bad manners or its ok when it`s ok for a goal?

### DIFF
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,0 +1,21 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+
+    private final Random random = new Random();
+
+    public enum Color {
+        RED,
+        WHITE,
+        BLACK,
+        BLUE,
+        YELLOW,
+        GREEN
+    }
+
+    public String getRandomColor() {
+        return Color.values()[random.nextInt(Color.values().length)].name();
+    }
+}
--- a/src/main/java/core/basesyntax/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/FigureSupplier.java
@@ -1,0 +1,71 @@
+package core.basesyntax;
+
+import core.basesyntax.figures.Circle;
+import core.basesyntax.figures.IsoscelesTrapezoid;
+import core.basesyntax.figures.Rectangle;
+import core.basesyntax.figures.RightTriangle;
+import core.basesyntax.figures.Square;
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final String DEFAULT_COLOR = "White";
+    private static final int DEFAULT_BOND = 100;
+    private static final int DEFAULT_RADIUS = 10;
+    private final Random random = new Random();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public enum FigureList {
+        CIRCLE,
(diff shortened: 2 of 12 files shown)


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,71 @@
...
+    public enum FigureList {
+        CIRCLE,
+        ISOSCELES_TRAPEZOID,
+        RECTANGLE,
+        RIGHT_TRIANGLE,
+        SQUARE
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(FigureSupplier.DEFAULT_COLOR, FigureSupplier.DEFAULT_RADIUS);
```
olekskov, line 26: There is no need to indicate own class
```suggestion
        return new Circle(DEFAULT_COLOR, DEFAULT_RADIUS);

```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    private String getFigureType() {
+        return FigureList.values()[random.nextInt(FigureList.values().length)].name();
+    }
```
olekskov, line 31: All private methods should be placed after public ones in order to see main purpose of class before details of implementation

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,21 @@
...
+    private final Random random = new Random();
+
+    public enum Color {
+        RED,
+        WHITE,
+        BLACK,
+        BLUE,
+        YELLOW,
+        GREEN
+    }
```
olekskov, line 16: It's NOT a mistake but I should aware you that using enums as inner classes may lead to additional code rework in case parent class is no longer required and should be deleted but enum is still used somewhere else in the code

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,23 @@
...
+            figures[i] = figureSupplier.getDefaultFigure();
+            figures[i].drawInfo();
+        }
+
+        System.out.println();
+
+        for (int i = figures.length / 2; i < figures.length; i++) {
+            figures[i] = figureSupplier.getRandomFigure();
+            figures[i].drawInfo();
+        }
```
olekskov, line 21:
```suggestion
        for (int i = 0; i < figures.length; i++) {
            if (i < figures.length / 2) {
                figures[i] = figureSupplier.getDefaultFigure();
            } else {
                figures[i] = figureSupplier.getRandomFigure();
            }
            figures[i].drawInfo();
        }
```
or
```suggestion
        for (int i = 0; i < figures.length; i++) {
            figures[i] = i < figures.length / 2 
                    ? figureSupplier.getDefaultFigure();
                    : figureSupplier.getRandomFigure();
            figures[i].drawInfo();
        }
```

olekskov (changes requested):
Good job overall, only minor code-style comments

olekskov:
> I decided to make array in method, but its also working as a class constant. Could i make array in class? Is it bad manners or its ok when its ok for a goal?

It was totally fine

GiantBear321:
> Good job overall, only minor code-style comments

Ty!

aaananas (approved):

PR #1192, opened 2023-07-28
Implemented Figure classes, ColorSupplier, FigureSupplier classes, Dr…

ystankevych:
…awable interface, psvm in Main class

### DIFF
--- a/src/main/java/core/basesyntax/Figure.java
+++ b/src/main/java/core/basesyntax/Figure.java
@@ -1,0 +1,25 @@
+package core.basesyntax;
+
+public abstract class Figure implements Drawable {
+    private String color;
+
+    public Figure(String color) {
+        this.color = color;
+    }
+
+    public String getColor() {
+        return color;
+    }
+
+    public abstract double getArea();
+
+    public void draw() {
+        System.out.printf("Figure: %s, area: %.2f sq.units",
+                getFigureName(), getArea());
+    }
+
+    private String getFigureName() {
+        return String.join(" ",
+                getClass().getSimpleName().split("(?=\\p{Upper})")).toLowerCase();
+    }
+}
--- a/src/main/java/core/basesyntax/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/FigureSupplier.java
@@ -1,0 +1,90 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    public static final int FIGURE_LENGTH_RANGE;
+    public static final int FIGURE_COUNT;
+    private final ColorSupplier colorSupplier;
+    private final Random random;
+
+    static {
+        FIGURE_LENGTH_RANGE = 10;
+        FIGURE_COUNT = 5;
+    }
+
+    {
+        colorSupplier = new ColorSupplier();
+        random = new Random();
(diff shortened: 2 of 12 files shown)


liliia-ponomarenko (changes requested):
Good job! Let's improve your solution ;)

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,23 @@
...
+
+    public void draw() {
+        System.out.printf("Figure: %s, area: %.2f sq.units",
+                getFigureName(), getArea());
+    }
+
+    private String getFigureName() {
+        return String.join(" ",
+                getClass().getSimpleName().split("(?=\\p{Upper})")).toLowerCase();
+    }
```
Andriy-dz, line 22: Do you really need these methods? You can already do this in an overridden method in shapes

src/main/java/core/basesyntax/IsoscelesTrapezoid.java
```diff
@@ -0,0 +1,26 @@
+package core.basesyntax;
+
+public class IsoscelesTrapezoid extends Figure {
+    private double baseOne;
+    private double baseTwo;
```
Andriy-dz, line 5:
```suggestion
    private double bottomBase;
    private double topBase;
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,74 @@
...
+                return createRectangle();
+            case 3:
+                return createRightTriangle();
+            default:
+                return createIsoscelesTrapezoid();
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle("white", MAX_LENGTH_RANGE);
```
Andriy-dz, line 34:
```suggestion
        return new Circle(Color.WHITE, MAX_LENGTH_RANGE);
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,74 @@
...
+        return new RightTriangle(color, firstLeg, secondLeg);
+    }
+
+    private IsoscelesTrapezoid createIsoscelesTrapezoid() {
+        String color = colorSupplier.getRandomColor();
+        double baseOne = getRandomSideLength();
+        double baseTwo = getRandomSideLength();
+        double height = getRandomSideLength();
+        return new IsoscelesTrapezoid(color, baseOne, baseTwo, height);
+    }
```
Andriy-dz, line 69: when you create each shape, you call the same method - getRandomColor. Is it better to call this method in the getRandomFigure method and pass the result in the parameters of each method? So you will get rid of duplication of code

Andriy-dz (changes requested):

src/main/java/core/basesyntax/Figure.java
ystankevych: Please, look at this peace of code one more time. I 've created getFigureMethod in Abstract class and all other subclasses just extend this method and add only their new options. But if you think it would be better to write this peace of code in each code - I'll changed it.

Andriy-dz (approved):

PR #1197, opened 2023-07-31
hw jv-oop-advanced

TarasYurkovskyi:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    public Color getRandomColor() {
+        int index = new Random().nextInt(Color.values().length);
```
d1sam, line 7: Consider which variables should be local in the method and which should be class-level.

src/main/java/core/basesyntax/RightTriangle.java
```diff
@@ -0,0 +1,24 @@
+package core.basesyntax;
+
+public class RightTriangle extends Figure {
+    private double firstLeg;
+    private double secondLeg;
+
+    public RightTriangle(Color color,double firstLeg, double secondLeg) {
```
d1sam, line 7: After ',' we need a space.

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,23 @@
...
+
+    public static void main(String[] args) {
+        Figure[] figures = new Figure[6];
+        for (int i = 0; i < figures.length; i++) {
+            if (i < figures.length / 2) {
+
+                figures[i] = new FigureSupplier().getRandomFigure();
+
+            } else {
+                figures[i] = new FigureSupplier().getDefaultFigure();
```
d1sam, line 16: Please don’t create a new FigureSupplier() inside the loop. Please remove empty lines. Firstly you should initialize an array and in the next loop you should print it.

d1sam (changes requested):
Some changes should be done.

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,21 @@
...
+    public static void main(String[] args) {
+        Figure[] figures = new Figure[ARRAY_SIZE];
+        FigureSupplier figureSupplier = new FigureSupplier();
+        for (int i = 0; i < ARRAY_SIZE; i++) {
+            if (ARRAY_SIZE / 2 > i) {
+                figures[i] = figureSupplier.getRandomFigure();
+            } else {
+                figures[i] = figures[i - ARRAY_SIZE / 2];
+            }
+            figures[i].draw();
```
kshuryhin, line 18: Well, what need to be done, is that first half of the array should be randomly generated figures. **The second part should be same circles.** **Знімок екрана 2023-09-26 о 09 02 49**
Here is how it looks now.

kshuryhin (changes requested):
Nice try, but some changes should be done to meet the task goals

kshuryhin (approved):
Good job!

PR #1202, opened 2023-08-02
the code compiles but maven gives an error

NazarOliinyk:
help with the error when I run maven

### DIFF
--- a/src/main/java/core/basesyntax/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/FigureSupplier.java
@@ -1,0 +1,61 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int VALUE_LIMIT = 10;
+    private static final String DEFAULT_COLOR = "WHITE";
+    private static final int DEFAULT_RADIUS = 10;
+
+    private final Random randomizer = new Random();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        switch (randomizer.nextInt(Figures.values().length)) {
+            case 1:
+                return prepareCircle();
+            case 2:
+                return prepareIsoscelesTrapezoid();
+            case 3:
+                return prepareRightTriangle();
+            case 4:
+                return prepareRectangle();
+            default:
+                return prepareSquare();
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(DEFAULT_RADIUS, DEFAULT_COLOR);
+    }
+
+    private int randomInteger() {
+        return randomizer.nextInt(VALUE_LIMIT);
+    }
+
+    private Circle prepareCircle() {
+        return new Circle(randomInteger(), colorSupplier.getRandomColor());
+    }
+
+    private IsoscelesTrapezoid prepareIsoscelesTrapezoid() {
+        return new IsoscelesTrapezoid(
+                randomInteger(),
+                randomInteger(),
+                randomInteger(),
+                colorSupplier.getRandomColor()
+        );
+    }
+
+    private Square prepareSquare() {
+        return new Square(randomInteger(), colorSupplier.getRandomColor());
+    }
+
+    private RightTriangle prepareRightTriangle() {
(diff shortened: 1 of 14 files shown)


olekskov (changes requested):
Please fix tests/build before we may proceed with mentor's review.
Please run 'mvn clean package' command locally to check if build is fine

NazarOliinyk:
locally, I don't have a build, although the build is successful here in git

src/main/java/core/basesyntax/IsoscelesTrapezoid.java
```diff
@@ -0,0 +1,30 @@
+package core.basesyntax;
+
+public class IsoscelesTrapezoid extends Figure {
+    private int base1;
```
okuzan, line 4: Naming with numbers is a bad practice

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,61 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int VALUE_LIMIT = 10;
+    private static final String DEFAULT_COLOR = "WHITE";
```
okuzan, line 7: you have enum for white

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,61 @@
...
+    private static final int DEFAULT_RADIUS = 10;
+
+    private final Random randomizer = new Random();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        switch (randomizer.nextInt(Figures.values().length)) {
+            case 1:
```
okuzan, line 15: perfect opportunity to use your enum for figures

src/main/java/core/basesyntax/Figures.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum Figures {
```
okuzan, line 3:
```suggestion
public enum FigureType {
```

okuzan (changes requested):
Good job overall!
Try to get rid of the .DS_store files, that are created automatically, they shouldn't be committed. Some of your files have a trailing new line, that should be deleted. Better use one System.out.println with a concatenated string instead of several calls as you have now.

NazarOliinyk:
echo .DS_Store >> ~/.gitignore_global
 ✗ git config --global core.excludesfile ~/.gitignore_global - не помогло 

 echo .DS_Store >> .gitignore
 git add .gitignore
 git commit -m '.DS_Store banished!' - не помогло 

nano .gitignore
Додав рядок .DS_Store до файлу .gitignore - не помогло

Знімок екрана 2023-08-04 о 13 01 37

NazarOliinyk:
I still have the error
Знімок екрана 2023-08-04 о 13 02 29

src/test/java/core/test/StructureTest.java
```diff
@@ -106,7 +106,7 @@ private void checkSuperclass(List<Class> figureClasses) {
   private void checkSuperClassInterfaces(Class clazz) {
     Class[] interfaces = clazz.getInterfaces();
     if (interfaces.length == 0) {
-      Assert.fail("Figures parent class " + clazz.getSuperclass().getSimpleName()
+      Assert.fail("FigureType parent class " + clazz.getSuperclass().getSimpleName()
```
okuzan, line 109: Be careful when renaming something project-wide and review what you are committing, you shouldn't make changes here

okuzan (changes requested):

src/test/java/core/test/StructureTest.java
NazarOliinyk: You said to name Color - Colortype; I thought that the figures can also be changed because it is also an enum. I fixed it and I want to finish this task and move to the others. My solution to the problem with the name Figures does not work.

NazarOliinyk: I tried with different names and there is an error

Ivan95kos (approved):
Great job

PR #1208, opened 2023-08-06
jv-oop-advasnced-solution

PeresadaKostya:


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,33 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final String COLOR = Color.WHITE.toString();
```
```suggestion
    private static final String COLOR = Color.WHITE.name();
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,21 @@
...
+
+        Figure[] figures = new Figure[ARRAY_LENGTH];
+        for (int i = 0; i < figures.length; i++) {
+            if (i < 3) {
+                figures[i] = figureSupplier.getRandomFigure();
+            } else {
+                figures[i] = figureSupplier.getDefaultFigure();
+            }
+        }
+        for (Figure figure : figures) {
```
okuzan, line 17: include this in your first loop, no need for two loops

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,21 @@
...
+
+public class Main {
+    private static final int ARRAY_LENGTH = 6;
+
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+
+        Figure[] figures = new Figure[ARRAY_LENGTH];
+        for (int i = 0; i < figures.length; i++) {
+            if (i < 3) {
```
okuzan, line 11: instead of 3 rely on array's length, this makes it easier to change number of figures (in one place instead of two)

src/main/java/core/basesyntax/PrintInformation.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface PrintInformation {
+    void print();
+}
```
okuzan, line 5: ununsed interface

okuzan (changes requested):

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,33 @@
...
+                new Square(colorSupplier.getRandomColor(), getRandomValue()),
+                new Rectangle(colorSupplier.getRandomColor(), getRandomValue(),
+                        getRandomValue()),
+                new RightTriangle(colorSupplier.getRandomColor(), getRandomValue(),
+                        getRandomValue()),
+                new IsoscelesTrapezoid(colorSupplier.getRandomColor(), getRandomValue(),
+                        getRandomValue(), getRandomValue())
+        };
+        return figures[random.nextInt(figures.length)];
+    }
```
olekskov, line 24: I don't like solution where we create 5 different objects each time method is called when we require just 1 of them.
Please redesign method to create only required type of figure

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    public String getRandomColor() {
+        int index = new Random().nextInt(Color.values().length);
```
olekskov, line 7: Please extract Random to class variable as done in Lottery class

olekskov (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
okuzan: not fixed

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,60 @@
...
+
+    private IsoscelesTrapezoid getIsoscelesTrapezoid() {
+        return new IsoscelesTrapezoid(colorSupplier.getRandomColor(), getRandomValue(),
+                getRandomValue(), getRandomValue());
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(COLOR, CIRCLE_DEFAULT_RADIUS);
+    }
```
okuzan, line 55: public should go before private methods

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,20 @@
...
+        FigureSupplier figureSupplier = new FigureSupplier();
+
+        Figure[] figures = new Figure[ARRAY_LENGTH];
+        for (int i = 0; i < figures.length; i++) {
+            if (i < figures.length / 2) {
+                figures[i] = figureSupplier.getRandomFigure();
+                figures[i].draw();
+            } else {
+                figures[i] = figureSupplier.getDefaultFigure();
+                figures[i].draw();
```
okuzan, line 16: extract common part from if

okuzan (changes requested):

src/main/java/core/basesyntax/Main.java
PeresadaKostya: I do not understand what i must do with it.

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,20 @@
...
+            }
```
okuzan, line 17:
```suggestion
        for (int i = 0; i < figures.length; i++) {
            if (i < figures.length / 2) {
                figures[i] = figureSupplier.getRandomFigure();
            } else {
                figures[i] = figureSupplier.getDefaultFigure();
            }
            figures[i].draw();
```

okuzan (changes requested):

okuzan (changes requested):
__to unblock platform__

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,60 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final String COLOR = Color.WHITE.name();
```
overpathz, line 6: Naming should be "DEFAULT_COLOR"

overpathz (approved):
Approved.

But take into account comments with minor things so as not to do them in the future.

PR #1211, opened 2023-08-09
do exercise

silchenko-arsen:


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,36 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final String WHITE = Color.WHITE.getColor();
```
```suggestion
    private static final String DEFAULT_COLOR = Color.WHITE.getColor();
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,15 @@
...
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] figures = new Figure[]{figureSupplier.getRandomFigure(),
+                figureSupplier.getRandomFigure(),
+                figureSupplier.getDefaultFigure(),
+                figureSupplier.getDefaultFigure()};
+        for (Figure figure : figures) {
+            figure.draw();
+            figure.getArea();
+        }
```
okuzan, line 13: better create an array of fixed length (constant), then fill it in the for loop (if the current index is less the half of length- random, else - default). Then call draw(). Why to call getArea() at all?

okuzan (changes requested):

okuzan (approved):
Well done!

PR #1213, opened 2023-08-09
implementation of figures

sonyasapsan:


src/main/java/core/basesyntax/figures/Drawable.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax.figures;
+
+public interface Drawable {
+    String draw();
```
NazarSmith228, line 4: seems you've missed the task requirement a bit:
`To 'draw' means to print out all information about a figure using System.out.println() (you shouldn't override toString() method for this)`

`draw()` method should not return anything - it should simply print the information about a figure
```suggestion
    void draw();

```

src/main/java/core/basesyntax/figures/Figure.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax.figures;
+
+public abstract class Figure implements Drawable, AreaCalculator {
+    protected String color;
+
+    Figure(String color) {
```
NazarSmith228, line 6: you might make this constructor `protected` as well, so that subclasses from other packages will also have access to it

no modifier it the same as `package-private` modifier, which in turn means that this constructor is accessible for all the classes within this package, however if you create a subclass of `Figure` in other package - it won't have access to that constructor

```suggestion
    protected Figure(String color) {

```

src/main/java/core/basesyntax/suppliers/Color.java
```diff
@@ -0,0 +1,20 @@
...
+    BLUE("blue"),
+    YELLOW("yellow"),
+    GREEN("green"),
+    WHITE("white"),
+    BLACK("black"),
+    RED("red");
+
+    private String color;
+
+    Color(String color) {
```
NazarSmith228, line 13: you've missed `private` modifier
```suggestion
    private Color(String color) {

```

src/main/java/core/basesyntax/suppliers/FigureSupplier.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.Random;
+
+public class FigureSupplier {
+    static final double MAX_LENGTH = 25.0;
+    static final int AMOUNT_OF_FIGURES = 5;
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getDefaultFigure() {
+        return new Circle(Color.WHITE.getColor(), 10);
```
NazarSmith228, line 18: what is `10`? a magic number? :)

src/main/java/core/basesyntax/suppliers/Color.java
sonyasapsan: Private Color()? Maybe Public Color()?

NazarSmith228: actually no, because enum is a constant, you don't create them via constructor:
```
Color color = Color.WHITE; // correct, here your private constructor will be called with color parameter equal to "white"
Color color = new Color("white") // incorrect, since enum is no more a constant as you can pass any value into the constructor
```

src/main/java/core/basesyntax/suppliers/ColorSupplier.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax.suppliers;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private Random random = new Random();
+
+    public String getRandomColor() {
+        return Color.values()[random.nextInt(Color.values().length - 1)].getColor();
```
NazarSmith228, line 9: you don't need to substact `1` from the array length, since parameter in `nextInt()` method is exclusive

take a look at `nextInt(int bound)` javadoc:
image

src/main/java/core/basesyntax/suppliers/FigureSupplier.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    public Figure getRandomFigure() {
+        int randomFigure = random.nextInt(AMOUNT_OF_FIGURES) + 1;
+        String color = colorSupplier.getRandomColor();
+        switch (randomFigure) {
+            case 1:
+                double radius = MAX_LENGTH * random.nextDouble();
+                return new Circle(color, radius);
```
NazarSmith228, line 27: you may create additional method called `getRandomUnits()` and reuse it in all places:
```
private double getRandomUnits() {
   return MAX_LENGTH * random.nextDouble();
}
```
```suggestion
                return new Circle(color, getRandomUnits());

```

src/main/java/core/basesyntax/suppliers/FigureSupplier.java
```diff
@@ -0,0 +1,47 @@
...
+import core.basesyntax.figures.Figure;
+import core.basesyntax.figures.IsoscelesTrapezoid;
+import core.basesyntax.figures.Rectangle;
+import core.basesyntax.figures.RightTriangle;
+import core.basesyntax.figures.Square;
+import java.util.Random;
+
+public class FigureSupplier {
+    static final double MAX_LENGTH = 25.0;
+    static final int AMOUNT_OF_FIGURES = 5;
```
NazarSmith228, line 13: seems you've missed `private` modifiers :)

NazarSmith228 (changes requested):
Great job, now it's time to work on some remarks :)

src/main/java/core/basesyntax/suppliers/Color.java
sonyasapsan: > actually no, because enum is a constant, you don't create them via constructor:
> 
> ```
> Color color = Color.WHITE; // correct, here your private constructor will be called with color parameter equal to "white"
> Color color = new Color("white") // incorrect, since enum is no more a constant as you can pass any value into the constructor
> ```

IDEA doesn't like that idea, it highlights key private in gray.

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,18 @@
...
+
+public class Main {
+    private static final int NUMBER_OF_FIGURES = 6;
+    private static FigureSupplier figureSupplier = new FigureSupplier();
+
+    public static void main(String[] args) {
+        Figure[] figures = new Figure[NUMBER_OF_FIGURES];
+        for (int i = 0; i < NUMBER_OF_FIGURES; i++) {
+            figures[i] = i >= NUMBER_OF_FIGURES / 2
+                    ? figureSupplier.getDefaultFigure() : figureSupplier.getRandomFigure();
```
liliia-ponomarenko, line 14:
```suggestion
                    ? figureSupplier.getDefaultFigure() 
                    : figureSupplier.getRandomFigure();

```

liliia-ponomarenko (approved):
Great job 👏

PR #1215, opened 2023-08-09
Created FigureSupplier class, created figures array

mmosii:


liliia-ponomarenko (changes requested):
good solution! Let’s  improve  it ;)

liliia-ponomarenko (changes requested):
Triggered  reject

liliia-ponomarenko (approved):
Great 😊

PR #1216, opened 2023-08-09
Implemented all tasks

teract10s:


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,27 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Main {
+    public static final Random RANDOM = new Random();
```
liliia-ponomarenko, line 6:
```suggestion
```
You don’t  use it in this class

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,27 @@
...
+                continue;
+            }
+            figures[i] = FIGURE_SUPPLIER.getDefaultFigure();
+        }
+        printFigures(figures);
+    }
+
+    private static void printFigures(Figure[] figures) {
+        for (Figure figure : figures) {
+            figure.draw();
```
liliia-ponomarenko, line 24: You can make this action in the first loop.

liliia-ponomarenko (changes requested):
triggered reject

liliia-ponomarenko (approved):
Great 😊

PR #1218, opened 2023-08-09
Classes for all figures were created, methods for random figure creat…

DenisGorda:
…ing were implemented.

### DIFF
--- a/src/main/java/core/basesyntax/Draw.java
+++ b/src/main/java/core/basesyntax/Draw.java
@@ -1,0 +1,5 @@
+package core.basesyntax;
+
+public interface Draw {
+    public abstract void draw();
+}
--- a/src/main/java/core/basesyntax/Figure.java
+++ b/src/main/java/core/basesyntax/Figure.java
@@ -1,0 +1,16 @@
+package core.basesyntax;
+
+public abstract class Figure implements Draw {
+    private String color;
+    private double square;
+
+    public Figure(String color) {
+        this.color = color;
+    }
+
+    public String getColor() {
+        return this.color;
+    }
+
+    public abstract double getArea();
+}
--- a/src/main/java/core/basesyntax/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/FigureSupplier.java
@@ -1,0 +1,40 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+class FigureSupplier {
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        Random random = new Random();
+        int figureType = random.nextInt(5);
+
+        switch (figureType) {
+            case 0:
+                return new Square(colorSupplier.getRandomColor(),
+                        random.nextDouble() * 10);
+            case 1:
+                return new Rectangle(colorSupplier.getRandomColor(),
+                        random.nextDouble() * 10,
+                        random.nextDouble() * 10);
+            case 2:
+                return new RightTriangle(colorSupplier.getRandomColor(),
+                        random.nextDouble() * 10,
+                        random.nextDouble() * 10);
+            case 3:
+                return new Circle(colorSupplier.getRandomColor(),
+                        random.nextDouble() * 10);
+            case 4:
+                return new IsoscelesTrapezoid(colorSupplier.getRandomColor(),
+                        random.nextDouble() * 10,
+                        random.nextDouble() * 10,
+                        random.nextDouble() * 10);
+            default:
+                return null;
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle("white", 10);
+    }
+}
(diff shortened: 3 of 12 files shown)


src/main/java/core/basesyntax/Draw.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Draw {
+    public abstract void draw();
```
liliia-ponomarenko, line 4:
```suggestion
    void draw();
``` 
this method  public abstract by default

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,40 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+class FigureSupplier {
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        Random random = new Random();
```
liliia-ponomarenko, line 9: Random also should  make class level variable

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,40 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+class FigureSupplier {
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        Random random = new Random();
+        int figureType = random.nextInt(5);
```
liliia-ponomarenko, line 10: all magic numbers should be constant

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,40 @@
...
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        Random random = new Random();
+        int figureType = random.nextInt(5);
+
+        switch (figureType) {
+            case 0:
+                return new Square(colorSupplier.getRandomColor(),
+                        random.nextDouble() * 10);
```
liliia-ponomarenko, line 15: why do you need to multiply each parameter

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,40 @@
...
+            case 3:
+                return new Circle(colorSupplier.getRandomColor(),
+                        random.nextDouble() * 10);
+            case 4:
+                return new IsoscelesTrapezoid(colorSupplier.getRandomColor(),
+                        random.nextDouble() * 10,
+                        random.nextDouble() * 10,
+                        random.nextDouble() * 10);
+            default:
+                return null;
```
liliia-ponomarenko, line 33: [Don’t return null from a method.](https://mate-academy.github.io/jv-program-common-mistakes/java-core/abstract-class-interface/oop-advanced#dont-return-null-from-a-method)

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,20 @@
...
+            figures[i] = figureSupplier.getRandomFigure();
+        }
+
+        for (int i = 3; i < 6; i++) {
+            figures[i] = figureSupplier.getDefaultFigure();
+        }
+
+        for (Figure figure : figures) {
+            figure.draw();
+        }
```
liliia-ponomarenko, line 18: Can you combine it in one loop?

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+public abstract class Figure implements Draw {
+    private String color;
+    private double square;
```
liliia-ponomarenko, line 5: Do you use this field?

liliia-ponomarenko (changes requested):
Good job! Let’s improve your solution ;)

src/main/java/core/basesyntax/FigureSupplier.java
DenisGorda: I thought that random,nextDouble() returns value between 0 and 1.0, that's why I use multiplication in every case. Now I will try to rewrite code witout this.

src/main/java/core/basesyntax/Draw.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Draw {
```
liliia-ponomarenko, line 3:
```suggestion
public interface Drawable {

```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,41 @@
...
+            case 3:
+                return new Circle(colorSupplier.getRandomColor(),
+                        random.nextDouble());
+            case 4:
+                return new IsoscelesTrapezoid(colorSupplier.getRandomColor(),
+                        random.nextDouble(),
+                        random.nextDouble(),
+                        random.nextDouble());
+            default:
+                return new Circle("white", 10);
```
liliia-ponomarenko, line 34:
```suggestion
                return getDefaultFigure();
```

liliia-ponomarenko (approved):
Great! Pay attention to the last recommendations ;)

PR #1234, opened 2023-08-14
homework

RomanYuzviak:


src/main/java/core/basesyntax/HelloWorld.java
```diff
@@ -1,8 +1,31 @@
 package core.basesyntax;
 
+import core.basesyntax.figures.Figure;
+import core.basesyntax.figures.figurehandlers.FigureSupplier;
 /**
  * Feel free to remove this class and create your own.
  */
+
 public class HelloWorld {
```
NazarSmith228, line 9: take a look at the comment :)

you should better remove this class and create your own with a `main` method

src/main/java/core/basesyntax/HelloWorld.java
```diff
@@ -1,8 +1,31 @@
...
 
+    private static final FigureSupplier figureSupplier = new FigureSupplier();
```
NazarSmith228, line 11: there is no need to declare this static variable

since you're using `FigureSupplier` in the `main` method - you can create its instance directly there

making such variables makes sense when you use them in methods, that are often invoked (which is not the case of `main` method) - this way you avoid creation of new instances each time the method is invoked

see [here](https://mate-academy.github.io/jv-program-common-mistakes/java-core/abstract-class-interface/oop-advanced#think-about-which-variables-should-be-local-in-the-method-and-which-should-be-class-level)

src/main/java/core/basesyntax/HelloWorld.java
NazarSmith228, line 9: take a look at the comment :)

you should better remove this class and create your own with a `main` method

src/main/java/core/basesyntax/HelloWorld.java
```diff
@@ -1,8 +1,31 @@
...
+    private static int ARRAY_LENGTH = 6;
```
NazarSmith228, line 12: this is not an informative constant name - what array? 

it would be better to name it like `NUMBER_OF_FIGURES`

src/main/java/core/basesyntax/HelloWorld.java
```diff
@@ -1,8 +1,31 @@
...
+        Figure[] figures = new Figure[ARRAY_LENGTH];
+
+        for (int i = 0; i < ARRAY_LENGTH; i++) {
+            if (i < 3) {
+                figures[i] = figureSupplier.getRandomFigure();
+            } else {
+                figures[i] = figureSupplier.getDefaultFigure();
+            }
+        }
+
```
NazarSmith228, line 25: there is no need to add blank lines between your code
```suggestion
        Figure[] figures = new Figure[ARRAY_LENGTH];
        for (int i = 0; i < ARRAY_LENGTH; i++) {
            if (i < 3) {
                figures[i] = figureSupplier.getRandomFigure();
            } else {
                figures[i] = figureSupplier.getDefaultFigure();
            }
        }
```

src/main/java/core/basesyntax/HelloWorld.java
```diff
@@ -1,8 +1,31 @@
...
+        for (Figure figure: figures) {
+            figure.draw();
+        }
```
NazarSmith228, line 28: you can do it in the loop above

src/main/java/core/basesyntax/figures/Circle.java
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    @Override
+    public double getArea() {
+        return Math.PI * radius * radius;
+    }
+
+    @Override
+    public void draw() {
+        StringBuilder stringBuilder = new StringBuilder();
```
NazarSmith228, line 18: there is no need to use `StringBuilder` here - additional cost

you can pass the formatted `String` directly into `println()`

**same should be applied to all other methods**

src/main/java/core/basesyntax/figures/figurehandlers/FigureSupplier.java
```diff
@@ -0,0 +1,60 @@
...
+import core.basesyntax.figures.IsoscelesTrapezoid;
+import core.basesyntax.figures.Rectangle;
+import core.basesyntax.figures.RightTriangle;
+import core.basesyntax.figures.Square;
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int FIGURES_COUNT = 4; //from 0 to 4
+    private final Random random = new Random();
+
```
NazarSmith228, line 14: redundant blank line :)

src/main/java/core/basesyntax/figures/figurehandlers/FigureSupplier.java
```diff
@@ -0,0 +1,60 @@
...
+                int side = random.nextInt(100) + 1;
+                return new Square(color, side);
+            }
+            default:
+                return new Circle("White", 10);
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle("White", 10);
```
NazarSmith228, line 58: I would suggest to make this circle a constant or a class-level variable so that you don't create a new one each time this method is invoked

src/main/java/core/basesyntax/figures/figurehandlers/FigureSupplier.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int FIGURES_COUNT = 4; //from 0 to 4
+    private final Random random = new Random();
+
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        int randomIndex = random.nextInt(FIGURES_COUNT) + 1;
```
NazarSmith228, line 18: there is no sense to assign `FIGURES_COUNT` to `4` and then increment it here - why not assigning it to `5`?

src/main/java/core/basesyntax/figures/figurehandlers/FigureSupplier.java
```diff
@@ -0,0 +1,60 @@
...
+                String color = colorSupplier.getRandomColor();
+                int radius = random.nextInt(30) + 1;
+                return new Circle(color, radius);
+            }
+            case 2: {
+                String color = colorSupplier.getRandomColor();
+                int lowBase = random.nextInt(30) + 1;
+                int highBase = random.nextInt(30) + 31;
+                int leg = random.nextInt(20) + 1;
+                int height = random.nextInt(10) + 1;
```
NazarSmith228, line 31: you have way to many magic numbers in each case

i would suggest not to care that much about random values for each figure property but rather create a simple method that would return a random number, bounded by some max number (for example `100`) and use it for all properties at the same time

src/main/java/core/basesyntax/figures/figurehandlers/FigureSupplier.java
```diff
@@ -0,0 +1,60 @@
...
+                return new IsoscelesTrapezoid(color, lowBase, highBase, leg, height);
+            }
+            case 3: {
+                String color = colorSupplier.getRandomColor();
+                int hypotenuse = random.nextInt(60) + 31;
+                int firstLeg = random.nextInt(30) + 1;
+                int secondLeg = random.nextInt(20) + 1;
+                return new RightTriangle(color, hypotenuse, firstLeg, secondLeg);
```
NazarSmith228, line 39: it might have sense to extract creation of each figure to separate methods, so that your `switch-case` statement will be more readable

NazarSmith228 (changes requested):
Good job! Now it's time to improve your solution

src/main/java/core/basesyntax/figures/figurehandlers/FigureSupplier.java
RomanYuzviak: I can`t use `FIGURES_COUNT`  with value 5 properly in nextInt() method I think, because it will return values from 0 to 5

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,24 @@
+package core.basesyntax;
+
+import core.basesyntax.figures.Figure;
+import core.basesyntax.figures.figurehandlers.FigureSupplier;
+
+public class Main {
+
```
liliia-ponomarenko, line 7: Redundant  empty line

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,24 @@
...
+            if (i < 3) {
+                figures[i] = figureSupplier.getRandomFigure();
+                figures[i].draw();
+            } else {
+                figures[i] = figureSupplier.getDefaultFigure();
+                figures[i].draw();
+            }
+        }
+    }
+
```
liliia-ponomarenko, line 23: Redundant empty line

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,24 @@
...
+        final FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] figures = new Figure[NUMBER_OF_FIGURES];
+        for (int i = 0; i < NUMBER_OF_FIGURES; i++) {
+            if (i < 3) {
+                figures[i] = figureSupplier.getRandomFigure();
+                figures[i].draw();
+            } else {
+                figures[i] = figureSupplier.getDefaultFigure();
+                figures[i].draw();
+            }
```
liliia-ponomarenko, line 20:
```suggestion
            } else {
                figures[i] = figureSupplier.getDefaultFigure();
            }
            figures[i].draw();
```
You can avoid code duplication

src/main/java/core/basesyntax/figures/figurehandlers/Colors.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax.figures.figurehandlers;
+
+public enum Colors {
```
liliia-ponomarenko, line 3: use singular  for class (enum) names

src/main/java/core/basesyntax/figures/figurehandlers/FigureSupplier.java
```diff
@@ -0,0 +1,89 @@
...
+    private Rectangle getRandomRectangle(String color, int firstSide, int secondSide) {
+        return new Rectangle(color, firstSide, secondSide);
+    }
+
+    private Square getRandomSquare(String color, int side) {
+        return new Square(color, side);
+    }
+
+    private int getRandomValue() {
+        return random.nextInt(RANDOM_VALUES_BORDER + 1) + 1;
```
liliia-ponomarenko, line 87:
```suggestion
        return random.nextInt(RANDOM_VALUES_BORDER);
```
Why do you need it?

src/main/java/core/basesyntax/figures/figurehandlers/FigureSupplier.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    private IsoscelesTrapezoid getRandomIsoscelesTrapezoid(String color, int lowBase,
+                                                           int highBase, int leg, int height) {
+        return new IsoscelesTrapezoid(color, lowBase, highBase, leg, height);
+    }
+
+    private RightTriangle getRandomRightTriangle(String color, int hypotenuse,
+                                                 int firstLeg, int secondLeg) {
+
```
liliia-ponomarenko, line 74: Redundant  empty line

src/main/java/core/basesyntax/figures/figurehandlers/FigureSupplier.java
```diff
@@ -0,0 +1,89 @@
...
+public class FigureSupplier {
+    private static final int FIGURES_COUNT = 5;
+    private static final int RANDOM_VALUES_BORDER = 30;
+    private final Figure defaultFigure = new Circle("White", 10);
+
+    private final Random random = new Random();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        int randomIndex = random.nextInt(FIGURES_COUNT) + 1;
```
liliia-ponomarenko, line 20:
```suggestion
        int randomIndex = random.nextInt(FIGURES_COUNT);
```
just numerate cases from 0 to 4

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,24 @@
...
+
+public class Main {
+
+    private static final int NUMBER_OF_FIGURES = 6;
+
+    public static void main(String[] args) {
+        final FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] figures = new Figure[NUMBER_OF_FIGURES];
+        for (int i = 0; i < NUMBER_OF_FIGURES; i++) {
+            if (i < 3) {
```
liliia-ponomarenko, line 14:
```suggestion
            if (i < NUMBER_OF_FIGURES / 2) {
```

liliia-ponomarenko (changes requested):
Good! Let’s fix a few minor mistakes ;)

src/main/java/core/basesyntax/figures/figurehandlers/Colors.java
RomanYuzviak: You mean Color instead of Colors?

src/main/java/core/basesyntax/figures/figurehandlers/FigureSupplier.java
Andriy-dz: not fixed

src/main/java/core/basesyntax/figures/figurehandlers/FigureSupplier.java
```diff
@@ -0,0 +1,88 @@
...
+                int secondSide = getRandomValue();
+                return new Rectangle(color, firstSide, secondSide);
+            }
+            case 4: {
+                String color = colorSupplier.getRandomColor();
+                int side = getRandomValue();
+                return new Square(color, side);
+            }
+            default:
+                return getDefaultFigure();
```
Andriy-dz, line 55: you create a color in each case, I would bring this field to the beginning and use it in each case. I also recommend not creating a variable for each random field and calling it immediately in the method parameters

src/main/java/core/basesyntax/figures/Square.java
```diff
@@ -0,0 +1,24 @@
+package core.basesyntax.figures;
+
+public class Square extends Figure {
+    private int side;
```
Andriy-dz, line 4: better to use double. Same in another cases. you also need to use nextDouble instead of nextInt

Andriy-dz (approved):

PR #1238, opened 2023-08-14
solution added

anastasiia-fi:


NazarSmith228 (changes requested):
Java CI build failed

src/main/java/core/basesyntax/AreaCounter.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface AreaCounter {
```
liliia-ponomarenko, line 3:
```suggestion
public interface AreaCalculator {

```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private final Random random = new Random();
+
+    public String getRandomColor() {
+        int index = new Random().nextInt(Colour.values().length);
+        return Colour.values()[index].toString().toLowerCase();
```
liliia-ponomarenko, line 10: Let’s  use method `name()`

src/main/java/core/basesyntax/Colour.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public enum Colour {
```
liliia-ponomarenko, line 3:
```suggestion
public enum Color {

```

src/main/java/core/basesyntax/Colour.java
```diff
@@ -0,0 +1,13 @@
...
+    BLUE,
+    PINK,
+    GREY,
+    YELLOW,
+    BLACK,
+    WHITE,
+    PURPLE,
+    ORANGE
+
```
liliia-ponomarenko, line 12: Redundant  empty line

src/main/java/core/basesyntax/Drawable.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Drawable {
+    public void draw();
```
liliia-ponomarenko, line 4:
```suggestion
  void draw();
```
this method is public by default

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public abstract class Figure implements Drawable, AreaCounter {
+    private final String colour;
```
liliia-ponomarenko, line 4:
```suggestion
    private final String color;
```
Check  it everywhere

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,37 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final String COLOUR = "white";
+    private static final int RADIUS = 10;
+
```
liliia-ponomarenko, line 8: Redundant empty line

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final String COLOUR = "white";
+    private static final int RADIUS = 10;
+
+    private final Random random = new Random();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final int randomFigureSide = random.nextInt(13);
```
liliia-ponomarenko, line 11: make 13 a constant variable, but `randomFigureSide` should  be the local variable

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public Figure getDefaultFigure() {
+        return new Circle(COLOUR, RADIUS);
+    }
+
+    public Figure getRandomFigure() {
+        String randomColor = colorSupplier.getRandomColor();
+        int randomFigureNumber = random.nextInt(6);
```
liliia-ponomarenko, line 19:
```suggestion
        int randomFigureNumber = random.nextInt(6);

```
Why 6? It’s also a magic number

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+public class Main {
+    private static FigureSupplier figureSupplier = new FigureSupplier();
```
liliia-ponomarenko, line 4: there is no need to declare this static variable

since you're using FigureSupplier in the main method - you can create its instance directly there

making such variables makes sense when you use them in methods, that are often invoked (which is not the case of main method) - this way you avoid creation of new instances each time the method is invoked

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,17 @@
...
+
+    public static void main(String[] args) {
+        Figure[] figures = new Figure[6];
```
liliia-ponomarenko, line 7: 6 - magic number

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,17 @@
...
+        for (int i = 0; i < figures.length; i++) {
+            if (i < 3) {
```
liliia-ponomarenko, line 9:
```suggestion
            if (i < figures.length / 2) {
```
3 - magic number, but better use expression

liliia-ponomarenko (changes requested):
Good start! Let’s improve your solution ;)

Elena-Bruyako (changes requested):
Good job! Let's improve your code

src/main/java/core/basesyntax/AreaCalculator.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface AreaCalculator {
+    public double getArea();
```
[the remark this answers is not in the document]
Andriy-dz: not fixed

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,18 @@
+package core.basesyntax;
+
+public class Main {
+    static final int TOTAL_FIGURES_AMOUNT = 6;
```
Andriy-dz, line 4: don't forget access modifiers

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,21 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+    private int radius;
```
Andriy-dz, line 4: why int? make it double) Same in another cases

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,38 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final String COLOUR = "white";
+    private static final int RADIUS = 10;
+    private static final int MAXSIZE = 13;
+    private static final int TOTAL_FIGURE_AMOUNT = 6;
```
Andriy-dz, line 9: why 6? you have only 5 figures

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,38 @@
...
+                return new Circle(randomColour, randomFigureSide);
+            case 1:
+                return new Square(randomColour, randomFigureSide);
+            case 2:
+                return new RightTriangle(randomColour, randomFigureSide, randomFigureSide);
+            case 3:
+                return new Rectangle(randomColour, randomFigureSide, randomFigureSide);
+            case 4:
+                return new IsoscelesTrapezoid(randomColour, randomFigureSide,
+                        randomFigureSide, randomFigureSide);
```
Andriy-dz, line 33: if you have a figure with 2 or more sides, then they will always be different in size. It is better to use random.nextDouble() * 50 everywhere - these will be values from 0 to 50

Andriy-dz (changes requested):
let's improve the code a little more

Andriy-dz (changes requested):
not all comments are fixed

src/main/java/core/basesyntax/FigureSupplier.java
anastasiia-fi: it was in task..
"In the main() method we need to create an array of figures (the size of array can be 3 or 6, it doesn't matter). "

NazarSmith228 (changes requested):
Pay more attention to the comments, left by mentors

Not all of them are fixed :)

src/main/java/core/basesyntax/FigureSupplier.java
Elena-Bruyako: it means how many figures you created it will be size of your array. you created 5 figures

Elena-Bruyako (approved):

PR #1239, opened 2023-08-14
Dev

romazan07:
implement task

### DIFF
--- a/src/main/java/core/basesyntax/Main.java
+++ b/src/main/java/core/basesyntax/Main.java
@@ -1,20 +1,20 @@
 package core.basesyntax;
 
 import core.basesyntax.suppliers.FigureSupplier;
-import java.util.Arrays;
 
 public class Main {
+    private static final int NUMBER_FIGURE = 6;
 
     public static void main(String[] args) {
         FigureSupplier figureSupplier = new FigureSupplier();
-        Figure[] arrFigure = new Figure[6];
+        Figure[] arrFigure = new Figure[NUMBER_FIGURE];
         for (int i = 0; i < arrFigure.length; i++) {
-            if (i != arrFigure.length / 2) {
+            if (i < arrFigure.length / 2) {
                 arrFigure[i] = figureSupplier.getRandomFigure();
             } else {
                 arrFigure[i] = figureSupplier.getDefaultFigure();
             }
+            arrFigure[i].drawFigure();
         }
-        System.out.println(Arrays.toString(arrFigure));
     }
 }
--- a/src/main/java/core/basesyntax/figure/Circle.java
+++ b/src/main/java/core/basesyntax/figure/Circle.java
@@ -10,21 +10,13 @@ public class Circle extends Figure {
         this.radius = radius;
     }
 
-    public void setRadius(int radius) {
-        this.radius = radius;
-    }
-
-    public double getRadius() {
-        return radius;
-    }
-
     @Override
     public double getArea() {
         return radius * Math.PI;
     }
 
     @Override
-    public void toDrawFigure() {
+    public void drawFigure() {
         System.out.println("Figure: circle, "
                 + "area: " + getArea() + " sq.units, "
                 + "radius: " + radius + " units, "
--- a/src/main/java/core/basesyntax/interfaces/DrawFigure.java
+++ b/src/main/java/core/basesyntax/interfaces/DrawFigure.java
@@ -1,6 +1,0 @@
-package core.basesyntax.interfaces;
-
-public interface DrawFigure {
-
-    void toDrawFigure();
-}
--- a/src/main/java/core/basesyntax/interfaces/DrawingFigure.java
+++ b/src/main/java/core/basesyntax/interfaces/DrawingFigure.java
@@ -1,0 +1,5 @@
+package core.basesyntax.interfaces;
+
+public interface DrawingFigure {
+    void drawFigure();
+}
--- a/src/main/java/core/basesyntax/suppliers/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/suppliers/FigureSupplier.java
@@ -9,36 +9,36 @@ import core.basesyntax.figure.Square;
 import java.util.Random;
 
 public class FigureSupplier {
+    private static int QUANTITY_FIGURES = 5;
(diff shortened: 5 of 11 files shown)


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,20 @@
...
+import core.basesyntax.suppliers.FigureSupplier;
+import java.util.Arrays;
+
+public class Main {
+
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] arrFigure = new Figure[6];
+        for (int i = 0; i < arrFigure.length; i++) {
+            if (i != arrFigure.length / 2) {
```
liliia-ponomarenko, line 12: Does your solution has expected behavior? I guess not.
```suggestion
            if (i < arrFigure.length / 2) {
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,20 @@
...
+                arrFigure[i] = figureSupplier.getRandomFigure();
+            } else {
+                arrFigure[i] = figureSupplier.getDefaultFigure();
+            }
+        }
+        System.out.println(Arrays.toString(arrFigure));
```
liliia-ponomarenko, line 18: use method toDrawFigure() for printing each figure

src/main/java/core/basesyntax/figure/Circle.java
```diff
@@ -0,0 +1,33 @@
...
+        return radius;
+    }
+
+    @Override
+    public double getArea() {
+        return radius * Math.PI;
+    }
+
+    @Override
+    public void toDrawFigure() {
```
liliia-ponomarenko, line 27:
```suggestion
    public void drawFigure() {
```

src/main/java/core/basesyntax/interfaces/DrawFigure.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax.interfaces;
+
+public interface DrawFigure {
+
+    void toDrawFigure();
```
liliia-ponomarenko, line 5:
```suggestion
public interface Drawble {
    void drawFigure();
```
Don’t use verbs for class names. Remove redundant empty lines. Avoid  using  articles  in variable (method) names

src/main/java/core/basesyntax/suppliers/FigureSupplier.java
```diff
@@ -0,0 +1,44 @@
...
+import core.basesyntax.figure.Square;
+import java.util.Random;
+
+public class FigureSupplier {
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public double getRandomlength() {
+        return random.nextDouble();
+    }
```
liliia-ponomarenko, line 17: this method can be private (and placed after public methods)

src/main/java/core/basesyntax/suppliers/FigureSupplier.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public Figure getRandomFigure() {
+        int indexOfFigure = random.nextInt(5);
```
liliia-ponomarenko, line 20: 5 - should  be constant 
```suggestion
        int figureIndex = random.nextInt(5);
```

src/main/java/core/basesyntax/suppliers/FigureSupplier.java
```diff
@@ -0,0 +1,44 @@
...
+import core.basesyntax.figure.Rectangle;
+import core.basesyntax.figure.RightTriangle;
+import core.basesyntax.figure.Square;
+import java.util.Random;
+
+public class FigureSupplier {
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public double getRandomlength() {
```
liliia-ponomarenko, line 15:
```suggestion
    public double getRandomLength() {
```

liliia-ponomarenko (changes requested):
Good start! Let’s improve your solution ;)

src/main/java/core/basesyntax/Main.java
romazan07: What's the difference?  🤷🏻‍♂️

src/main/java/core/basesyntax/suppliers/FigureSupplier.java
romazan07: Why private?

src/main/java/core/basesyntax/Main.java
Elena-Bruyako: in your code if i = 5 the if statement will return true
we need to get elements that less then arr.length / 2 = 3 so need to get 0, 1, 2 elements from array

src/main/java/core/basesyntax/suppliers/FigureSupplier.java
Elena-Bruyako: encapsulation principle. object that will use your FigureSupplier, shouldn't know anything about how you calculate or create you figures, it should get and know only default figure or some figure without other details.

src/main/java/core/basesyntax/interfaces/DrawFigure.java
Elena-Bruyako: no

src/main/java/core/basesyntax/interfaces/DrawingFigure.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax.interfaces;
+
+public interface DrawingFigure {
+    void drawFigure();
```
Elena-Bruyako, line 4: what if you are going to draw not only figure but other objects and using the same interface and method? it would be better to nor clarify what exactly you draw so you can rename interface to Drawable and method to draw.  It helps to implement polymorphism principle.

Elena-Bruyako (approved):

src/main/java/core/basesyntax/interfaces/DrawingFigure.java
romazan07: But this task is about figures. Or do you mean we can use this method in another task in the future?

PR #1241, opened 2023-08-16
Added solution

AlexRosenkreuz:
Created 5 figure class, abstract super class Figures and two interfaces for stating their behavior. Also created ColorSupplier and FigureSupplier classes for creating random figures. In method main implemented block of code that creates 23 random figures and 3 default figures.

### DIFF
--- a/src/main/java/core/basesyntax/DrawFigure.java
+++ b/src/main/java/core/basesyntax/DrawFigure.java
@@ -1,0 +1,5 @@
+package core.basesyntax;
+
+public interface DrawFigure {
+    void draw();
+}
--- a/src/main/java/core/basesyntax/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/FigureSupplier.java
@@ -1,0 +1,33 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int FIGURES = 5;
+    private static final double DEFAULT_LENGTH = 40.0;
+    private static final int DEFAULT_RADIUS = 15;
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getDefaultFigure() {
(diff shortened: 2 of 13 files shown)


src/main/java/core/basesyntax/DrawFigure.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface DrawFigure {
```
NazarSmith228, line 3: bad interface name - it should not contain verbs

how about `Drawable`?

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,33 @@
...
+    public Figure getRandomFigure() {
+        int randomFigureIndex = random.nextInt(FIGURES);
+        String color = colorSupplier.getRandomColor();
+        switch (randomFigureIndex) {
+            case 0: return new Circle(color, getRandomUnits());
+            case 1: return new IsoscelesTrapezoid(color, getRandomUnits(),
+                    getRandomUnits(), getRandomUnits());
+            case 2: return new Rectangle(color, getRandomUnits(), getRandomUnits());
+            case 3: return new RightTriangle(color, getRandomUnits(), getRandomUnits());
+            case 4: return new Square(color, getRandomUnits());
```
NazarSmith228, line 25: passing expressions as constructor arguments decreases the readability

I would recommend to declare separate variables for each figure property and pass those variable into a constructor

NazarSmith228 (changes requested):
Great job! Just some little remarks to fix

src/main/java/core/basesyntax/FigureSupplier.java
AlexRosenkreuz: I tried to resolve this, but there are so much variables now. Is there any other way to do that? Or it's readable the way I wrote it now?

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #1243, opened 2023-08-18
Liliia hrebinyk hw oop advanced

Mnemozine-Oh:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private static final String[] COLORS = {"blue", "green", "pink","red", "black", "white"};
```
olekskov, line 6: It's better to introduce separate enum class instead of this constant

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,17 @@
...
+    public static void main(String[] args) {
+        Figure[] randomFiguresArray = new Figure[RANDOM_FIGURES_ARRAY_SIZE];
+        randomFiguresArray[RANDOM_FIGURES_ARRAY_SIZE - 1] = figureSupplier.getDefaultFigure();
+        for (int i = 0; i < (randomFiguresArray.length - 1); i++) {
+            randomFiguresArray[i] = figureSupplier.getRandomFigure();
+        }
+        for (int i = 0; i < randomFiguresArray.length; i++) {
+            randomFiguresArray[i].draw();
+        }
+    }
```
olekskov, line 16: It's better to use 1 cycle for all logic and also make code dependant on single constant value in order to change only 1 number in code to make it work for 8, 10 or 999 balls properly
```suggestion
    public static void main(String[] args) {
        Figure[] randomFiguresArray = new Figure[RANDOM_FIGURES_ARRAY_SIZE];
        for (int i = 0; i < (randomFiguresArray.length); i++) {
            if (i < randomFiguresArray.length / 2) {
                randomFiguresArray[i] = figureSupplier.getRandomFigure();
            } else {
                randomFiguresArray[i] = figureSupplier.getDefaultFigure();
            }
            randomFiguresArray[i].draw();
        }
    }
```

olekskov (changes requested):
Good job overall.
Please check comments and recheck build failure due to checkstyle (please run mvn clean package command locally and check results)

PR #1244, opened 2023-08-22
homework-solution

Yaromyros:
In the main class, through cycles, I created 3 random shapes and 3 default ones that are stored in an array. The parent class is Figure, which is abstract and implements two drawing and calculation interfaces. All other classes extend to the parent class. 

ColorSupplier class returns random color from the enum. 

Figure supplier class creates random figures and default figures.

All subclasses of parental have a constructor to create objects and implementation of two interfaces, if some class have no implementation it will return the result of Figure class implementation.

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -1,0 +1,23 @@
+package core.basesyntax;
+
+public class Application {
+    private static final int NUMBER_SIX = 6;
+    private static final int NUMBER_TWO = 2;
+
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] figures = new Figure[NUMBER_SIX];
+
+        for (int i = 0; i < figures.length / NUMBER_TWO; i++) {
+            figures[i] = figureSupplier.getRandomFigure();
+        }
+
+        for (int i = figures.length / NUMBER_TWO; i < figures.length; i++) {
+            figures[i] = figureSupplier.getDefaultFigure();
+        }
+
+        for (Figure figure : figures) {
+            figure.draw();
+        }
+    }
+}
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -1,23 +1,19 @@
 package core.basesyntax;
 
 public class Application {
-    private static final int NUMBER_SIX = 6;
-    private static final int NUMBER_TWO = 2;
 
     public static void main(String[] args) {
         FigureSupplier figureSupplier = new FigureSupplier();
-        Figure[] figures = new Figure[NUMBER_SIX];
+        Figure[] figures = new Figure[6];
 
-        for (int i = 0; i < figures.length / NUMBER_TWO; i++) {
-            figures[i] = figureSupplier.getRandomFigure();
-        }
+        for (int i = 0; i < figures.length; i++) {
+            if (i < figures.length / 2) {
+                figures[i] = figureSupplier.getRandomFigure();
+            } else {
+                figures[i] = figureSupplier.getDefaultFigure();
+            }
 
-        for (int i = figures.length / NUMBER_TWO; i < figures.length; i++) {
-            figures[i] = figureSupplier.getDefaultFigure();
-        }
-
-        for (Figure figure : figures) {
-            figure.draw();
+            figures[i].draw();
         }
     }
 }
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -1,7 +1,6 @@
 package core.basesyntax;
 
 public class Application {
-
     public static void main(String[] args) {
         FigureSupplier figureSupplier = new FigureSupplier();
         Figure[] figures = new Figure[6];
--- a/src/main/java/core/basesyntax/Color.java
+++ b/src/main/java/core/basesyntax/Color.java
@@ -1,14 +1,14 @@
 package core.basesyntax;
 
 public enum Color {
-    black,
-    blue,
-    yellow,
-    red,
-    orange,
-    brown,
-    grey,
-    white,
-    green,
-    purple
+    BLACK,
+    BLUE,
+    YELLOW,
+    RED,
+    ORANGE,
+    BROWN,
+    GREY,
+    WHITE,
+    GREEN,
+    PURPLE
 }
--- a/src/main/java/core/basesyntax/Color.java
+++ b/src/main/java/core/basesyntax/Color.java
@@ -1,0 +1,14 @@
+package core.basesyntax;
+
+public enum Color {
+    black,
+    blue,
+    yellow,
+    red,
+    orange,
+    brown,
+    grey,
+    white,
+    green,
+    purple
+}
--- a/src/main/java/core/basesyntax/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/FigureSupplier.java
@@ -3,12 +3,12 @@ package core.basesyntax;
 import java.util.Random;
 
 public class FigureSupplier {
-    private final Random random = new Random();
     private static final int NUMBER_OF_BOUND = 6;
     private static final int DEFAULT_SIZE = 10;
     private static final String DEFAULT_COLOR = "WHITE";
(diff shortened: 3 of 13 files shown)


src/main/java/core/basesyntax/Application.java
```diff
@@ -0,0 +1,23 @@
+package core.basesyntax;
+
+public class Application {
+    private static final int NUMBER_SIX = 6;
+    private static final int NUMBER_TWO = 2;
```
okuzan, line 5: this is probably not worth it to extract to a constant :)

src/main/java/core/basesyntax/Color.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public enum Color {
+    black,
```
okuzan, line 4:
```suggestion
    BLACK,
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,59 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final Random random = new Random();
+    private static final Random randomNumber = new Random();
```
```suggestion
    private final Random random = new Random();
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,59 @@
...
+                break;
+            }
+            case 5: {
+                randomFigure = new IsoscelesTrapezoid(
+                        randomNumber.nextDouble() * TEN_NUMBER + ONE_NUMBER,
+                        randomNumber.nextDouble() * TEN_NUMBER + ONE_NUMBER,
+                        randomNumber.nextDouble() * TEN_NUMBER + ONE_NUMBER,
+                        colorSupplier.getRandomColor());
+                break;
+            }
```
okuzan, line 48: extract creation of each figure to a dedicated private method

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,59 @@
...
+            default: {
+                return figureSupplier.getDefaultFigure();
+            }
+        }
+        return randomFigure;
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(TEN_NUMBER, "white");
```
okuzan, line 57:
```suggestion
        return new Circle(DEFAULT_RADIUS, DEFAULT_COLOR);
```
reference your enum for color. Naming 10 as TEN_NUMBER constant doesn't achieve much - use the constant name as an opportunity to clarify its role and your intentions to the one who will read your code one day

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -0,0 +1,26 @@
+package core.basesyntax;
+
+public class Rectangle extends Figure {
+    private double firstSideLength;
+    private double secondSideLength;
+
+    public Rectangle(double firstSideLength, double secondSideLength, String color) {
+        this.firstSideLength = firstSideLength;
+        this.secondSideLength = secondSideLength;
+        setColor(color);
```
okuzan, line 10:
```suggestion
        super(color);
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,59 @@
...
+                        randomNumber.nextDouble() * TEN_NUMBER + ONE_NUMBER,
+                        colorSupplier.getRandomColor());
+                break;
+            }
+            case 4: {
+                randomFigure = new Circle(randomNumber.nextDouble() * TEN_NUMBER + ONE_NUMBER,
+                        colorSupplier.getRandomColor());
+                break;
+            }
+            case 5: {
```
okuzan, line 41: this case is unreachable, please pay attention to IDEA warnings. BOUND is exclusive for method nextInt (from documentation):
image

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,59 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final Random random = new Random();
+    private static final Random randomNumber = new Random();
+    private static final int NUMBER_OF_FIGURES = 5;
+    private static final int TEN_NUMBER = 10;
+    private static final int ONE_NUMBER = 1;
```
okuzan, line 10:
```suggestion
    private static final int DEFAULT_PROPERTY_LENGTH = 10;
```

src/main/java/core/basesyntax/RightTriangle.java
```diff
@@ -0,0 +1,25 @@
+package core.basesyntax;
+
+public class RightTriangle extends Figure {
+    private static final int NUMBER_TWO = 2;
+    private double sideLength;
+    private double heightLength;
```
okuzan, line 6:
```suggestion
    private double side;
    private double height;
```

src/main/java/core/basesyntax/RightTriangle.java
```diff
@@ -0,0 +1,25 @@
+package core.basesyntax;
+
+public class RightTriangle extends Figure {
+    private static final int NUMBER_TWO = 2;
```
okuzan, line 4: it's ok to hardcode here, rules of geometry won't change anytime soon

src/main/java/core/basesyntax/Application.java
```diff
@@ -0,0 +1,23 @@
...
+            figures[i] = figureSupplier.getRandomFigure();
+        }
+
+        for (int i = figures.length / NUMBER_TWO; i < figures.length; i++) {
+            figures[i] = figureSupplier.getDefaultFigure();
+        }
+
+        for (Figure figure : figures) {
+            figure.draw();
+        }
```
okuzan, line 21: you can use one for loop here

okuzan (changes requested):

src/main/java/core/basesyntax/IsoscelesTrapezoid.java
```diff
@@ -0,0 +1,31 @@
+package core.basesyntax;
+
+public class IsoscelesTrapezoid extends Figure {
+    private static final int NUMBER_TWO = 2;
+    private double firstSideLength;
+    private double secondSideLength;
+    private double heightLength;
```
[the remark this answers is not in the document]
Yaromyros: повністю не згоден)

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,76 @@
...
+            case 5: {
+                randomFigure = createRandomIsoscelesTrapezoid(colorSupplier);
+                break;
+            }
+            default: {
+                return figureSupplier.getDefaultFigure();
+            }
+        }
+        return randomFigure;
+    }
```
olekskov, line 42: random.nextInt(NUMBER_OF_BOUND) will give us value from 0 to 5 (when NUMBER_OF_BOUND == 6).
this means that in about 17% of cases it will be 0 and as a result method will produce default figure
We need to start our cases with 0 and return NUMBER_OF_BOUND value back to 5.
With such changes default value will be used only if we messed up with code after adding new figures
```suggestion
        int indexOfColor = random.nextInt(NUMBER_OF_BOUND);
        Figure randomFigure;
        switch (indexOfColor) {
            case 0: {
                randomFigure = createRandomSquare(colorSupplier);
                break;
            }
            case 1: {
                randomFigure = createRandomRectangle(colorSupplier);
                break;
            }
            case 2: {
                randomFigure = createRandomRightTriangle(colorSupplier);
                break;
            }
            case 3: {
                randomFigure = createRandomCircle(colorSupplier);
                break;
            }
            case 4: {
                randomFigure = createRandomIsoscelesTrapezoid(colorSupplier);
                break;
            }
            default: {
                return figureSupplier.getDefaultFigure();
            }
        }
        return randomFigure;
    }
```

olekskov (changes requested):
Good job overall.
Need to fix last problem

src/main/java/core/basesyntax/Application.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax;
+
+public class Application {
+
```
kozhukhovsky, line 4:
```suggestion
```

kozhukhovsky (approved):

PR #1255, opened 2023-08-29
Implemented OOP Principles for Figure Classes and Random Figure Gener…

Petro-Smoliar:
…ation

### DIFF
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,0 +1,25 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private static final int MAX_NUMBER = 6;
+    private Random random = new Random();
+
+    public String getRandomColor() {
+        switch (random.nextInt(MAX_NUMBER)) {
+            case 0:
+                return "blue";
+            case 1:
+                return "red";
+            case 2:
+                return "black";
+            case 3:
+                return "white";
+            case 4:
+                return "green";
+            default:
+                return "grey";
+        }
+    }
+}
--- a/src/main/java/core/basesyntax/Figure.java
+++ b/src/main/java/core/basesyntax/Figure.java
@@ -1,0 +1,11 @@
+package core.basesyntax;
+
+public abstract class Figure implements AreaCalculator {
+    protected String color;
+
+    public Figure(String color) {
+        this.color = color;
+    }
+
+    public abstract void draw();
+}
--- a/src/main/java/core/basesyntax/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/FigureSupplier.java
@@ -1,0 +1,34 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int MAX_NUMBER = 5;
+    private static final int MAX_PARAMETR = 10;
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        String color = colorSupplier.getRandomColor();
+        int firstLeg = random.nextInt(MAX_PARAMETR);
+        int secondLeg = random.nextInt(MAX_PARAMETR);
+        int radius = random.nextInt(MAX_PARAMETR);
+        int height = random.nextInt(MAX_PARAMETR);
+        switch (random.nextInt(MAX_NUMBER)) {
+            case 0:
+                return new Circle(color,radius);
+            case 1:
+                return new IsoscelesTrapezoid(color, firstLeg,secondLeg,height);
+            case 2:
+                return new Rectangle(color,firstLeg,secondLeg);
+            case 3:
+                return new RightTriangle(color, firstLeg,secondLeg);
+            default:
+                return new Square(color, firstLeg);
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle("white", 10);
+    }
+}
--- a/src/main/java/core/basesyntax/Main.java
+++ b/src/main/java/core/basesyntax/Main.java
@@ -1,0 +1,19 @@
+package core.basesyntax;
+
+public class Main {
+
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] figures = new Figure[6];
+        for (int i = 0; i < figures.length; i++) {
+            if (i < figures.length / 2) {
+                figures[i] = figureSupplier.getRandomFigure();
+            } else {
+                figures[i] = figureSupplier.getDefaultFigure();
+            }
+        }
+        for (Figure figure: figures) {
+            figure.draw();
+        }
+    }
+}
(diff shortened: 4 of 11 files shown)


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,34 @@
...
+
+    public Figure getRandomFigure() {
+        String color = colorSupplier.getRandomColor();
+        int firstLeg = random.nextInt(MAX_PARAMETR);
+        int secondLeg = random.nextInt(MAX_PARAMETR);
+        int radius = random.nextInt(MAX_PARAMETR);
+        int height = random.nextInt(MAX_PARAMETR);
+        switch (random.nextInt(MAX_NUMBER)) {
+            case 0:
+                return new Circle(color,radius);
```
okuzan, line 19: autoformat your files, should be spaces after comma

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,19 @@
...
+        for (int i = 0; i < figures.length; i++) {
+            if (i < figures.length / 2) {
+                figures[i] = figureSupplier.getRandomFigure();
+            } else {
+                figures[i] = figureSupplier.getDefaultFigure();
+            }
+        }
+        for (Figure figure: figures) {
+            figure.draw();
+        }
```
okuzan, line 17: place inside the first loop

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,25 @@
...
+                return "red";
+            case 2:
+                return "black";
+            case 3:
+                return "white";
+            case 4:
+                return "green";
+            default:
+                return "grey";
+        }
```
okuzan, line 23: create enum for color

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,34 @@
...
+                return new Rectangle(color,firstLeg,secondLeg);
+            case 3:
+                return new RightTriangle(color, firstLeg,secondLeg);
+            default:
+                return new Square(color, firstLeg);
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle("white", 10);
```
okuzan, line 32: these values should be constants

okuzan (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
...
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private static final int MAX_NUMBER = 6;
+    private final Random random = new Random();
+
+    public String getRandomColor() {
+        return Color.values()[random.nextInt(MAX_NUMBER)].name();
+    }
```
okuzan, line 11: in this impl, when you add a new color, you will need to remember to come here and change this value or one color will never be generated. instead of 6 depend on length of color values

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,37 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+
+    public static final int FIX_RADIUS = 10;
+    private static final int MAX_NUMBER = 5;
+    private static final int MAX_PARAMETR = 10;
+    private static final String COLOR_WHITE = Color.WHITE.name();
```
okuzan, line 10: name of a constant should describe not its value, but its role
```suggestion
    private static final String DEFAULT_COLOR = Color.WHITE.name();

```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,37 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+
```
```suggestion
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,37 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+
+    public static final int FIX_RADIUS = 10;
```
```suggestion
    public static final int DEFAULT_RADIUS = 10;

```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,37 @@
...
+    private static final String COLOR_WHITE = Color.WHITE.name();
+    private final Random random = new Random();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        String color = colorSupplier.getRandomColor();
+        int firstLeg = random.nextInt(MAX_PARAMETR) + 1;
+        int secondLeg = random.nextInt(MAX_PARAMETR) + 1;
+        int radius = random.nextInt(MAX_PARAMETR) + 1;
+        int height = random.nextInt(MAX_PARAMETR) + 1;
```
okuzan, line 19: create a private utility method for that, Also, why not create it right inside of cases, this way we will create only the values we will use

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,18 @@
...
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] figures = new Figure[6];
+        for (int i = 0; i < figures.length; i++) {
+            if (i < figures.length / 2) {
+                figures[i] = figureSupplier.getRandomFigure();
+                figures[i].draw();
+            } else {
+                figures[i] = figureSupplier.getDefaultFigure();
+                figures[i].draw();
+            }
```
okuzan, line 15:
```suggestion
            } else {
                figures[i] = figureSupplier.getDefaultFigure();
            }
            figures[i].draw();
```

okuzan (changes requested):

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,37 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    public static final int DEFAULT_RADIUS = 10;
```
```suggestion
    private static final int DEFAULT_RADIUS = 10;
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+public class Main {
+
```
okuzan, line 4:
```suggestion
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,37 @@
...
+            default:
+                return new Square(color, getRandomValue());
+        }
+    }
+
+    private int getRandomValue() {
+        return random.nextInt(MAX_PARAMETR) + 1;
+    }
+
+    public Figure getDefaultFigure() {
```
okuzan, line 34: private methods should go after public ones

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public abstract class Figure implements AreaCalculator {
+    protected String color;
+
+    public Figure(String color) {
+        this.color = color;
+    }
+
+    public abstract void draw();
```
okuzan, line 10: ok as is, but if areacalculator is interface, maybe make draw an interface too

okuzan (changes requested):

kozhukhovsky (approved):

PR #1257, opened 2023-08-30
first commit

LeonIllya:


src/main/java/core/basesyntax/Application.java
```diff
@@ -0,0 +1,27 @@
+package core.basesyntax;
+
+import core.basesyntax.abstracts.Figure;
+import core.basesyntax.random.FigureSupplier;
+
+/**
+ * Feel free to remove this class and create your own.
+ */
```
okuzan, line 8: remove comments

src/main/java/core/basesyntax/Application.java
```diff
@@ -0,0 +1,27 @@
...
+public class Application {
+    private static final int DEFAULT_FIGURE = 6;
+
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] figures = new Figure[DEFAULT_FIGURE];
+
+        for (int i = 0; i < figures.length; i++) {
+            if (i <= 2) {
```
okuzan, line 17: use constant.length / 2

src/main/java/core/basesyntax/Application.java
```diff
@@ -0,0 +1,27 @@
...
+                figures[i] = figureSupplier.getDefaultFigure();
+                figures[i].draw();
```
okuzan, line 19: move draw() out if else - it's common for both clauses

src/main/java/core/basesyntax/random/ColorSupplier.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax.random;
+
+import core.basesyntax.enams.Color;
+import java.util.Random;
+
+public class ColorSupplier {
+    private final Random random = new Random();
+    private String color;
```
okuzan, line 8:
```suggestion
```
We don't need it as a field, we just use object returned from this method, we don't need to keep it as a state of the class that will change on each method invocation

src/main/java/core/basesyntax/random/ColorSupplier.java
```diff
@@ -0,0 +1,19 @@
...
+
+    public String getRandomColor() {
+        int value = random.nextInt(Color.values().length);
+        color = String.valueOf(Color.values()[value]);
+        return color;
```
okuzan, line 13:
```suggestion
        return Color.values()[index].name();
```

src/main/java/core/basesyntax/Application.java
```diff
@@ -0,0 +1,27 @@
...
+            } else {
+                figures[i] = figureSupplier.getRandomFigure();
+                figures[i].draw();
+            }
+        }
+    }
+
```
okuzan, line 26:
```suggestion
```

okuzan (changes requested):
Good job overall!

okuzan (changes requested):
Build not fixed

src/main/java/core/basesyntax/interfaces/FigureArea.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax.interfaces;
+
+public interface FigureArea {
```
aaananas, line 3:
```suggestion
public interface AreaCalculator {
```

aaananas (approved):

PR #1258, opened 2023-09-02
Jv oop advanced solution

willwildred:


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,20 @@
...
+
+
+    public static void main(String[] args) {
+        Figure[] figures = new Figure[6];
+        for (int i = 0; i < figures.length; i++) {
+            if (i < figures.length / 2) {
+                figureSupplier.getRandomFigure();
+            } else {
+                figureSupplier.getDefaultFigure();
+            }
```
olekskov, line 17: as per task requirement: 
"After generating the array, we need to display the entire list of objects that we have"

src/main/java/core/basesyntax/abstracts/Figure.java
```diff
@@ -0,0 +1,24 @@
...
+        return color;
+    }
+
+    public void setColor(String color) {
+        this.color = color;
+    }
+
+    public String setDecimalFormat(double number) {
+        return decimalFormat.format(number);
+    }
```
olekskov, line 23: 1. You have class variable decimalFormat and method setDecimalFormat which is obviously should be setter. But it's not, it even doesn't set anything, it returns formatted value as String. This makes naming absolutely confusing
2. You use this method inside child classes, why public then?
```suggestion
    protected String getFormattedDecimal(double number) {
        return decimalFormat.format(number);
    }
```

src/main/java/core/basesyntax/suppliers/ColorSupplier.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax.suppliers;
+
+import core.basesyntax.enums.Colors;
+import java.util.Random;
+
+public class ColorSupplier {
+    private final Random random = new Random();
+
+
```
olekskov, line 9:
```suggestion
```

src/main/java/core/basesyntax/suppliers/FigureSupplier.java
```diff
@@ -0,0 +1,68 @@
...
+    private final IsoscelesTrapezoid isoscelesTrapezoid = new IsoscelesTrapezoid();
+    private final Rectangle rectangle = new Rectangle();
+    private final RightTriangle rightTriangle = new RightTriangle();
+    private final Square square = new Square();
+    private final Random random = new Random();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private static final int ALL_FIGURES = 5;
+    private static final int DEF_PARAMETER = 10;
+
+
```
olekskov, line 19:
```suggestion
```

src/main/java/core/basesyntax/suppliers/FigureSupplier.java
```diff
@@ -0,0 +1,68 @@
...
+            case 4 : {
+                square.setSide(random.nextInt(DEF_PARAMETER));
+                square.setColor(colorSupplier.getRandomColor());
+                square.draw();
+                return square;
+            }
+        }
+        return allFigures[trueRandom];
+    }
+
```
olekskov, line 60: there is no sense to make 2-lines split between any blocks, please recheck whole code
```suggestion
```

src/main/java/core/basesyntax/suppliers/FigureSupplier.java
```diff
@@ -0,0 +1,68 @@
...
+    public Figure getRandomFigure() {
+        Figure[] allFigures = {circle, isoscelesTrapezoid, rectangle, rightTriangle, square};
+        int trueRandom = random.nextInt(ALL_FIGURES);
+        switch (trueRandom) {
+            case 0 : {
+                circle.setRadius(random.nextInt(DEF_PARAMETER));
+                circle.setColor(colorSupplier.getRandomColor());
+                circle.draw();
+                return circle;
```
olekskov, line 28: 1. Please create all-args constructors to simplify code
2. draw() method should be called in Main class after figure is added to array, not during generation
```suggestion
                return new Circle(random.nextInt(DEF_PARAMETER), colorSupplier.getRandomColor());
```

olekskov (changes requested):
Please run mvn clean package command locally and fix issues with build

src/main/java/core/basesyntax/abstracts/Figure.java
willwildred: probably most stupidest thing I ever done, I'm really very sorry you should read this :(

src/main/java/core/basesyntax/enums/Colors.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax.enums;
+
+public enum Colors {
```
aaananas, line 3:
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/figures/Rectangle.java
```diff
@@ -0,0 +1,28 @@
...
+import core.basesyntax.abstracts.Figure;
+
+public class Rectangle extends Figure {
+    private double length;
+    private double width;
+
+    public Rectangle(double length, double width, String color) {
+        this.length = length;
+        this.width = width;
+        setColor(color);
```
aaananas, line 12:
```suggestion
    public Rectangle(double length, double width, String color) {
        super(color);
        this.length = length;
        this.width = width;
```

src/main/java/core/basesyntax/suppliers/FigureSupplier.java
```diff
@@ -0,0 +1,50 @@
...
+            }
+            default : {
+                return new Square(random.nextInt(DEF_PARAMETER),
+                        colorSupplier.getRandomColor());
+            }
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(10, Colors.WHITE.name().toLowerCase());
```
aaananas, line 48: lets add constants

aaananas (changes requested):

src/main/java/core/basesyntax/enums/Colors.java
willwildred: I had your review before in another exercise, and you made enum "Color" to "Colors", now now the opposite, so how is right?

kozhukhovsky (approved):

PR #1263, opened 2023-09-09
add clases Square, Rectangle, RightTriangle, Circle, IsoscelesTrapezo…

VoitseshukV:
…id, ColorSupplier, FigureSupplier

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -1,0 +1,20 @@
+package core.basesyntax;
+
+public class Application {
+    public static final int FIGURE_COUNT = 6;
+    private static final FigureSupplier FIGURE_SUPPLIER = new FigureSupplier();
+
+    public static void main(String[] args) {
+        Figure[] figures = new Figure[6];
+        for (int i = 0; i < FIGURE_COUNT; i++) {
+            if (i < FIGURE_COUNT / 2) {
+                figures[i] = FIGURE_SUPPLIER.getRandomFigure();
+            } else {
+                figures[i] = FIGURE_SUPPLIER.getDefaultFigure();
+            }
+        }
+        for (Figure figure: figures) {
+            System.out.println(figure);
+        }
+    }
+}
--- a/src/main/java/core/basesyntax/Figure.java
+++ b/src/main/java/core/basesyntax/Figure.java
@@ -1,0 +1,27 @@
+package core.basesyntax;
+
+public abstract class Figure implements HasArea, HasPerimeter {
+    private Color color;
+
+    public Figure(Color color) {
+        this.color = color;
+    }
+
+    public Color getColor() {
+        return color;
+    }
+
+    public void setColor(Color color) {
+        this.color = color;
+    }
+
+    public abstract int area();
+
+    public abstract int perimeter();
+
+    @Override
+    public String toString() {
+        return "Figure: " + this.getClass().getSimpleName() + ", area: " + area()
+                + ", perimeter: " + perimeter();
+    }
+}
--- a/src/main/java/core/basesyntax/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/FigureSupplier.java
@@ -1,0 +1,39 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final Random RANDOM = new Random();
+    private static final ColorSupplier COLOR_SUPPLIER = new ColorSupplier();
+    private static final int FIGURES_COUNT = 5;
+    private static final int DEFAULT_RADIUS = 10;
+    private static final int MIN_SIZE = 1;
+    private static final int MAX_SIZE = 10;
+
+    private static int randomSize() {
+        return RANDOM.nextInt(MAX_SIZE - MIN_SIZE) + MIN_SIZE;
+    }
+
+    public Figure getRandomFigure() {
+        switch (RANDOM.nextInt(FIGURES_COUNT)) {
+            case 0:
+                return new Square(COLOR_SUPPLIER.getRandomColor(), randomSize());
+            case 1:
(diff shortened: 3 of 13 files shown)


src/main/java/core/basesyntax/Application.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax;
+
+public class Application {
+    public static final int FIGURE_COUNT = 6;
+    private static final FigureSupplier FIGURE_SUPPLIER = new FigureSupplier();
+
+    public static void main(String[] args) {
+        Figure[] figures = new Figure[6];
```
olekskov, line 8:
```suggestion
        Figure[] figures = new Figure[FIGURE_COUNT];
```

src/main/java/core/basesyntax/Application.java
```diff
@@ -0,0 +1,20 @@
...
+        for (int i = 0; i < FIGURE_COUNT; i++) {
+            if (i < FIGURE_COUNT / 2) {
+                figures[i] = FIGURE_SUPPLIER.getRandomFigure();
+            } else {
+                figures[i] = FIGURE_SUPPLIER.getDefaultFigure();
+            }
+        }
+        for (Figure figure: figures) {
+            System.out.println(figure);
+        }
```
olekskov, line 18: 1. according to task description we should use draw() method for printing info about figure
2. there is no need for second loop, we can do required within single one
```suggestion
        for (int i = 0; i < FIGURE_COUNT; i++) {
            if (i < FIGURE_COUNT / 2) {
                figures[i] = FIGURE_SUPPLIER.getRandomFigure();
            } else {
                figures[i] = FIGURE_SUPPLIER.getDefaultFigure();
            }
            figures[i].draw();
        }
```

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,27 @@
...
+
+    public abstract int area();
+
+    public abstract int perimeter();
+
+    @Override
+    public String toString() {
+        return "Figure: " + this.getClass().getSimpleName() + ", area: " + area()
+                + ", perimeter: " + perimeter();
+    }
```
olekskov, line 26: according to task we should not override toString() method but create separate interface (e.g. Drawable) and make void method in it that will print information about figure.
```suggestion
    @Override
    public void draw() {
        System.out.printl("Figure: " + this.getClass().getSimpleName() + ", area: " + area()
                + ", perimeter: " + perimeter());
    }
```
Please also include information about color

src/main/java/core/basesyntax/HasArea.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface HasArea {
+    int area();
```
olekskov, line 4: methods naming is much better when it contains verb
```suggestion
    int getArea();
```

src/main/java/core/basesyntax/HasPerimeter.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface HasPerimeter {
+    int perimeter();
```
olekskov, line 4:
```suggestion
    int getPerimeter();
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final Random RANDOM = new Random();
+    private static final ColorSupplier COLOR_SUPPLIER = new ColorSupplier();
+    private static final int FIGURES_COUNT = 5;
+    private static final int DEFAULT_RADIUS = 10;
+    private static final int MIN_SIZE = 1;
+    private static final int MAX_SIZE = 10;
```
olekskov, line 11: 1. Services we rely on are usually private final
2. Also services are not immutable classes so technically they cannot be constants even when private static final
```suggestion
    private static final int FIGURES_COUNT = 5;
    private static final int DEFAULT_RADIUS = 10;
    private static final int MIN_SIZE = 1;
    private static final int MAX_SIZE = 10;
    
    private final Random random = new Random();
    private final ColorSupplier colorSupplier = new ColorSupplier();
```

olekskov (changes requested):

src/main/java/core/basesyntax/FigureSupplier.java
VoitseshukV: Will this not lead to redundancy? Not in this case, but in general. One Random object is enough for any number of created FigureSupplier. I did not proceed from immutability, but from the fact that the constructor will be called only once.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,39 @@
...
+    private final Random random = new Random();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    private int randomSize() {
+        return random.nextInt(MAX_SIZE - MIN_SIZE) + MIN_SIZE;
+    }
```
okuzan, line 15: private methods should go after public ones

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,39 @@
...
+            case 4:
+                return new IsoscelesTrapezoid(colorSupplier.getRandomColor(), randomSize(),
+                        randomSize(), randomSize());
+            default:
+                return null;
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(Color.WHITE, DEFAULT_RADIUS);
```
okuzan, line 37:
```suggestion
        return new Circle(DEFAULT_COLOR, DEFAULT_RADIUS);
```

okuzan (changes requested):

okuzan (approved):
Well done!

PR #1264, opened 2023-09-10
Hv 4 solution

StVadym:


src/main/java/core/basesyntax/AreaCalculator.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax;
+
+public interface AreaCalculator {
+    double obtainArea();
+
+}
```
Ivan95kos, line 6: It is more common to use `get`
remove the last empty line in all classes/interfaces/methods where it is present
```suggestion
public interface AreaCalculator {
    double getArea();
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,43 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int FIGURE_COUNT = 5;
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private Figure randomFigure;
+    private final Figure defaultFigure = new Circle(10, Color.WHITE.name());
```
Ivan95kos, line 10: It is better to create a default Figure in the method
```suggestion
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,43 @@
...
+
+public class FigureSupplier {
+    private static final int FIGURE_COUNT = 5;
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private Figure randomFigure;
+    private final Figure defaultFigure = new Circle(10, Color.WHITE.name());
+
+    private int getRandomSide() {
+        return random.nextInt((100 - 1) + 1) + 1;
```
Ivan95kos, line 13: All magic numbers, strings, and symbols in your code should be constants.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int FIGURE_COUNT = 5;
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private Figure randomFigure;
+    private final Figure defaultFigure = new Circle(10, Color.WHITE.name());
+
+    private int getRandomSide() {
```
Ivan95kos, line 12: All private methods should be below public methods

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,43 @@
...
+                        getRandomSide(), colorSupplier.getRandomColor());
+                break;
+            case 4:
+                randomFigure = new Rectangle(getRandomSide(),
+                        getRandomSide(), colorSupplier.getRandomColor());
+                break;
+            default:
+                randomFigure = new IsoscelesTrapezoid(getRandomSide(), getRandomSide(),
+                        getRandomSide(), colorSupplier.getRandomColor());
+                break;
```
Ivan95kos, line 35: Move the creation of Figures in private methods
default should return the default value
```suggestion
            case 5:
                return prepareIsoscelesTrapezoid();
            default:
                return getDefaultFigure();
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,43 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int FIGURE_COUNT = 5;
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private Figure randomFigure;
```
Ivan95kos, line 9: useless variable
```suggestion
```

Ivan95kos (changes requested):
Good job overall!

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax;
+
+public abstract class Figure implements FigurePrinter, AreaCalculator {
+    private String color;
```
okuzan, line 4: would be more idiomatic to create a constructor for color, area, instead of using setters in constructors we would call super()

okuzan (approved):
Good job

src/main/java/core/basesyntax/Figure.java
StVadym: added constructor for color, but can't figure out how to make this with area, because I sould obtained it first and then add to parameters

PR #1268, opened 2023-09-13
Implemented all figure classes and interfaces

hehetenya:


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,49 @@
...
+            case CIRCLE:
+                figure = new Circle(colorSupplier.getRandomColor(),
+                        random.nextInt(MAX_SIDE_LENGTH));
+                break;
+            case RECTANGLE:
+                figure = new Rectangle(colorSupplier.getRandomColor(),
+                        random.nextInt(MAX_SIDE_LENGTH), random.nextInt(MAX_SIDE_LENGTH));
+                break;
+            case SQUARE:
+                new Square(colorSupplier.getRandomColor(), random.nextInt(MAX_SIDE_LENGTH));
```
kshuryhin, line 29: Looks like you forgot to assign `new Square(..) `to `figure`. Also, i think we can improve current solution:
I believe we do not need `figure` variable. In each case we can return new instance directly, and in the default clause - return `defaultFigure`.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,49 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int MAX_SIDE_LENGTH = 101;
+    private final Figure defaultFigure = new Circle(Color.WHITE, 10);
```
kshuryhin, line 7: It would be more readable, if we can make our default radius a constant.

kshuryhin (changes requested):
Great job! Just some minor comments

kshuryhin (approved):
Cool!

PR #1269, opened 2023-09-14
Solution for opp-advanced

druhov:
Methods and classes are created according to OOP, the code is checked for checkstyle

### DIFF
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private static final String[] COLORS = {"red", "green", "blue", "yellow", "white"};
+    private Random random;
+
+    public String getRandomColor() {
+        Random random = new Random();
+        int index = random.nextInt(COLORS.length);
+        return COLORS[index];
+    }
+}
--- a/src/main/java/core/basesyntax/Figure.java
+++ b/src/main/java/core/basesyntax/Figure.java
@@ -1,0 +1,17 @@
+package core.basesyntax;
+
+abstract class Figure implements Drawable, AreaCalculable {
+    private final String color;
+
+    public Figure(String color) {
+        this.color = color;
+    }
+
+    public abstract double getArea();
+
+    public abstract void draw();
+
+    public String getColor() {
+        return color;
+    }
+}
--- a/src/main/java/core/basesyntax/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/FigureSupplier.java
@@ -1,0 +1,32 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final Random RANDOM = new Random();
+
+    public Figure getRandomFigure() {
+        int randomType = RANDOM.nextInt(5);
+        String randomColor = new ColorSupplier().getRandomColor();
+        switch (randomType) {
+            case 0:
+                return new Square(RANDOM.nextDouble() * 10, randomColor);
+            case 1:
+                return new Rectangle(RANDOM.nextDouble() * 10,
+                        RANDOM.nextDouble() * 10, randomColor);
+            case 2:
+                return new RightTriangle(RANDOM.nextDouble() * 10 + 1,
+                        RANDOM.nextDouble() * 10 + 1, randomColor);
+            case 3:
+                return new Circle(RANDOM.nextDouble() * 10, randomColor);
+            default:;
+                return new IsoscelesTrapezoid(
+                        RANDOM.nextDouble() * 10, RANDOM.nextDouble() * 10,
+                        RANDOM.nextDouble() * 10, randomColor);
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(10, "white");
+    }
+}
--- a/src/main/java/core/basesyntax/Main.java
+++ b/src/main/java/core/basesyntax/Main.java
@@ -1,0 +1,20 @@
+package core.basesyntax;
+
+public class Main {
+    public static void main(String[] args) {
+        Figure[] figures = new Figure[6];
+        FigureSupplier figureSupplier = new FigureSupplier();
+
+        for (int i = 0; i < figures.length / 2; i++) {
+            figures[i] = figureSupplier.getRandomFigure();
+        }
+
+        for (int i = figures.length / 2; i < figures.length; i++) {
+            figures[i] = figureSupplier.getDefaultFigure();
+        }
+
+        for (Figure figure : figures) {
+            figure.draw();
+        }
+    }
+}
--- a/src/main/java/core/basesyntax/AreaCalculable.java
+++ b/src/main/java/core/basesyntax/AreaCalculable.java
@@ -1,0 +1,5 @@
+package core.basesyntax;
+
+public interface AreaCalculable {
+    double getArea();
+}
--- a/src/main/java/core/basesyntax/Circle.java
+++ b/src/main/java/core/basesyntax/Circle.java
@@ -1,0 +1,21 @@
+package core.basesyntax;
+
+class Circle extends Figure {
+    private final double radius;
+
+    public Circle(double radius, String color) {
+        super(color);
+        this.radius = radius;
+    }
+
+    @Override
+    public double getArea() {
+        return Math.PI * radius * radius;
+    }
+
+    @Override
+    public void draw() {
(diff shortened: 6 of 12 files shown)


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,32 @@
...
+            default:;
+                return new IsoscelesTrapezoid(
+                        RANDOM.nextDouble() * 10, RANDOM.nextDouble() * 10,
+                        RANDOM.nextDouble() * 10, randomColor);
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(10, "white");
+    }
```
olekskov, line 31: 1. Magic number should be constants with proper naming
2. Services our class rely on are usually private final as they are mutable and cannot be fully constants
3. Initialization of service with new in method will cause creation of new object each time method is called so should be moved to class level to avoid this
4. As you require colors to be available here and in ColorSupplier it's better to create enum with colors in order not to hardcode values in different places

```suggestion
public class FigureSupplier {
    private static final int SIDE_LIMIT = 10;
    private static final String DEFAULT_COLOR = Color.WHITE.name();
    private static final int DEFAULT_RADIUS = 10;
    
    private final Random random = new Random();
    private final ColorSupplier colorSupplier = new ColorSupplier();

    public Figure getRandomFigure() {
        int randomType = random.nextInt(Color.values.length);
        String randomColor = colorSupplier.getRandomColor();
        switch (randomType) {
            case 0:
                return new Square(getRandomSide(), randomColor);
            case 1:
                return new Rectangle(getRandomSide(),
                        getRandomSide(), randomColor);
            case 2:
                return new RightTriangle(getRandomSide(),
                        getRandomSide(), randomColor);
            case 3:
                return new Circle(getRandomSide(), randomColor);
            default:;
                return new IsoscelesTrapezoid(
                        getRandomSide(), getRandomSide(),
                        getRandomSide(), randomColor);
        }
    }

    public Figure getDefaultFigure() {
        return new Circle(DEFAULT_RADIUS, DEFAULT_COLOR);
    }
    
    private double getRandomSide() {
        return random.nextDouble() * SIDE_LIMIT;
    }
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,14 @@
...
+public class ColorSupplier {
+    private static final String[] COLORS = {"red", "green", "blue", "yellow", "white"};
+    private Random random;
+
+    public String getRandomColor() {
+        Random random = new Random();
+        int index = random.nextInt(COLORS.length);
+        return COLORS[index];
+    }
+}
```
olekskov, line 14: Please rework supplier by checking points from previous comment

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,20 @@
...
+            figures[i] = figureSupplier.getRandomFigure();
+        }
+
+        for (int i = figures.length / 2; i < figures.length; i++) {
+            figures[i] = figureSupplier.getDefaultFigure();
+        }
+
+        for (Figure figure : figures) {
+            figure.draw();
+        }
```
olekskov, line 18: 1. Please extract array length to constant with proper naming
2. Please rework solution using one cycle instead of 3

olekskov (changes requested):

src/main/java/core/basesyntax/ColorEnum.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum ColorEnum {
```
okuzan, line 3:
```suggestion
public enum Color {

```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,19 @@
...
+        Figure[] figures = new Figure[NUM_FIGURES];
+        FigureSupplier figureSupplier = new FigureSupplier();
+
+        for (int i = 0; i < NUM_FIGURES; i++) {
+            figures[i] = figureSupplier.getRandomFigure();
+        }
+
+        for (Figure figure : figures) {
+            System.out.println("Figure: " + figure.getClass().getSimpleName()
+                    + ", area: " + figure.getArea() + " sq.units, color: " + figure.getColor());
```
okuzan, line 16: replace figure with figures[i] and move to the previous loop

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,39 @@
...
+    public Figure getRandomFigure() {
+        int randomType = random.nextInt(NUM_FIGURE_TYPES);
+        String randomColor = new ColorSupplier().getRandomColor();
+        switch (randomType) {
+            case 0:
+                return new Square(getRandomSide(), randomColor);
+            case 1:
+                return new Rectangle(getRandomSide(), getRandomSide(), randomColor);
+            case 2:
+                return new RightTriangle(getRandomSide() + 1, getRandomSide() + 1, randomColor);
```
okuzan, line 23: why do you add + 1 only here?

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,17 @@
...
+abstract class Figure implements Drawable, AreaCalculable {
+    private final String color;
+
+    public Figure(String color) {
+        this.color = color;
+    }
+
+    public abstract double getArea();
+
+    public abstract void draw();
```
okuzan, line 12:
```suggestion

```

okuzan (changes requested):
GJ

src/main/java/core/basesyntax/FigureSupplier.java
druhov: Since it's a triangle, and the condition doesn't specify whether it's right or not, it seemed to me that this was the right solution.

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum Colors {
```
Ivan95kos, line 3: naming convention: https://mate-academy.github.io/style-guides/java/java.html#s4.8.1-enum-classes
```suggestion
public enum Color {

```

src/main/java/core/basesyntax/Figure.java
Ivan95kos: not fixed
You are duplicating what is in the interfaces, what is the use of this, what have you done?

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,39 @@
...
+    private static final double SIDE_LIMIT = 10.0;
+    private static final int DEFAULT_RADIUS = 10;
+    private static final String DEFAULT_COLOR = "white";
+
+    private final Random random = new Random();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        int randomType = random.nextInt(NUM_FIGURE_TYPES);
+        String randomColor = new ColorSupplier().getRandomColor();
```
Ivan95kos, line 16:
```suggestion
        String randomColor = colorSupplier.getRandomColor();

```

Ivan95kos (changes requested):

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,17 @@
...
+    public static void main(String[] args) {
+        Figure[] figures = new Figure[NUM_FIGURES];
+        FigureSupplier figureSupplier = new FigureSupplier();
+
+        for (int i = 0; i < NUM_FIGURES; i++) {
+            figures[i] = figureSupplier.getRandomFigure();
+            System.out.println("Figure: " + figures[i].getClass().getSimpleName()
+                    + ", area: " + figures[i].getArea()
+                    + " sq.units, color: " + figures[i].getColor());
+        }
```
olekskov, line 15:
```suggestion
        for (int i = 0; i < NUM_FIGURES; i++) {
            if (i < NUM_FIGURES / 2) {
                figures[i] = figureSupplier.getRandomFigure();
            } else {
                figures[i] = figureSupplier.getDefaultFigure();
            }
            figures[i].draw;
        }

```

olekskov (approved):
Good job overall, approved.
In main method you didn't fully comply with task (make array consist of both random and default figures and using draw() method)

src/main/java/core/basesyntax/Main.java
druhov: Oh, right, I'm a little confused now, I replaced the solution with your hint, thanks!)

PR #1270, opened 2023-09-17
OOP - advanced 

rafik2211:
Defines an enumeration for different colors.
ColorSupplier Class (ColorSupplier):
Generates a random color using the Color enumeration. Figure Interface (Figure):
Declares methods getArea() and draw() which concrete figure classes must implement. FigureSupplier Class (FigureSupplier):
Generates random figures based on the random choice and color from ColorSupplier. Provides a default figure (a white circle with a radius of 10) when needed. Shape Abstract Class (Shape):
Serves as a common base class for various geometric figure implementations. Contains a color field and constructor.
Concrete Figure Classes (Circle, IsoscelesTrapezoid, Rectangle, RightTriangle, Square): Implement the Figure interface and extend the Shape class. Define specific properties and methods for each type of geometric figure. Calculate the area and provide a draw method to display figure information. The main method creates random instances of geometric figures and prints their details. The color for each figure is chosen randomly from the Color enumeration. The draw method for each figure prints out information about the figure, including its color, dimensions, and area.


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,45 @@
...
+                        random.nextDouble() * 10 + 1,
+                        random.nextDouble() * 10 + 1,
+                        random.nextDouble() * 10 + 1);
+            default:
+                return null;
+        }
+    }
+
+    public AreaCalculatable getDefaultFigure() {
+        return new Circle("WHITE", 10);
```
okuzan, line 43: Color.WHITE and 10 should be DEFAULT_COLOR, DEFAULT_RADIUS constants

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,45 @@
...
+        this.random = random;
+    }
+
+    public Figure getFigureRandom() {
+        int choice = random.nextInt(5);
+
+        switch (choice) {
+            case 0:
+                return new Square(colorSupplier.getRandomColor().toString(),
+                        random.nextInt(10) + 1);
```
okuzan, line 20: why some figures use integers and some use doubles? Also, extract a repeating piece of logic to a private method. Better make 10 a constant. You can also create a private utility method for creation of each type of figure, if the resulting code will look too bulky

okuzan (changes requested):
Good job on the first try!

olekskov (approved):

PR #1271, opened 2023-09-18
Final version of jv-oop-advanced challenge

MateuszRostek:


src/main/java/core/basesyntax/AreaCalculator.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax;
+
+public interface AreaCalculator {
+
```
nowickimj, line 4: Redundant empty line

nowickimj, line 4: Redundant empty line

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,21 @@
...
+    }
+
+    @Override
+    public double calculateArea() {
+        return Math.PI * radius * radius;
+    }
+
+    @Override
+    public void draw() {
+        System.out.printf("Figure: Circle | radius: %.2f | area: %.2f |"
```
nowickimj, line 18: According to requirements, you are supposed to use `println` method.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private static final Random random = new Random();
+
+    public String getRandomColor() {
+        Colors[] colors = Colors.values();
+
```
nowickimj, line 10: Redundant new line.

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax;
+
+public enum Colors {
+
```
nowickimj, line 4: I see that in some places you add extra new line at the beginning of block whereas in other you don't. 
Please see that according to `Common mistakes` doc, you shouldn't add any new lines here: https://mate-academy.github.io/jv-program-common-mistakes/java-core/abstract-class-interface/oop-advanced

src/main/java/core/basesyntax/Drawable.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax;
+
+public interface Drawable {
+
```
nowickimj, line 4: Same goes here.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final Random random = new Random();
+    private static final int FIGURES_COUNT = 5;
+    private static final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        final int randomSizeRange = random.nextInt(10) + 1;
```
nowickimj, line 11: If we consider `10` as a max value in range, it could also be stored as constant.
Also I'm not sure if 'range' in variable name is correct - from what I see you picked a random fixed value here so it's not range anymore.

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,25 @@
+package core.basesyntax;
+
+public class Main {
+
+    public static void main(String[] args) {
+
```
nowickimj, line 6: Lots of empty lines here - you don't need to put them between every two lines of code :)

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,33 @@
...
+
+        switch (random.nextInt(FIGURES_COUNT)) {
```
nowickimj, line 13: I'm not a big fan of this approach - although the solution is correct, I think it would be more appropriate to pick a random figure type instead of a number/identifier representing that figure.
You can define an enum that stores all figure types and pick a type in the same manner as you do with colors.
Then, your code will be more understandable:
```
case TRIANGE: ...
case SQUARE: ...
...
```

nowickimj (changes requested):
Your code looks very good, most of my remarks are minor.
What you could additionally do is to move all figure classes into separate `figures` package.

MateuszRostek:
I wanted to use _**`import core.basesyntax.figures.*;`**_ in  **FigureSupplier** but maven clean package had some problems with it, so I had to add all figures one by one

MateuszRostek:
Also, I have a question about using `println` instead of `printf` in this task:
I changed it, but in my opinion the code becomes much less readable. If I find a solution that seems better to me, and we get the same result, should I continue to rigidly follow the instructions? Or am I missing something in my first approach?

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public enum Colors {
```
Rommelua, line 3:
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,39 @@
...
+import core.basesyntax.figures.RightTriangle;
+import core.basesyntax.figures.Square;
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final Random random = new Random();
+    private static final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        final int maxValue = 10;
```
Rommelua, line 15: it should be constant - `private static final `field

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,39 @@
...
+            case RIGHT_TRIANGLE:
+                return new RightTriangle(colorSupplier.getRandomColor(),
+                        randomSize, randomSize);
+            default:
+                return new Square(colorSupplier.getRandomColor(), randomSize);
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(Colors.WHITE.name(), 10);
```
Rommelua, line 37: image

src/main/java/core/basesyntax/Figures.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public enum Figures {
```
Rommelua, line 3:
```suggestion
public enum Figure {
```

Rommelua (changes requested):

Rommelua:
> Also, I have a question about using `println` instead of `printf` in this task: I changed it, but in my opinion the code becomes much less readable. If I find a solution that seems better to me, and we get the same result, should I continue to rigidly follow the instructions? Or am I missing something in my first approach?

Both using `println` or `printf` is acceptable in my opinion)

MateuszRostek:
I've named enum `FigureType`, because I already have `Figure` class

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,41 @@
...
+    private static final ColorSupplier colorSupplier = new ColorSupplier();
+    private static final int MAX_VALUE = 10;
+    private static final int DEFAULT_FIGURE_VALUE = 10;
+    private static final String DEFAULT_COLOR = Color.WHITE.name();
+
+    public Figure getRandomFigure() {
+        final int randomSize = random.nextInt(MAX_VALUE) + 1;
+        FigureType[] figures = FigureType.values();
+
+        switch (figures[random.nextInt(figures.length)]) {
```
nowickimj, line 21: Minor thing here: note that the `figures` variable (btw, it should be named `figureTypes`) is not needed elsewhere and the only thing that interests you in this part of code is a random figure type. 
That means you can improve the code readibility by extracting this code into a separate method. 

Think of a appropriate name for it, for example: `getRandomFigureType()`.

nowickimj (changes requested):
One small improvement to be made here - everything else looks great :)

Rommelua (approved):
Good job!

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,45 @@
...
+import core.basesyntax.figures.RightTriangle;
+import core.basesyntax.figures.Square;
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final Random random = new Random();
+    private static final ColorSupplier colorSupplier = new ColorSupplier();
+    private static final int MAX_VALUE = 10;
+    private static final int DEFAULT_FIGURE_VALUE = 10;
+    private static final String DEFAULT_COLOR = Color.WHITE.name();
```
nowickimj, line 15: a minor thing: please keep in mind that names of constants and variables should be as self-explanatory as possible.
It is very typical for programmers to write something and then have no clue what's going in the code when they return to it some time later and have to fix something.
I think it would be better to name those constants
"DEFAULT_SIZE" and "MAX_SIZE".

src/main/java/core/basesyntax/figures/Circle.java
```diff
@@ -0,0 +1,24 @@
...
+    @Override
+    public double calculateArea() {
+        return Math.PI * radius * radius;
+    }
+
+    @Override
+    public void draw() {
+        System.out.println("Figure: Circle | radius: " + String.format("%.2f", radius)
+                + " | area: " + String.format("%.2f", calculateArea())
+                + " | color: " + getColor());
```
nowickimj, line 22: Just a small remark here: when using `String.format()` method, you can use a pattern with placeholders and pass real values as args - that way you write it in a single line or even a string block (since Java 12), which makes it easier to read and modify:

single line:
```
String.format("Figure: Circle, radius: %.2f, area: %.2f",  color: %s", radius, calculateArea(), getColor());
```

nowickimj (approved):
Two advices for future development, besides that looks great - keep it going! :)

PR #1272, opened 2023-09-19
First solution of jv-oop-advanced

QbaSekowski:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    public String getRandomColor() {
+        Random random = new Random();
```
Rommelua, line 7: image

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public enum Colors {
```
Rommelua, line 3:
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,15 @@
...
+
+    public Figure(String color) {
+        this.color = color;
+    }
+
+    public String getColor() {
+        return color.toLowerCase();
+    }
+
+    public abstract void draw();
```
Rommelua, line 14: image

Rommelua (changes requested):

Before you create PR with your homework, you need to go through the checklist under the task and correct all the points described there. The mentor will not check the work until the checklist points are corrected

QbaSekowski:
Fixes made in 3 classes:
- ColorSupplier: corrected variable to class-level
- Drawable: created interface for drawing
- Figure: now implements Drawable interface, draw method deleted

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,43 @@
...
+    public static final int UPPER_BOUND = 20;
+    // the minimum length of each segment in the figure
+    public static final int LOWER_BOUND = 1;
+
+    public Figure getRandomFigure() {
+        Random random = new Random();
+        int randomIndex = random.nextInt(FIGURE_COUNT);
+        String randomColor = new ColorSupplier().getRandomColor();
+        switch (randomIndex) {
+            case 0:
```
nowickimj, line 18: In one of PRs of other student, I proposed a bit different approach - instead of using indexes, you can introduce an enum (named for example `FigureType`) that will hold names of each figure type and use it with `switch` instruction. 
So, the result would be:
```java
FigureType figureType = getRandomFigureType();
switch(figureType) {
    case TRIANGLE:
    // handler for triangle   
    case SQUARE:
    // handler for square etc. etc.
```

That seems to be more natural approach comparing to using artificial identifier for each shape.

nowickimj (approved):
More of an advice rather than actual change request but, in my opinion, worth mentioning :)

src/main/java/core/basesyntax/FigureSupplier.java
QbaSekowski: Yes, indeed that would make the code easier to understand. I'll try to remember it.
Thank you.

PR #1274, opened 2023-09-20
Solution

Kalixone:
Interface "GetArea" was neccesary in this case?, after i wrote command "mvn clean package" it told me to create interface, but i dont know why, do we need it here?

### DIFF
--- a/src/main/java/core/basesyntax/Circle.java
+++ b/src/main/java/core/basesyntax/Circle.java
@@ -1,0 +1,29 @@
+package core.basesyntax;
+
+import java.text.DecimalFormat;
+
+public class Circle extends Figure {
+    private DecimalFormat decimalFormat = new DecimalFormat("#.##");
+    private double radius;
+
+    public Circle(double radius, String color) {
+        super.setColor(color);
+        this.radius = radius;
+    }
+
+    public double getRadius() {
+        return radius;
+    }
+
+    public static Circle createCircle(double radius, String color) {
+        return new Circle(radius, color);
+    }
+
+    @Override
+    public void getArea() {
+        double area = (Math.PI * radius * radius);
+        System.out.println("Figure: circle, area: " + decimalFormat.format(area)
+                + " sq.units, radius: " + decimalFormat.format(radius)
+                + " units, color: " + getColor());
+    }
+}
--- a/src/main/java/core/basesyntax/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/FigureSupplier.java
@@ -1,0 +1,54 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        FigureType figureType = FigureType.values()[random.nextInt(FigureType.values().length)];
+        Figure figure = null;
+        String color = colorSupplier.getRandomColor();
+
+        switch (figureType) {
+            case SQUARE:
+                double squareSideLength = random.nextDouble() * 5;
+                figure = Square.createSquare(squareSideLength, color);
+                break;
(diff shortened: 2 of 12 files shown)


src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,29 @@
+package core.basesyntax;
+
+import java.text.DecimalFormat;
+
+public class Circle extends Figure {
+    private DecimalFormat decimalFormat = new DecimalFormat("#.##");
```
nowickimj, line 6: If you're not providing any way to change the format, this field can be considered `final`.
Also, if you want to provide the same decimal format for every shape, you can move it to `Figure.class` and make it protected, so every subclass may access it.

nowickimj (changes requested):

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,54 @@
...
+    public Figure getRandomFigure() {
+        FigureType figureType = FigureType.values()[random.nextInt(FigureType.values().length)];
+        Figure figure = null;
+        String color = colorSupplier.getRandomColor();
+
+        switch (figureType) {
+            case SQUARE:
+                double squareSideLength = random.nextDouble() * 5;
+                figure = Square.createSquare(squareSideLength, color);
+                break;
```
nowickimj, line 18: I'm not a big fan of dangling null variables - instead of initializing nullable variable, you can return the figure immediately within the `case` block - that would also remove the need for `break` instruction.

src/main/java/core/basesyntax/GetArea.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface GetArea {
```
nowickimj, line 3: The name for this class is wrong:

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,54 @@
...
+                figure = getDefaultFigure();
+                break;
+
+        }
+        return figure;
+    }
+
+    public Figure getDefaultFigure() {
+        double defaultRadius = 10.0;
+        String defaultColor = "white";
```
nowickimj, line 50: If you consider this values as default, you should make them constants instead of local variables

nowickimj (changes requested):

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,49 @@
...
+                        isoscelesTrapezoidLowerBase, isoscelesTrapezoidHeight, color);
+            default:
+                return getDefaultFigure();
+
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        double defaultRadius = 10.0;
+        String defaultColor = "white";
```
nowickimj, line 45: You've created a constants but forgot to remove local variables :)

src/main/java/core/basesyntax/AreaCalculator.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface AreaCalculator {
```
nowickimj, line 3: I see few issues with this interface:
- The name of your interface is "AreaCalculator" whereas its method is named "getArea()" which is a void type. In Java naming convention, by 'get' prefix you imply that your method is supposed to return something and which you're not doing.
- Additionally, it doesn't actually provide information just on area but also other properties of the figure.

Please see the task requirement:
"behaviour - we can obtain the area of any figure and are able to draw it. To 'draw' means to print out all information about a figure using System.out.println() (you shouldn't override toString() method for this)."

How about 'Drawable' interface with 'draw()' method?

nowickimj (changes requested):

nowickimj (approved):

PR #1277, opened 2023-09-24
jv-oop-advanced solution

BolsunovaNataliia:


src/main/java/core/basesyntax/figures/Figure.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax.figures;
+
+public abstract class Figure implements IPrintable, IFigure {
+    private String color;
```
kshuryhin, line 4: As far as we already have `Color` enum, representing colors, we can change our `color` variable type to `Color`

src/main/java/core/basesyntax/figures/IFigure.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax.figures;
+
+public interface IFigure {
```
kshuryhin, line 3: I would recommend to be more precise with interface name here. For example, `AreaCalculator` name will be more descriptive. Also, when renaming interface think about method `getArea()`. I would say, it is more suitable here to use `calculateArea()`, as indeed we have some computation, rather than fetching predefined data.

src/main/java/core/basesyntax/color/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax.color;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    public static String getRandomColor() {
+        int index = new Random().nextInt(Color.values().length);
```
kshuryhin, line 7: I would recommend here to make `Random` instance a class member. In such way we will avoid creating instance each time method is invoked.

src/main/java/core/basesyntax/color/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax.color;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    public static String getRandomColor() {
+        int index = new Random().nextInt(Color.values().length);
+        Color color = Color.values()[index];
+
+        return color.toString().toLowerCase();
```
kshuryhin, line 10: Please, do not use `toString()` to obtain a String representation of enum. Instead, use `name()` method.

[Code recommendations](https://mate-academy.github.io/jv-program-common-mistakes/java-core/abstract-class-interface/oop-advanced#use-name-for-getting-string-representation-of-enum-constants)

src/main/java/core/basesyntax/figuresuppliers/FigureSupplier.java
```diff
@@ -0,0 +1,40 @@
...
+import core.basesyntax.figures.Square;
+import java.util.Random;
+
+public class FigureSupplier {
+    public Figure getRandomFigure() {
+        int sizeOne = new Random().nextInt(10) + 1;
+        int sizeTwo = new Random().nextInt(10) + 1;
+        int sizeThree = new Random().nextInt(10) + 1;
+
+        int index = new Random().nextInt(FigureType.values().length);
```
kshuryhin, line 18: I would recommend to make `Random` instance a class member. It brings some benefits, like we do not need to create instance each time method is invoked.

src/main/java/core/basesyntax/figuresuppliers/FigureSupplier.java
```diff
@@ -0,0 +1,40 @@
...
+                return new Circle(ColorSupplier.getRandomColor(), sizeOne);
+            case ISOSCELES_TRAPEZOID:
+                return new IsoscelesTrapezoid(ColorSupplier.getRandomColor(),
+                        sizeOne, sizeTwo, sizeThree);
+            default: return null;
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle("white", 10);
```
kshuryhin, line 38: Please, try to avoid using magic values. Instead, make them constants.

[Code recommendations ](https://mate-academy.github.io/jv-program-common-mistakes/java-core/abstract-class-interface/oop-advanced#all-magic-numbers-in-your-code-should-be-constants)

src/main/java/core/basesyntax/figuresuppliers/FigureSupplier.java
```diff
@@ -0,0 +1,40 @@
...
+            case RECTANGLE:
+                return new Rectangle(ColorSupplier.getRandomColor(), sizeOne, sizeTwo);
+            case RIGHT_TRIANGLE:
+                return new RightTriangle(ColorSupplier.getRandomColor(), sizeOne, sizeTwo);
+            case CIRCLE:
+                return new Circle(ColorSupplier.getRandomColor(), sizeOne);
+            case ISOSCELES_TRAPEZOID:
+                return new IsoscelesTrapezoid(ColorSupplier.getRandomColor(),
+                        sizeOne, sizeTwo, sizeThree);
+            default: return null;
```
kshuryhin, line 33: Please, try to avoid returning `null` from the method. 

[Code recommendations](https://mate-academy.github.io/jv-program-common-mistakes/java-core/abstract-class-interface/oop-advanced#dont-return-null-from-a-method)

src/main/java/core/basesyntax/figuresuppliers/FigureSupplier.java
```diff
@@ -0,0 +1,40 @@
...
+import core.basesyntax.figures.Rectangle;
+import core.basesyntax.figures.RightTriangle;
+import core.basesyntax.figures.Square;
+import java.util.Random;
+
+public class FigureSupplier {
+    public Figure getRandomFigure() {
+        int sizeOne = new Random().nextInt(10) + 1;
+        int sizeTwo = new Random().nextInt(10) + 1;
+        int sizeThree = new Random().nextInt(10) + 1;
```
kshuryhin, line 16: I believe we can improve this and create each side  directly in `case` section. By doing so, we will avoid creation of not needed sides. Also, please do not create new `Random` instance each time you need to create a side value. Instead, as I mentioned, make `Random` a class member and reuse it.

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,24 @@
+package core.basesyntax;
+
+import core.basesyntax.figures.Figure;
+import core.basesyntax.figuresuppliers.FigureSupplier;
+
+public class Main {
+    public static void main(String[] args) {
+        int size = 4;
```
kshuryhin, line 8: I would recommend to make `size` variable a constant.

kshuryhin (changes requested):
Good job! Just some recommendations, mostly related to level of variables

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,25 @@
+package core.basesyntax;
+
+import core.basesyntax.figures.Figure;
+import core.basesyntax.figuresuppliers.FigureSupplier;
+
+public class Main {
+    private static final int SIZE = 4;
+
+    public static void main(String[] args) {
+        Figure[] arr = new Figure[SIZE];
```
mnyshenko, line 10: Please try to choose more descriptive variable names. arr is too generic, code is not always read from top to bottom)) There might be a case when some developer opens your code on the 13th line and wants to understand what you do there, but in this case, he would need to go all the way up to figure out what is stored in arr variable. A better naming choice would be "figures" or something similar

src/main/java/core/basesyntax/figuresuppliers/FigureSupplier.java
```diff
@@ -0,0 +1,51 @@
...
+import core.basesyntax.figures.Rectangle;
+import core.basesyntax.figures.RightTriangle;
+import core.basesyntax.figures.Square;
+import java.util.Random;
+
+public class FigureSupplier {
+    static final Color DEFAULT_COLOR = Color.WHITE;
+    static final int MAX_EDGE = 10;
+    static final int ANNEX = 1;
+    static final int DEFAULT_RADIUS = 10;
```
mnyshenko, line 17: Please pay attention to the access modifiers. By setting no modifiers you implicitly set package-private modifier, do we really need to open these variables for other classes in this package?

src/main/java/core/basesyntax/figuresuppliers/FigureSupplier.java
```diff
@@ -0,0 +1,51 @@
...
+                return new RightTriangle(randomColor, sizeOne, sizeTwo);
+            case CIRCLE:
+                return new Circle(randomColor, sizeOne);
+            case ISOSCELES_TRAPEZOID:
+                sizeTwo = random.nextInt(MAX_EDGE) + ANNEX;
+                int sizeThree = random.nextInt(MAX_EDGE) + ANNEX;
+                return new IsoscelesTrapezoid(randomColor,
+                        sizeOne, sizeTwo, sizeThree);
+            default: return getDefaultFigure();
+        }
```
mnyshenko, line 45: This part has a problem with code duplication. You generate random sizes using one formula. Let's imagine tomorrow you'll need to change this formula, then you'll have to go over each line and make a change. It's insufficient. In this case, you can create a separate method for this computation and reuse it.

src/main/java/core/basesyntax/figures/IPrintable.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax.figures;
+
+public interface IPrintable {
```
mnyshenko, line 3: Debatable topic but I would not recommend you put the prefix "I" before interface names, this is not common practice at least for Java. Here's an interesting discussion on it if you want to dig deeper into the pros and cons of this choice https://stackoverflow.com/questions/541912/interface-naming-in-java

mnyshenko (changes requested):

src/main/java/core/basesyntax/color/ColorSupplier.java
```diff
@@ -0,0 +1,14 @@
...
+import java.util.Random;
+
+public class ColorSupplier {
+    private Random random = new Random();
+
+    public Color getRandomColor() {
+        int index = random.nextInt(Color.values().length);
+        Color color = Color.values()[index];
+
+        return color;
```
d1sam, line 12: Please don't add redundant empty line. 
Don’t create temporary variables if you can directly return value from the method
Bad example:
```
public User createUser(String name, int age) {
  User user = new User(name, age);
  return user;
}
```
Improved example:
```
public User createUser(String name, int age) {
  return new User(name, age);
}
```

src/main/java/core/basesyntax/figures/AreaCalculator.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax.figures;
+
+public interface AreaCalculator {
+
```
d1sam, line 4: Please don't add redundant empty line.

d1sam (changes requested):
Good job, but some minor changes should be done.

d1sam (approved):
Well done!

PR #1279, opened 2023-09-25
practiced in OOP with random figures

HovorukhaBohdan:


src/main/java/core/basesyntax/randomizer/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax.randomizer;
+
+import core.basesyntax.model.Colors;
+import java.util.Random;
+
+public class ColorSupplier {
+    private int index = new Random().nextInt(Colors.values().length - 1);
```
kshuryhin, line 7: I would recommend to make `index` a local variable in this case, as it is not constant nor class dependency. Also, fell free to do 
`.nextInt(Color.values().length);`
Because `nextInt(int bound)` method generates values between 0 (inclusive) and the specified value (exclusive). 
[Java Doc](https://docs.oracle.com/javase/8/docs/api/java/util/Random.html#nextInt-int-)

src/main/java/core/basesyntax/randomizer/FigureSupplier.java
```diff
@@ -0,0 +1,41 @@
...
+import core.basesyntax.model.Colors;
+import core.basesyntax.model.Figure;
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int MAX_VALUE = 50;
+
+    public Figure getRandomFigure() {
+        Random random = new Random();
+        ColorSupplier colorSupplier = new ColorSupplier();
```
kshuryhin, line 17: And here I would recommend to make `random` and `colorSupplier` class members. It can be beneficial for several reasons, especially we don not need to create both instances each time method is called.

src/main/java/core/basesyntax/randomizer/FigureSupplier.java
```diff
@@ -0,0 +1,41 @@
...
+        IsoscelesTrapezoid isoscelesTrapezoid = new IsoscelesTrapezoid(random.nextInt(MAX_VALUE),
+                                            random.nextInt(MAX_VALUE), random.nextInt(MAX_VALUE));
+        Rectangle rectangle = new Rectangle(random.nextInt(MAX_VALUE), random.nextInt(MAX_VALUE));
+        RightTriangle rightTriangle = new RightTriangle(random.nextInt(MAX_VALUE),
+                                                        random.nextInt(MAX_VALUE));
+        Square square = new Square(random.nextInt(MAX_VALUE));
+
+        Figure[] figures = {circle, isoscelesTrapezoid, rectangle, rightTriangle, square};
+
+        Figure figure = figures[new Random().nextInt(figures.length - 1)];
```
kshuryhin, line 29: My suggestion here is to use `random.nextInt(figures.length)` for selecting a random figure. By using `figures.length - 1`, you're excluding the last figure (`square` in this case) from the possible selections.
Also, you should not create here another `new Random()` instance, as you already have predefined one.

src/main/java/core/basesyntax/randomizer/FigureSupplier.java
```diff
@@ -0,0 +1,41 @@
...
+
+        Circle circle = new Circle(random.nextInt(MAX_VALUE));
+        IsoscelesTrapezoid isoscelesTrapezoid = new IsoscelesTrapezoid(random.nextInt(MAX_VALUE),
+                                            random.nextInt(MAX_VALUE), random.nextInt(MAX_VALUE));
+        Rectangle rectangle = new Rectangle(random.nextInt(MAX_VALUE), random.nextInt(MAX_VALUE));
+        RightTriangle rightTriangle = new RightTriangle(random.nextInt(MAX_VALUE),
+                                                        random.nextInt(MAX_VALUE));
+        Square square = new Square(random.nextInt(MAX_VALUE));
+
+        Figure[] figures = {circle, isoscelesTrapezoid, rectangle, rightTriangle, square};
```
kshuryhin, line 27: That's a valid approach, however I see a drawback here - we are creating too many objects, when we need just one. I would recommend to consider another approach, probably using `switch` construction.
Hint: you can create another enum, which will contain all figure names in it.

src/main/java/core/basesyntax/randomizer/FigureSupplier.java
```diff
@@ -0,0 +1,41 @@
...
+
+        Figure figure = figures[new Random().nextInt(figures.length - 1)];
+        figure.setColor(colorSupplier.getRandomColor());
+
+        return figure;
+    }
+
+    public Figure getDefaultFigure() {
+        Circle circle = new Circle(10);
```
kshuryhin, line 36: Please, try to avoid magic numbers in your code.

[Code recommendations](https://mate-academy.github.io/jv-program-common-mistakes/java-core/abstract-class-interface/oop-advanced#all-magic-numbers-in-your-code-should-be-constants)

src/main/java/core/basesyntax/randomizer/ColorSupplier.java
kshuryhin, line 7: Also, it would be also beneficial to move `Random` instance to class level.

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax;
+
+import core.basesyntax.model.Figure;
+import core.basesyntax.randomizer.FigureSupplier;
+
+public class Main {
+    public static void main(String[] args) {
+        int lengthOfArray = 6;
```
kshuryhin, line 8: I would recommend to make `lengthOfArray` a constant.

[Code recommendations](https://mate-academy.github.io/jv-program-common-mistakes/java-core/abstract-class-interface/oop-advanced#all-magic-numbers-in-your-code-should-be-constants)

src/main/java/core/basesyntax/model/Figure.java
```diff
@@ -0,0 +1,23 @@
+package core.basesyntax.model;
+
+public class Figure implements AreaCalculator, Drawable {
+    private String color;
```
kshuryhin, line 4: As far as you already have `Colors` enum, feel free to use it as a type of color variable. Also, I would say it is more preferable to use singular rather than plural: `Color`

kshuryhin (changes requested):
Nice try! But there are some points to consider, mostly related to variable levels

src/main/java/core/basesyntax/randomizer/FigureSupplier.java
HovorukhaBohdan: No problem! But when I use "mvn clean package", maven says that I need to do them private. Idk why(

src/main/java/core/basesyntax/randomizer/FigureSupplier.java
HovorukhaBohdan: Is it a good practice to make a static method to get a random value, instead of writing "random.nextInt(MAX_VALUE)" every time?

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax;
+
+import core.basesyntax.model.Figure;
+import core.basesyntax.randomizer.FigureSupplier;
+
+public class Main {
+    public static void main(String[] args) {
+        final int lengthOfArray = 6;
```
kshuryhin, line 8: Please, pay more attention to what constant variable means in context of Java. 
Here is [hint](https://mate-academy.github.io/jv-program-common-mistakes/java-core/abstract-class-interface/oop-advanced#all-magic-numbers-in-your-code-should-be-constants).

src/main/java/core/basesyntax/randomizer/FigureSupplier.java
kshuryhin: @HovorukhaBohdan generally using static methods is not best practice and for our case it will not bring any benefits. We have one `random` instance and each time calling `nextInt` method for random value, that's totally fine :)

src/main/java/core/basesyntax/randomizer/FigureSupplier.java
kshuryhin: @HovorukhaBohdan feel free to make them private! That is even more than best practice, because generally it improves code encapsulation and safety.

src/main/java/core/basesyntax/figures/Square.java
```diff
@@ -0,0 +1,28 @@
...
+
+import core.basesyntax.model.Color;
+import core.basesyntax.model.Figure;
+
+public class Square extends Figure {
+    private double side;
+
+    public Square(double side, Color color) {
+        this.side = side;
+        super.setColor(color);
```
kshuryhin, line 11: I would recommend to create constructor in `Figure` class
`public Figure(color)`

and use it, when creating concrete figures. It will help us avoid having partial-stated objects and sounds more logical: when we create figure, it cannot be without color

kshuryhin (changes requested):
Great job, some minor points to consider

src/main/java/core/basesyntax/Main.java
HovorukhaBohdan: I have a question, when I write "final int ARRAY_LENGTH" maven is failing build, why? Hint says that I need to name constants like "public static final int FIGURE_COUNT = 5;".

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax;
+
+import core.basesyntax.model.Figure;
+import core.basesyntax.randomizer.FigureSupplier;
+
+public class Main {
+    public static void main(String[] args) {
+        final int Array_Length = 6;
```
d1sam, line 8: All magic numbers in your code should be constants.
Please see [this article](https://mate-academy.github.io/style-guides/java/java.html#s5.2.4-constant-names) to learn about constant fields and their naming requirements.

src/main/java/core/basesyntax/model/Figure.java
```diff
@@ -0,0 +1,23 @@
...
+
+    @Override
+    public double getArea() {
+        return 0;
+    }
+
+    @Override
+    public void draw() {
+        System.out.println("Drawing...");
+    }
```
d1sam, line 22: You should make these methods `abstract`.

src/main/java/core/basesyntax/model/Figure.java
```diff
@@ -0,0 +1,23 @@
+package core.basesyntax.model;
+
+public class Figure implements AreaCalculator, Drawable {
```
d1sam, line 3: When you made some methods `abstract` your class should be defined also as a `abstract`

d1sam (changes requested):
Good job, but some minor changes should be done.

src/main/java/core/basesyntax/model/AreaCalculator.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax.model;
+
+public interface AreaCalculator {
+    double getArea();
```
kshuryhin, line 4: I would say it is better here to use `calculateArea` method name as we actually do some computation rather than get stored value.

kshuryhin (approved):
Nice job!

PR #1280, opened 2023-09-27
Realization of full application

forzen123212:


src/main/java/core/basesyntax/Application.java
```diff
@@ -0,0 +1,18 @@
+package core.basesyntax;
+
+public class Application {
+    private static final int FIGURE_COUNT = 6;
+
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        GeometricFigure[] figures = new GeometricFigure[FIGURE_COUNT];
+        for (int i = 0; i < figures.length; i++) {
```
mnyshenko, line 9: Minor: Since you already have FIGURE_COUNT to represent the length of an array, I suggest using that variable in for expression for better readability

src/main/java/core/basesyntax/Color.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public enum Color {
+    BLACK, WHITE, YELLOW, GREEN, BLUE, ORANGE, PURPLE, PINK
+}
```
mnyshenko, line 5: It's better practice to put each value on a new line

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private final Random random = new Random();
+
+    public String getRandomColor() {
+        int index = random.nextInt(Color.values().length);
+        Color color = Color.values()[index];
```
mnyshenko, line 10: I suggest extracting Color.values() on a class level for further reuse of that variable

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,37 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final ColorSupplier COLOR_SUPPLIER = new ColorSupplier();
```
mnyshenko, line 6: Each FigureSupplier should have its own ColorSupplier, it should not be a constant

src/main/java/core/basesyntax/GeometricFigure.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+public abstract class GeometricFigure implements Drawable, AreaCalculator {
+    protected String color;
+
+    public GeometricFigure(String colour) {
+        this.color = colour;
+    }
```
mnyshenko, line 8: minor: please rename colour -> color

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,37 @@
...
+    private static final int MAX_LENGTH = 100;
+    private static final int FIGURE_COUNT = 5;
+    private static final int DEFAULT_RADIUS = 10;
+    private static final String DEFAULT_COLOR = String.valueOf(Color.GREEN);
```
mnyshenko, line 10: Minor: as the task says the default color should be white

src/main/java/core/basesyntax/GeometricFigure.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+public abstract class GeometricFigure implements Drawable, AreaCalculator {
```
mnyshenko, line 3: Please pay attention to the task requirements which ask you to create a Figure class, not GeometricFigure

src/test/java/core/test/StructureTest.java
mnyshenko, line 1: I guess you're not supposed to change this file

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,37 @@
...
+            case 1:
+                return new IsoscelesTrapezoid(randomProperty, randomProperty,
+                        randomProperty, randomColor);
+            case 2:
+                return new Rectangle(randomProperty, randomProperty, randomColor);
+            case 3:
+                return new RightTriangle(randomProperty, randomProperty, randomColor);
+            default:
+                return new Square(randomProperty, randomColor);
+        }
```
mnyshenko, line 31: It seems not quite the right implementation. For example, you will **always** pass 2 equal numbers to the Rectangle constructor, so you'll always get a Square actually. Similar problems occur for other figures

mnyshenko (changes requested):

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,14 @@
...
+    protected String color;
+
+    public Figure(String color) {
+        this.color = color;
+    }
+
+    public String getColor() {
+        return color;
+    }
+
```
d1sam, line 13: Please don't add this empty redundant line.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private final Random random = new Random();
+    private final Color[] colorSet = Color.values();
```
d1sam, line 7: Please choose better name for this variable, because, it's not a Set, it's an array :) You will learn about `Set` in the nearest future.

src/test/java/core/test/StructureTest.java
d1sam, line 1: Please don't make changes inside test package.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,43 @@
...
+
+    public Figure getDefaultFigure() {
+        return new Circle(DEFAULT_RADIUS, DEFAULT_COLOR);
+    }
+
+    private int getRandomNumber() {
+        if (random.nextInt(MAX_LENGTH) == 0) {
+            return random.nextInt(MAX_LENGTH);
+        }
+        return random.nextInt(MAX_LENGTH);
```
d1sam, line 41: Please simplify this. 
The best solution is to just always add 1 and request random numbers from a smaller range:
`rand.random(MAX_LENGTH - 1) + 1;`

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,43 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int MAX_LENGTH = 100;
+    private static final int FIGURE_COUNT = 5;
+    private static final int DEFAULT_RADIUS = 10;
+    private static final String DEFAULT_COLOR = String.valueOf(Color.WHITE);
```
d1sam, line 9: Use name() for getting String representation of enum constants
Don’t use toString() or String.valueOf()(it will call toString() under the hood) for getting String representation of enum constants. toString() is common for all enum constants. If you override this method like below:
 ```
@Override
 public String() toString() {
    return "default";
 }
```
then for every constant toString() will be returning default, that’s not ok. So it’s better to use standard method of enum name() that will be returning always String representation of concrete enum constant.

**Review by d1sam — CHANGES_REQUESTED** [review]
Good job, but some changes should be done.

File: src/main/java/core/basesyntax/FigureSupplier.java (hunk 12 above)
Comment (forzen123212) [reply src/main/java/core/basesyntax/FigureSupplier.java]: image

File: src/test/java/core/test/StructureTest.java [hunk 13]
```diff
@@ -16,7 +16,7 @@
 
 public class StructureTest {
   private static final List<String> figureClassNames = List
-      .of("Circle", "Square", "IsoscelesTrapezoid", "Rectangle", "RightTriangle");
+          .of("Circle", "Square", "IsoscelesTrapezoid", "Rectangle", "RightTriangle");
```
bhdnchui, line 19: I think you should only provide a solution and not change tests.

src/test/java/core/test/StructureTest.java
```diff
@@ -25,8 +25,8 @@ public static void init() {
       allClasses = getClasses("core.basesyntax");
       if (allClasses.size() == 0) {
         Assert.fail("You should not rename base core.basesyntax package "
-            + "and path to project and project name should not contain spaces "
-            + "or words in cyrillic");
+                + "and path to project and project name should not contain spaces "
+                + "or words in cyrillic");
```
bhdnchui, line 29: I think you should only provide a solution and not change tests.

src/test/java/core/test/StructureTest.java
```diff
@@ -152,4 +152,4 @@ private static List<Class> findClasses(File directory, String packageName)
     }
     return classes;
   }
-}
+}
```
bhdnchui, line 155: The same as above.

bhdnchui (approved):
Good job, just pay attention to my comments.

PR #1284, opened 2023-09-28
implemented all the methods from the task

YevheniiVlasenko:
I'm not sure if it could be solved without interface implementation on the Figure class. But in this case, it comes to overridden methods from the interface in the Top Class, and it looks like the interface becomes useless because of method inheritance. Kinda closed circle.

### DIFF
--- a/src/main/java/core/basesyntax/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/FigureSupplier.java
@@ -1,0 +1,38 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int MAX_FIGURES_TYPES = 5;
+    private static final int SIZE_LIMIT = 5; //додано виключно щоб не генерувалися великі int
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private Random random = new Random(); //Об"єкт рандому
+
+    public Figure getRandomFigure() {
+        Color randomColor = colorSupplier.getRandomColor();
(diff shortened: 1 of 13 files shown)


bhdnchui (changes requested):
Good job!
Please, pay attention to my comments.

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,22 @@
...
+            if (i < figures.length / 2) {
+                figures[i] = figureSupplier.getDefaultFigure();
+            } else {
+                figures[i] = figureSupplier.getRandomFigure();
+            }
+        }
+
+        for (Drawable figure : figures) {
+            figure.draw();
+        }
```
mnyshenko, line 20: You can draw a figure right after you've created it, eliminating the need to iterate over all the figures

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,43 @@
...
+                int trapezoidSideB = getRandomSize();
+                int trapezoidHeight = getRandomSize();
+                return new IsoscelesTrapezoid(sideA, trapezoidSideB, trapezoidHeight, randomColor);
+            default:
+                return getDefaultFigure();
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(DEFAULT_CIRCLE_RADIUS, Color.WHITE);
```
mnyshenko, line 41: Minor: it's better to make Color.WHITE as a constant too with a proper name to give some context as you did with radius

mnyshenko (approved):

PR #1295, opened 2023-10-01
Added solution for jv-oop-advanced task

AgroFix:


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,41 @@
...
+            case 4:
+                return new Square(colorSupplier.getRandomColor(),
+                        random.nextInt(MAX_LENGTH));
+            default:
+                return getDefaultFigure();
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(Color.WHITE,DEFAULT_RADIUS);
```
d1sam, line 39: Please prettify format of this class. e.g. 
```suggestion
        return new Circle(Color.WHITE, DEFAULT_RADIUS);
```

d1sam (changes requested):
Good job, but there are some changes, that should be done. Please pay attention to places where you use String representation of enum Color. 
Use name() for getting String representation of enum constants
Don’t use toString() or String.valueOf()(it will call toString() under the hood) for getting String representation of enum constants. toString() is common for all enum constants. If you override this method like below:
 ```
@Override
 public String() toString() {
    return "default";
 }
```
then for every constant toString() will be returning default, that’s not ok. So it’s better to use standard method of enum name() that will be returning always String representation of concrete enum constant.
Also please check [this](https://www.javatpoint.com/java-string-format) article. It's better to use this method inside `draw`, instead String concatenation.

File: src/main/java/core/basesyntax/ColorSupplier.java [hunk 2]
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private Random random = new Random();
+
+    public Color getRandomColor() {
+        int index = random.nextInt(Color.values().length);
+        return Color.values()[index];
```
bhdnchui, line 10: I think you can make `Color.values()` result as a class field.

src/main/java/core/basesyntax/HelloWorld.java
```diff
@@ -1,8 +1,21 @@
...
+        Figure[] figures = new Figure[ARRAY_LENGTH];
+        for (int i = 0; i < ARRAY_LENGTH; i++) {
+            if (i < ARRAY_LENGTH / 2) {
+                figures[i] = figureSupplier.getRandomFigure();
+            } else {
+                figures[i] = figureSupplier.getDefaultFigure();
+            }
+        }
+
+        for (Figure f : figures) {
```
bhdnchui, line 17: Do not use one-letter names.

bhdnchui (changes requested):
Good job, just minor updates are required.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,41 @@
...
+                        random.nextInt(MAX_LENGTH));
+            case 3:
+                return new Rectangle(colorSupplier.getRandomColor(),
+                        random.nextInt(MAX_LENGTH),
+                        random.nextInt(MAX_LENGTH));
+            case 4:
+                return new Square(colorSupplier.getRandomColor(),
+                        random.nextInt(MAX_LENGTH));
+            default:
+                return getDefaultFigure();
```
mnyshenko, line 34: 1) This part has a problem with code duplication. You generate random sizes using one formula. Let's imagine tomorrow you'll need to change this formula, then you'll have to go over each line and make a change. It's insufficient. In this case, you can create a separate method for this computation and reuse it.
2) Since each figure needs a color you can extract colorSupplier.getRandomColor() to a separate variable and reuse

src/main/java/core/basesyntax/HelloWorld.java
```diff
@@ -1,8 +1,21 @@
...
+            if (i < ARRAY_LENGTH / 2) {
+                figures[i] = figureSupplier.getRandomFigure();
+            } else {
+                figures[i] = figureSupplier.getDefaultFigure();
+            }
+        }
+
+        for (Figure figure : figures) {
+            figure.draw();
+        }
```
mnyshenko, line 19: You can draw a figure right after you've created it, eliminating the need to iterate over all the figures

mnyshenko (approved):

PR #1296, opened 2023-10-01
added solution according to README.md

vadymhrnk:


src/main/java/core/basesyntax/IsoscelesTrapezoid.java
```diff
@@ -0,0 +1,37 @@
...
+            Color color, double topParallelSide,
+            double bottomParallelSide, double side) {
+        super(color);
+        if (topParallelSide < bottomParallelSide) {
+            this.topParallelSide = topParallelSide;
+            this.bottomParallelSide = bottomParallelSide;
+        } else {
+            this.topParallelSide = bottomParallelSide;
+            this.bottomParallelSide = topParallelSide;
+        }
```
d1sam, line 18: If you are confident that your users will always pass the correct values (i.e., topParallelSide is always less than or equal to bottomParallelSide), then you could disable this check. However, it is usually better to be more secure and add this check to avoid incorrect use of the class.

In summary, good catch 👍

d1sam (approved):
Good job, I approve your work, but please check [this](https://www.javatpoint.com/java-string-format) article. It's better to use this method inside draw, instead String concatenation.

PR #1301, opened 2023-10-02
printed information about created figures

denysokhrymovych:


mnyshenko (changes requested):

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,31 @@
...
+                return new Circle(colorSupplier.getRandomColor(), random.nextDouble());
+            case 2:
+                return new RightTriangle(colorSupplier.getRandomColor(),
+                    random.nextDouble(), random.nextDouble());
+            case 3:
+                return new Rectangle(colorSupplier.getRandomColor(),
+                    random.nextDouble(), random.nextDouble());
+            default:
+                return new Square(colorSupplier.getRandomColor(), random.nextDouble());
+        }
```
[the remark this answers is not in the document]
denysokhrymovych: Thank you, but I'm wondering how to do it because it seems to me that if we assign a random value to a variable, this variable is the same every time

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private static final String[] colors = {"white", "yellow", "green", "blue", "purple"};
```
d1sam, line 6: You should use enum for you colors .

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,15 @@
...
+
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        for (int i = 0; i < NUMBER_OF_FIGURES_IN_EACH_OF_TWO_GROUPS; i++) {
+            figureSupplier.getRandomFigure().draw();
+        }
+        for (int i = 0; i < NUMBER_OF_FIGURES_IN_EACH_OF_TWO_GROUPS; i++) {
+            figureSupplier.getDefaultFigure().draw();
+        }
+    }
```
d1sam, line 14: Please pay attention to condition of the task:
In the main() method we need to create an array of figures (the size of array can be 3 or 6, it doesn't matter). The first half of figures in this array should be generated with random parameters.

d1sam (changes requested):
Good job, but please see my comments below.

kshuryhin (approved):
Great job!

PR #1309, opened 2023-10-05
Created all classes of figures, calculated their area and generated a…

tbasy:
… random color

### DIFF
--- a/src/main/java/core/basesyntax/Circle.java
+++ b/src/main/java/core/basesyntax/Circle.java
@@ -1,0 +1,22 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+    private final int radius;
+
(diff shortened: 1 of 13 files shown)


src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,22 @@
...
+
+public class Circle extends Figure {
+    private final int radius;
+
+    public Circle(int radius, String color) {
+        super(color);
+        this.radius = radius;
+    }
+
+    public double getArea() {
```
olekskov, line 11: while @override annotation is technically optional it's better to use it to clearly indicate inherited methods and separate them from own ones

olekskov (approved):
Good job, approved
Please check comment below for further tasks

PR #1311, opened 2023-10-06
make task

Paradoxallist:


src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,23 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+    private int radius;
+
+    public Circle(int radius,String color) {
+        this.radius = radius;
+        setColor(color);
+        getArea();
```
Rommelua, line 9:
```suggestion
        super(color);
        this.radius = radius;
```

Rommelua (changes requested):

Before you create PR with your homework, you need to go through the checklist under the task and correct all the points described there. The mentor will not check the work until the checklist points are corrected

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,22 @@
...
+    @Override
+    public double getArea() {
+        return Math.PI * radius * radius;
+    }
+
+    @Override
+    public String toString() {
+        System.out.println("Circle{" + "radius=" + radius
+                + ", color='" + getColor() + '\'' + ", area=" + getArea() + '}');
+        return "";
```
sokimaaa, line 20:
```suggestion
        return "Circle{" + "radius=" + radius
                + ", color='" + getColor() + '\'' + ", area=" + getArea() + '}';
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private final Random rand = new Random();
+    private final String[] colors = new String[]{"white", "black", "blue", "red"};
```
sokimaaa, line 7: make Color enum

src/main/java/core/basesyntax/Circle.java
sokimaaa: behaviour - we can obtain the area of any figure and are able to draw it. To 'draw' means to print out all information about a figure using System.out.println() (`you shouldn't override toString() method for this)`.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,43 @@
...
+
+public class FigureSupplier {
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final Random random = new Random();
+    private final String[] listFigure = new String[]
+            {"Circle", "Square", "Rectangle", "RightTriangle", "IsoscelesTrapezoid"};
+
+    public Figure getRandomFigure() {
+        String randomFigure = listFigure[random.nextInt(listFigure.length)];
+        int randomRange = 100;
```
sokimaaa, line 13: should be class level constant.

`randomRange` is misleading name

src/main/java/core/basesyntax/IsoscelesTrapezoid.java
```diff
@@ -0,0 +1,28 @@
...
+    public double getArea() {
+        return (upSide + downSide) * height / 2;
+    }
+
+    @Override
+    public String toString() {
+        System.out.println("IsoscelesTrapezoid{" + "upSide=" + upSide
+                + ", downSide=" + downSide
+                + ", height=" + height + ", color='"
+                + getColor() + '\'' + ", area=" + getArea() + '}');
```
sokimaaa, line 25: behaviour - we can obtain the area of any figure and are able to draw it. To 'draw' means to print out all information about a figure using System.out.println() `(you shouldn't override toString() method for this).`

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,15 @@
...
+public class Main {
+    public static void main(String[] args) {
+        Figure[] figures = new Figure[4];
+        FigureSupplier figureSupplier = new FigureSupplier();
+        figures[0] = figureSupplier.getRandomFigure();
+        figures[1] = figureSupplier.getRandomFigure();
+        figures[2] = new Circle("white", 10);
+        figures[3] = new Square("black", 5);
+        for (Figure figure : figures) {
+            figure.toString();
```
sokimaaa, line 12: consider adding interface instead of overriding toString

sokimaaa (changes requested):

src/main/java/core/basesyntax/Color.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public enum Color {
+    white,
+    black,
```
Rommelua, line 5: uppercase

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,15 @@
...
+
+    public Figure(Color color) {
+        this.color = color;
+    }
+
+    public Color getColor() {
+        return color;
+    }
+
+    public abstract void draw();
```
Rommelua, line 14: image

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,41 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private final int maxNumber = 100;
```
Rommelua, line 6: image

src/main/java/core/basesyntax/FigureType.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum FigureType {
+    Circle,
+    Square,
+    Rectangle,
+    RightTriangle,
+    IsoscelesTrapezoid
```
Rommelua, line 8: uppercase

Rommelua (changes requested):
Before you create PR with your homework, you need to go through the checklist under the task and correct all the points described there.

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,21 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+    private final int radius;
+
+    public Circle(Color color,int radius) {
```
kerrrusha, line 6: There should be a space after the comma in the parameters list of the constructor.

src/main/java/core/basesyntax/Square.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax;
+
+public class Square extends Figure {
+    private final double side;
```
kerrrusha, line 4: this fields is marked as 'final' which means that it can't be changed after it's initialized. This is okay if you will never need to change this value, but if you want your objects to be mutable, you should remove the 'final' keyword.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,41 @@
...
+                        random.nextInt(MAX_NUMBER),
+                        random.nextInt(MAX_NUMBER),
+                        random.nextInt(MAX_NUMBER));
+            default:
+                return getDefaultFigure();
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(Color.WHITE,10);
```
kerrrusha, line 38: image

kerrrusha (changes requested):

src/main/java/core/basesyntax/FigureSupplier.java
Paradoxallist: In this case, I create the default shape Circle (color - white, radius - 10) and in this case this is the case, since I know that I am creating a circle that only has a radius and color. Because in this case, I have to make the color a constant, and when I want to see what kind of circle I’m creating, I’ll have to look for it somewhere above. In your example, this was the case because we define the important variable FIGURE_COUNT, and in this case we are dealing with 1 default figure, and we even know that it will be a circle. I understand that you are trying to teach me the correct code style, but it seems to me that not all “magic numbers” need to be written into a constant, at least to optimize time and not clutter up the class with constants.

src/main/java/core/basesyntax/Square.java
Paradoxallist: This object is designed to be immutable, which does not change after it is created. In fact, I should have made the shape classes themselves final (Circle, Square and others).

kerrrusha (approved):

PR #1322, opened 2023-10-11
jv-oop-advanced-solution to check

EugeneKorniuk:


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,43 @@
...
+            case 4:
+                figure = new RightTriangle(colorSupplier.getRandomColor(),
+                        random.nextInt(20), random.nextInt(20));
+                break;
+            case 5:
+                figure = new IsoscelesTrapezoid(colorSupplier.getRandomColor(),
+                        random.nextInt(20), random.nextInt(20), random.nextInt(20));
+                break;
+            default:
+                return null;
```
andrii-hoienko, line 35: I believe in the default case it is better to return defaultFigure instead of null, user does not expect null to be returned in this methid

src/main/java/core/basesyntax/AreaCalculator.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
```
andrii-hoienko, line 1: It would be great if you will create packages and put figures in 'figure' package, suppliers to 'supplier'

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+public class Main {
+    public static void main(java.lang.String[] args) {
+        Figure[] figures = new Figure[6];
```
andrii-hoienko, line 5: Magic number, it makes sense to reuse Figure Supplier#FIGURE_COUNT

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,17 @@
...
+        FigureSupplier figureSupplier = new FigureSupplier();
+        for (int i = 0; i < figures.length / 2; i++) {
+            figures[i] = figureSupplier.getRandomFigure();
+        }
+        for (int y = figures.length / 2; y < figures.length; y++) {
```
andrii-hoienko, line 10: Optional: it is common to use 'i' then 'j' as a variable name in for cycle, but you can you 'i' here as well

andrii-hoienko: Also to make code more readable it makes sense to create a method something like 'polulateFigures' or 'getFigures' and put the logic of creating and populating array there

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,30 @@
...
+    @Override
+    public String draw() {
+        return "Figure: circle, area: " + areaCalculator()
+                + " sq.units, radius: " + this.radius
+                + " units, color: " + this.getColor();
+    }
+
+    @Override
+    public int areaCalculator() {
+        return (int)Math.PI * this.radius * this.radius;
```
andrii-hoienko, line 28: When you cast result of this expression to int value can be lost, so maybe it will be better to make areaCalculator return double

src/main/java/core/basesyntax/AreaCalculator.java
```diff
@@ -0,0 +1,5 @@
...
+
+public interface AreaCalculator {
+    int areaCalculator();
```
andrii-hoienko, line 4: Bad method name, 'calculateArea' or something like this will be better

andrii-hoienko (changes requested):

PR #1328, opened 2023-10-17
oop_advanced_try_#1

slava-moiseev-kramatorsk:


src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,58 @@
+package core.basesyntax;
+
+public abstract class Figure implements Draw {
+    private final String name;
+    private int sideA;
+    private int sideB;
+    private int side;
+    private int height;
+    private final String color;
```
olekskov, line 9: Parent class should hold only common properties of child classes.
Current approach makes code hard to use.
e.g. you have constructor of Circle with "side" parameter. Which side? Radius? Diameter? Can only investigate based on formulas used
```suggestion
    private final String color;
```

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+    public Circle(String name, int side, String color) {
+        super(name, side, color);
+    }
+
+    @Override
+    public void getsSquare() {
+        System.out.println("Figure: " + getName()
```
olekskov, line 10: Instead of name property you may use class name and predefined method to get it
```suggestion
        System.out.println("Figure: " + getClass().getSimpleName()
```

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,16 @@
...
+                + "side " + getSide()
+                + " Area " + (Math.PI * getSide() * getSide())
+                + " Color: " + getColor());
+    }
```
olekskov, line 14: You merged 2 different things.
1 - getArea() that should return some decimal value
2 - draw() that should print info regarding figure (and may reuse getArea() method of course)

Please introduce second interface and split logic between 2 different methods

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,31 @@
...
+        Figure[] figure = new Figure[] {new Square("square",
+                sideOfFigureA, colorSupplier.getRandomColor()),
+                new RightTriangle("RightTriangle", sideOfFigureA, sideOfFigureB,
+                colorSupplier.getRandomColor()),
+                new Circle("circle", sideOfFigureA,
+                colorSupplier.getRandomColor()),
+                new Rectangle("rectangle", sideOfFigureA, sideOfFigureB,
+                colorSupplier.getRandomColor()),
+                new IsoscelesTrapezoid("IsoscelesTrapezoid", sideOfFigureA, sideOfFigureB,
+                height, colorSupplier.getRandomColor())};
```
olekskov, line 22: You create several objects to get only 1 of them. It's not good.
Please add logic to determine random figure type and then create only 1 object

olekskov (changes requested):

slava-moiseev-kramatorsk:
Зробив ркомендаціі, та виправив звуваження окрім одного. Я можу поки зрозуміти яку логіку треба зробити для рандомного вибору фігури, та ще й передаваті туди рандомні аргументи. Оскільки у різних фігур різна кільцість аргументів конструктора. Прошу підказку, не хочу дивитись розв'язку ментора, адже зробив 80% задачі.

src/main/java/core/basesyntax/HelloWorld.java
```diff
@@ -4,5 +4,20 @@
...
 
+        Figure[] array = new Figure[6];
+        FigureSupplier supplier = new FigureSupplier();
+        for (int i = 0; i < array.length / 2; i++) {
+            array[i] = supplier.getRandomFigure();
+        }
+
+        for (int i = array.length / 2; i < array.length; i++) {
+            array[i] = supplier.getDefault();
+        }
```
kozhukhovsky, line 17:
```suggestion
        for (int i = 0; i < array.length / 2; i++) {
            array[i] = supplier.getRandomFigure();
        }

        for (int i = array.length / 2; i < array.length; i++) {
            array[i] = supplier.getDefault();
        }
```

use one loop

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,18 @@
...
+
+    public Figure(String color) {
+        this.color = color;
+    }
+
+    public String getColor() {
+        return color;
+    }
+
+    public abstract double getSquare();
```
kozhukhovsky, line 14: make separate interface for area, like draw

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,32 @@
...
+                new Rectangle(sideA, sideB,
+                colorSupplier.getRandomColor()),
+                new IsoscelesTrapezoid(sideA, sideB,
+                height, colorSupplier.getRandomColor())};
+        int index = random.nextInt(figure.length);
+        return figure[index];
+    }
+
+    public Figure getDefault() {
+        return new Circle(10,
```
kozhukhovsky, line 29: move 10 to the constant

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,32 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final Random random = new Random();
+    private final int radius = random.nextInt(1,10);
```
kozhukhovsky, line 8: you need to create these fields in the method, then they will be different each time for each call

Now every call will receive the same message

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,32 @@
...
+    private final int sideA = random.nextInt(1,10);
+    private final int sideB = random.nextInt(1,10);
+    private final int height = random.nextInt(1, 10);
+
+    public Figure getRandomFigure() {
+        Figure[] figure = new Figure[] {new Square(sideA,
```
kozhukhovsky, line 14: You need to create a switch/case
Where we will get the number by the number of figures. And create a separate figure for each number. We will also not get different shapes through the array.

kozhukhovsky (changes requested):

src/main/java/core/basesyntax/HelloWorld.java
slava-moiseev-kramatorsk: image
  can't use one loop, the condition is always true

src/main/java/core/basesyntax/Calculate.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Calculate {
```
kozhukhovsky, line 3:
```suggestion
public interface AreaCalculator {
```

src/main/java/core/basesyntax/Draw.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Draw {
```
kozhukhovsky, line 3:
```suggestion
public interface Drawable {
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,35 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final Random random = new Random();
+    private final int radiusOfCircle = 10;
```
kozhukhovsky, line 8:
```suggestion
    private static final int DEFAULT_RADIUS = 10;
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,35 @@
...
+                    colorSupplier.getRandomColor());
+            case 4 -> new Rectangle(sideA, sideB,
+                    colorSupplier.getRandomColor());
+            case 5 -> new IsoscelesTrapezoid(sideA, sideB,
+                    height, colorSupplier.getRandomColor());
+            default -> null;
+        };
+    }
+
+    public Figure getDefault() {
```
kozhukhovsky, line 31:
```suggestion
    public Figure getDefaultFigure() {
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,35 @@
...
+                    colorSupplier.getRandomColor());
+            case 2 -> new RightTriangle(sideA, sideB,
+                    colorSupplier.getRandomColor());
+            case 3 -> new Circle(radius,
+                    colorSupplier.getRandomColor());
+            case 4 -> new Rectangle(sideA, sideB,
+                    colorSupplier.getRandomColor());
+            case 5 -> new IsoscelesTrapezoid(sideA, sideB,
+                    height, colorSupplier.getRandomColor());
+            default -> null;
```
kozhukhovsky, line 27:
```suggestion
            default -> getDefaultFigure();
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,35 @@
...
+        };
+    }
+
+    public Figure getDefault() {
+        return new Circle(radiusOfCircle,
+                Color.WHITE.toString().toLowerCase());
```
kozhukhovsky, line 33:
```suggestion
                Color.WHITE.name().toLowerCase());
```

src/main/java/core/basesyntax/HelloWorld.java
kozhukhovsky: You need if else, not do while.

src/main/java/core/basesyntax/Square.java
```diff
@@ -0,0 +1,26 @@
+package core.basesyntax;
+
+public class Square extends Figure {
+    private final int side;
+
+    public Square(int side, String color) {
+        super(color);
+        this.side = side;
+    }
+
```
kozhukhovsky, line 10: Not everywhere there is an `override` annotation for this method. Add in all necessary classes.

kozhukhovsky (changes requested):

src/main/java/core/basesyntax/HelloWorld.java
```diff
@@ -4,5 +4,29 @@
...
+            }
+        }
+
+        for (int i = 0; i < array.length / 2; i++) {
+            array[i] = supplier.getRandomFigure();
+        }
+
+        for (int i = array.length / 2; i < array.length; i++) {
+            array[i] = supplier.getDefaultFigure();
+        }
```
olekskov, line 26: these lines fully duplicates logic on lines 12-18
```suggestion
```

src/main/java/core/basesyntax/HelloWorld.java
```diff
@@ -4,5 +4,29 @@
...
+        for (Figure figure : array) {
+            figure.getArea();
+            figure.draw();
+        }
```
olekskov, line 30: 1) result of getArea() method is ignored so no sense to call it
2) call of draw() method mey be included into first cycle
```suggestion
```

src/main/java/core/basesyntax/HelloWorld.java
```diff
@@ -4,5 +4,29 @@
...
+        Figure[] array = new Figure[6];
+        FigureSupplier supplier = new FigureSupplier();
+
+        for (int i = 0; i < array.length; i++) {
+            if (i <= array.length / 2) {
+                array[i] = supplier.getRandomFigure();
+            } else {
+                array[i] = supplier.getDefaultFigure();
+            }
+        }
```
olekskov, line 18: like this:
```suggestion
        for (int i = 0; i < array.length; i++) {
            if (i <= array.length / 2) {
                array[i] = supplier.getRandomFigure();
            } else {
                array[i] = supplier.getDefaultFigure();
            }
            array[i].draw();
        }
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,35 @@
...
+public class FigureSupplier {
+    private static final int DEFAULT_RADIUS = 10;
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final Random random = new Random();
+
+    public Figure getRandomFigure() {
+        int radius = random.nextInt(1,10);
+        int sideA = random.nextInt(1,10);
+        int sideB = random.nextInt(1,10);
+        int height = random.nextInt(1, 10);
```
olekskov, line 14: better to avoid code repeatability
```suggestion
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,35 @@
...
+            case 3 -> new Circle(radius,
+                    colorSupplier.getRandomColor());
+            case 4 -> new Rectangle(sideA, sideB,
+                    colorSupplier.getRandomColor());
+            case 5 -> new IsoscelesTrapezoid(sideA, sideB,
+                    height, colorSupplier.getRandomColor());
+            default -> getDefaultFigure();
+        };
+    }
+
```
olekskov, line 30: you can introduce private method and reuse it:
```suggestion
        return switch (index) {
            case 1 -> new Square(getRandomSide(),
                    colorSupplier.getRandomColor());
            case 2 -> new RightTriangle(getRandomSide(), getRandomSide(),
                    colorSupplier.getRandomColor());
            case 3 -> new Circle(getRandomSide(),
                    colorSupplier.getRandomColor());
            case 4 -> new Rectangle(getRandomSide(), getRandomSide(),
                    colorSupplier.getRandomColor());
            case 5 -> new IsoscelesTrapezoid(getRandomSide(), getRandomSide(),
                    getRandomSide(), colorSupplier.getRandomColor());
            default -> getDefaultFigure();
        };
    }

    private int getRandomSide() {
        random.nextInt(1, SIDE_LIMIT)
    }
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,35 @@
...
+    private static final int DEFAULT_RADIUS = 10;
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final Random random = new Random();
+
+    public Figure getRandomFigure() {
+        int radius = random.nextInt(1,10);
+        int sideA = random.nextInt(1,10);
+        int sideB = random.nextInt(1,10);
+        int height = random.nextInt(1, 10);
+        int index = random.nextInt(1, 5);
```
olekskov, line 15: Better to extract magic numbers to constant with proper naming
```suggestion
        int index = random.nextInt(1, NUMBER_OF_FIGURES);
```

olekskov (changes requested):

src/main/java/core/basesyntax/HelloWorld.java
```diff
@@ -4,5 +4,18 @@
  * Feel free to remove this class and create your own.
  */
 public class HelloWorld {
+    public static void main(String[] args) {
 
+        Figure[] array = new Figure[6];
```
okuzan, line 9:
```suggestion
        Figure[] figures = new Figure[6];
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,37 @@
...
+        };
+    }
+
+    private int getRandomSide() {
+        return random.nextInt(1, SIDE_LIMIT);
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(DEFAULT_RADIUS,
+                Color.WHITE.name().toLowerCase());
```
okuzan, line 35:
```suggestion
                DEFAULT_COLOR);
```

okuzan (approved):
Good job!

PR #1360, opened 2023-11-13
solved the task using OOP principles

ynik404:


src/main/java/core/basesyntax/Application.java
```diff
@@ -0,0 +1,25 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Application {
+    private static final Random RANDOM_NUMBER = new Random();
```
sokimaaa, line 6: Should be inside main method
```suggestion
    private static final Random RANDOM_NUMBER = new Random();
```

src/main/java/core/basesyntax/Application.java
```diff
@@ -0,0 +1,25 @@
...
+    private static final int ARRAY_LENGTH = RANDOM_NUMBER.nextInt(10);
```
sokimaaa, line 7: why do you need array_length to be random?

src/main/java/core/basesyntax/FigureDrawer.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface FigureDrawer {
```
sokimaaa, line 3: it's better to have a generic interface that could be reused for multiple cases, e.g. drawing not only figure.
```suggestion
public interface Drawer {
```

sokimaaa (changes requested):

sokimaaa (approved):

PR #1361, opened 2023-11-14
first commit

bkostanetskyi:


src/main/java/core/basesyntax/AreaCalculator.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax;
+
+public interface AreaCalculator {
+    double getArea();
+
+}
```
Rommelua, line 6:
```suggestion
    double getArea();
}
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private Random random = new Random();
+
+    public String getRandomColor() {
+        Color[] colors = Color.values();
+        return colors[random.nextInt(colors.length)].toString();
```
Rommelua, line 10: image

src/main/java/core/basesyntax/Drawable.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax;
+
+public interface Drawable {
+    void draw();
+
+}
```
Rommelua, line 6:
```suggestion
    void draw();
}
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax;
+
+public class Main {
+
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] arr = new Figure[6];
```
Rommelua, line 7: use constant

Rommelua (changes requested):
good job

src/main/java/core/basesyntax/Main.java
bkostanetskyi: Can you explain to me why we need to use an interface here if we use it in literally every class, i.e. we could just make an abstract class?

Rommelua (approved):

PR #1366, opened 2023-11-16
frist solution of this aplication

KoderNL:


src/main/java/core/basesyntax/Draw.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Draw {
```
sokimaaa, line 3: bad name for interface, it should describe some action, e.g. drawable

sokimaaa (approved):

PR #1374, opened 2023-11-29
done task

RomanSoproniuk:


kozhukhovsky (changes requested):
You don't need to set random numbers in the figures right away. There is also no need to print the figure directly in the constructor. You will need to call the draw() method in main for each shape created.
Regarding random parameters, each figure must accept each of its sides as input.
The FigureSupplier class should be responsible for creating random sides. Everything is the same there, but you need to add generation for each side for each shape before creating it.

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,23 @@
+package core.basesyntax;
+
+public class Figure implements AreaCalculator, Drawable {
```
kozhukhovsky, line 3:
```suggestion
public abstract class Figure implements AreaCalculator, Drawable {
```

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,23 @@
...
+
+    @Override
+    public double getArea() {
+        return 0;
+    }
+
+    @Override
+    public void draw() {
+
+    }
```
kozhukhovsky, line 22:
```suggestion
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,21 @@
+package core.basesyntax;
+
+public class Main {
+
+    public static void main(String[] args) {
+
```
kozhukhovsky, line 6:
```suggestion
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,21 @@
...
+                figures[i] = randomFigure;
+                figures[i].draw();
+            } else {
+                figures[i] = figureSupplier.getDefaultFigure();
+                figures[i].draw();
+            }
+        }
+    }
+}
+
```
kozhukhovsky, line 21:
```suggestion
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,21 @@
...
+        Figure[] figures = new Figure[6];
+        FigureSupplier figureSupplier = new FigureSupplier();
+        for (int i = 0; i < figures.length; i++) {
+            if (i < figures.length / 2) {
+                Figure randomFigure = figureSupplier.getRandomFigure();
+                figures[i] = randomFigure;
+                figures[i].draw();
+            } else {
+                figures[i] = figureSupplier.getDefaultFigure();
+                figures[i].draw();
```
kozhukhovsky, line 16: better do separate loop for drawing figures

kozhukhovsky (changes requested):
Good job!

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,27 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+    private final String circleColor;
```
kozhukhovsky, line 4: no need this field for each figures type
you have field color in Figure class

kozhukhovsky (changes requested):
Need small fixes :)

pom.xml
```diff
@@ -9,7 +9,7 @@
     <version>1.0-SNAPSHOT</version>
 
     <properties>
-        <jdk.version>17</jdk.version>
+        <jdk.version>11</jdk.version>
```
olekskov, line 12: It's not critical for this task but I really recommend to uplift your local java version to 17 as soon your tasks will have problem with compilation

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,23 @@
...
```
olekskov, line 22: not removed

olekskov (approved):
Good job overall, approved.
Please pay attention to comments for future

PR #1379, opened 2023-12-02
solutin oop-advanced VdBondarev

VdBondarev:


okuzan (changes requested):

src/main/java/core/basesyntax/AreaCalculator.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface AreaCalculator {
+    double getArea();
```
kerrrusha, line 4: if you've named interface like Calculator - let him calculate)
```suggestion
    double calculateArea();
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private int index = new Random().nextInt(Colors.values().length);
```
kerrrusha, line 6: index should be calculated inside `getRandomColor()` method. `Random` should be `private final` class-level field

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,11 @@
...
+
+    public String getRandomColor() {
+        return String.valueOf(Colors.values()[index]);
```
kerrrusha, line 9:
```suggestion
        return Colors.values()[index].name();
```

src/main/java/core/basesyntax/Drawing.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Drawing {
```
kerrrusha, line 3: invalid interface name
```suggestion
public interface Drawable {
```

kerrrusha (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+
+    public String getRandomColor() {
+        int index = new Random().nextInt(Color.values().length);
+        //return String.valueOf(Colors.values()[index]);
```
okuzan, line 9: remove comments before submitting a solution

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,36 @@
...
+    private static final String DEFAULT_COLOR = "WHITE";
+    private static final int DEFAULT_VALUE = 10;
+    private static final int NUMBER_OF_FIGURES = 5;
+    private static final int MAX_VALUE = 15;
+    private final Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        int number = random.nextInt(NUMBER_OF_FIGURES);
+        int first = random.nextInt(MAX_VALUE);
```
okuzan, line 15: bad naming
let's call it smth like `geometricProperty`
or even better create a method for it. Also, make sure this method can't return 0

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,13 @@
...
+public abstract class Figure implements AreaCalculator, Drawable {
+    private String color;
+
+    public Figure(String color) {
+        this.color = color;
+    }
+
+    public String getColor() {
+        return color;
+    } // I need this getter to get color in overrided method "draw" in every figure class
```
okuzan, line 12: We know, remove the comment :) 
If you want to clarify smth better do it in the commit messages than in the actual source code

PR #1385, opened 2023-12-11
Solution

suprun-anna:
created classes for figures, interfaces for behavior, abstract class for state, main method


Rommelua (approved):
good

PR #1388, opened 2023-12-14
Hw oop advanced

RCherniavskiy:


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,29 @@
...
+        Figure circle = new Circle(colorSupplier.getRandomColor(),random.nextInt(MAX_FOR_FIGURE));
+        Figure trapezoid = new IsoscelesTrapezoid(colorSupplier.getRandomColor(),
+                random.nextInt(MAX_FOR_FIGURE),random.nextInt(MAX_FOR_FIGURE),
+                random.nextInt(MAX_FOR_FIGURE));
+        Figure rectangle = new Rectangle(colorSupplier.getRandomColor(),
+                random.nextInt(MAX_FOR_FIGURE),random.nextInt(MAX_FOR_FIGURE));
+        Figure triangle = new RightTriangle(colorSupplier.getRandomColor(),
+                random.nextInt(MAX_FOR_FIGURE),random.nextInt(MAX_FOR_FIGURE));
+        Figure square = new Square(colorSupplier.getRandomColor(),random.nextInt(MAX_FOR_FIGURE));
+        Figure[] figures = new Figure[]{circle,trapezoid,rectangle,triangle,square};
```
Sanderfuck, line 21: creating array you can separate in private method

src/main/java/core/basesyntax/FindArea.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax;
+
+public interface FindArea {
```
Sanderfuck, line 3: Better naming Calculate than Find

src/main/java/core/basesyntax/Print.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Print {
+    public void printInfo();
```
Sanderfuck, line 4:
```suggestion
    void printInfo();
```

src/main/java/core/basesyntax/FindArea.java
```diff
@@ -0,0 +1,6 @@
...
+    public double getArea();
```
Sanderfuck, line 4:
```suggestion
    double getArea();
```

src/main/java/core/basesyntax/IsoscelesTrapezoid.java
```diff
@@ -0,0 +1,25 @@
...
+    private double baseUp;
+    private double height;
+
+    public IsoscelesTrapezoid(String color,double baseDown,double baseUp,double height) {
+        super(color);
+        this.baseDown = baseDown;
+        this.baseUp = baseUp;
+        this.height = height;
+    }
+
```
Sanderfuck, line 14: Where is annotation Override?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -0,0 +1,23 @@
...
+    private double width;
+    private double height;
+
+    public Rectangle(String color,double width,double height) {
+        super(color);
+        this.width = width;
+        this.height = height;
+    }
+
+    public void printInfo() {
```
Sanderfuck, line 13: Where is annotation Override?

Sanderfuck: Fix it for every figure

Sanderfuck (changes requested):

src/main/java/core/basesyntax/CalculateArea.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax;
+
+public interface CalculateArea {
```
olekskov, line 3: naming of interfaces is better when it's noun or adjective
verbs are usually refer to methods naming
```suggestion
public interface AreaCalculator {
```

src/main/java/core/basesyntax/Print.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Print {
```
olekskov, line 3: same here
```suggestion
public interface Printable {
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,33 @@
...
+        return figures;
+    }
+
+    public Figure getRandomFigure() {
+        int index = random.nextInt(getArrayFigures().length);
+        return getArrayFigures()[index];
+    }
```
olekskov, line 28: With such approach you're creating 5 figures each time method is called while you need only one of them.
Please reconsider logic so you firstly determine type of random figure and only then create this particular figure without creation of all other ones

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,14 @@
...
+public class Main {
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        for (int i = 0; i < 6; i++) {
+            if (i < 3) {
+                figureSupplier.getRandomFigure().printInfo();
+            } else {
+                figureSupplier.getDefaultFigure().printInfo();
+            }
+        }
```
olekskov, line 12: task condition is not implemented:
"In the main() method we need to create an array of figures (the size of the array can be 3 or 6, it doesn't matter). The first half of figures in this array should be generated with random parameters."

olekskov (changes requested):

olekskov (approved):
Good job!

PR #1401, opened 2023-12-24
created and implemented all interfaces, classes and methods, as far a…

Ovikt0r:
…s I understand it

### DIFF
--- a/src/main/java/core/basesyntax/figures/Shape.java
+++ b/src/main/java/core/basesyntax/figures/Shape.java
@@ -1,0 +1,24 @@
+package core.basesyntax.figures;
+
+import core.basesyntax.figures.interfaces.Drawable;
+import core.basesyntax.figures.interfaces.ShapeCalculable;
+
+public class Shape implements ShapeCalculable, Drawable {
+    protected String color;
+    protected double area;
+
+    Shape(String color) {
+        this.color = color;
+        area = calculateArea();
+    }
+
+    @Override
+    public double calculateArea() {
+        return area;
+    }
+
+    @Override
+    public void draw() {
+        System.out.println("This is just a shape");
+    }
+}
--- a/src/main/java/core/basesyntax/suppliers/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/suppliers/FigureSupplier.java
@@ -1,0 +1,42 @@
+package core.basesyntax.suppliers;
+
+import core.basesyntax.figures.Circle;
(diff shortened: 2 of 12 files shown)


src/main/java/core/basesyntax/suppliers/FigureSupplier.java
```diff
@@ -0,0 +1,42 @@
...
+    public Shape getRandomFigure() {
+
+        Shape[] shapes = new Shape[]{
+                new Circle(colorSupplier.getRandomColor(),random.nextInt(numberLimit)),
+                new IsoscelesTrapezoid(colorSupplier.getRandomColor(),getRandomSide(),
+                        getRandomSide() * coefficient, getRandomSide()),
+                new Rectangle(colorSupplier.getRandomColor(),getRandomSide(),getRandomSide()),
+                new RightTriangle(colorSupplier.getRandomColor(), getRandomSide(), getRandomSide()),
+                new Square(colorSupplier.getRandomColor(), getRandomSide()),
+                new Circle(colorSupplier.getRandomColor(), getRandomSide())
```
Sanderfuck, line 28: You can create method for random color like for random size

src/main/java/core/basesyntax/suppliers/FigureSupplier.java
```diff
@@ -0,0 +1,42 @@
...
+        };
+
+        int randomIndex = random.nextInt(shapes.length);
+        return shapes[randomIndex];
```
Sanderfuck, line 32: Better use switch case for this situation, because every invoke of method create all objects but you get only one

Sanderfuck (changes requested):

src/main/java/core/basesyntax/figures/Shape.java
```diff
@@ -0,0 +1,22 @@
...
+    @Override
+    public void draw() {
+        System.out.println("It's just a shape");
+    }
+
+    @Override
+    public double calculateArea() {
+        return 0;
+    }
+}
```
olekskov, line 22: If we do not plan to have instance of parent class it's better to make it abstract
And then we may omit methods overriding and delegate it to children classes
```suggestion
public abstract class Shape implements Drawable, ShapeCalculable {
    protected String color;

    Shape(String color) {
        this.color = color;
    }
}
```

olekskov (approved):
Good job, approved.
Please pay attention to comments below

PR #1409, opened 2023-12-30
displayed the entire list of objects from the array

YuliiaNisha:


src/main/java/core/basesyntax/figures/Figure.java
```diff
@@ -0,0 +1,27 @@
...
+    @Override
+    public void drawTheFigure() {
+
+    }
+
+    @Override
+    public double getArea() {
+        return 0;
+    }
+}
```
Ivan95kos, line 27:
```suggestion
public abstract class AbstractFigure implements AreaCalculator, Drawable {
    protected String color;

    public AbstractFigure(String color) {
        this.color = color;
    }
}
```

Ivan95kos (changes requested):
Good job overall!

src/main/java/core/basesyntax/ColorList.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+public enum ColorList {
```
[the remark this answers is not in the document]
YuliiaNisha: Sorry, I didn't really get... enum should be called only using 1 word? I thought it is like class, so I can call it like "FigureSupplier" for example. Thanks for helping :)

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,67 @@
...
+            case "RIGHT_TRIANGLE" : return getRandomRightTriangle();
+            default : return getRandomSquare();
+        }
+    }
+
+    private AbstractFigure getRandomIsoscelesTrapezoid() {
+        IsoscelesTrapezoid isoscelesTrapezoid = new IsoscelesTrapezoid(
+                colorSupplier.getRandomColor(),
+                getRandomSize(), getRandomSize(), getRandomSize());
+        return isoscelesTrapezoid;
```
okuzan, line 35: In situations like this, there is no need to store the object in a local variable if it will be immediately used. Instead, you can pass the expression directly without assigning it to a variable

okuzan (changes requested):
Good job! Read the comments to find out how to improve your code 👇

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,62 @@
...
+    public AbstractFigure getRandomFigure() {
+        int index = random.nextInt(Figure.values().length);
+        Figure figureName = Figure.values()[index];
+
+        switch (figureName) {
+            case CIRCLE: return getRandomCircle();
+            case ISOSCELES_TRAPEZOID: return getRandomIsoscelesTrapezoid();
+            case RECTANGLE: return getRandomRectangle();
+            case RIGHT_TRIANGLE: return getRandomRightTriangle();
+            default : return getRandomSquare();
```
okuzan, line 27: default means unexpected figure has been generated - it's advisable to throw IllegalStateException in this case

okuzan (approved):
Well done!

PR #1411, opened 2024-01-05
created figure randomizer

vfilimo:


src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,24 @@
...
+        return Math.PI * radius * radius;
+    }
+
+    @Override
+    public String drawing() {
+        return "Figure: " + className + ", "
+                + "area: " + getArea() + " sq. units, "
+                + "radius: " + radius + " units, "
+                + "color: " + getColor();
+    }
```
olekskov, line 23: There is no need to extract to class variable rather simple expression that is used in a single method for a single time
```suggestion
public class Circle extends Figure {
    private int radius;

    public Circle(int radius, String color) {
        super(color);
        this.radius = radius;
    }

    @Override
    public double getArea() {
        return Math.PI * radius * radius;
    }

    @Override
    public String drawing() {
        return "Figure: " + getClass().getSimpleName() + ", "
                + "area: " + getArea() + " sq. units, "
                + "radius: " + radius + " units, "
                + "color: " + getColor();
    }
```
Please recheck other classes for same issue

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
...
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private Random random = new Random();
+    private int inedex = random.nextInt(Color.values().length);
+
+    public String getRandomColor() {
+        return Color.values()[inedex].name();
+    }
```
olekskov, line 11: with current approach your index will remain unchanged but you need it to change for every time method is called
```suggestion
public class ColorSupplier {
    private Random random = new Random();

    public String getRandomColor() {
        int index = random.nextInt(Color.values().length);
        return Color.values()[index].name();
    }
```

src/main/java/core/basesyntax/Draw.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Draw {
+    public abstract String drawing();
+}
```
olekskov, line 5: Interface names are better to be nouns or adjectives
Methods names are better to be verbs
```suggestion
public interface Drawable {
    public abstract String draw();
}
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,34 @@
...
+    private Figure rectangle = new Rectangle(random.nextInt(MIN_SIZE, MAX_SIZE),
+            random.nextInt(MIN_SIZE, MAX_SIZE),
+            new ColorSupplier().getRandomColor());
+    private Figure[] figures = new Figure[] {rectangle, circle, rightTriangle,
+            isoscelesTrapezoid, square};
+
+    public Figure getRandomFigure() {
+        int randomIndex = random.nextInt(figures.length);
+        return figures[randomIndex];
+    }
```
olekskov, line 29: Please rework this solution.
Now you have 5 predefined figures. So if you need 100 random figures you'll get about 20 duplicates of each figure.
So it not really works like random.

During method execution you should firstly to generate random number to determine shape you need and then generate this shape with random properties
So if you create 100 random figures you anyway will have about 20 squares but all of them will be different in color and size

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,34 @@
...
+
+    public Figure getDefaultFigure() {
+        return new Circle(10, Color.WHITE.name());
```
olekskov, line 32: magic numbers should be constants with proper naming

src/main/java/core/basesyntax/GetArea.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax;
+
+public interface GetArea {
+    public abstract double getArea();
+}
+
```
olekskov, line 6: only 1 blank line is required in the end of file, not 2

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,15 @@
...
+    private String color;
+
+    public Figure(String color) {
+        this.color = color;
+    }
+
+    public String getColor() {
+        return color;
+    }
+
```
olekskov, line 14:
```suggestion
```

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+public abstract class Figure implements GetArea, Draw {
+
```
olekskov, line 4:
```suggestion
```

olekskov (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
vfilimo: I have corrected. but my solution also gives random color

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,51 @@
...
+                        new ColorSupplier().getRandomColor());
+                break;
+            default:
+                randomFigure = new Circle(DEFAULT_RADIUS, Color.WHITE.name());
+        }
+        return randomFigure;
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(DEFAULT_RADIUS, Color.WHITE.name());
```
Ivan95kos, line 49:
```suggestion
        return new Circle(DEFAULT_RADIUS, DEFAULT_COLOR);
```

src/main/java/core/basesyntax/Drawable.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Drawable {
+    public abstract String draw();
```
Ivan95kos, line 4: To 'draw' means to print out all information about a figure using System.out.println()
You should `print` inside the method `draw`
```suggestion
    public void draw();
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,51 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int MIN_SIZE = 1;
+    private static final int MAX_SIZE = 9;
+    private static final int DEFAULT_RADIUS = 10;
+    private Figure randomFigure;
```
Ivan95kos, line 9: useless variable
```suggestion
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,51 @@
...
+
+            case 4:
+                randomFigure = new Rectangle(random.nextInt(MIN_SIZE, MAX_SIZE),
+                        random.nextInt(MIN_SIZE, MAX_SIZE),
+                        new ColorSupplier().getRandomColor());
+                break;
+            default:
+                randomFigure = new Circle(DEFAULT_RADIUS, Color.WHITE.name());
+        }
+        return randomFigure;
```
Ivan95kos, line 45:
```suggestion
```

src/main/java/core/basesyntax/GetArea.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface GetArea {
+    public abstract double getArea();
```
Ivan95kos, line 4: Interface methods are by default abstract
```suggestion
    public double getArea();
```

src/main/java/core/basesyntax/ColorSupplier.java
okuzan: it gives random color for each time you start the program, not for each  time the method is called

src/main/java/core/basesyntax/Color.java
```diff
@@ -0,0 +1,12 @@
...
+
+public enum Color {
+    BLUE,
+    GREEN,
+    ORANGE,
+    RED,
+    VIOLET,
+    WHITE,
+    YELLOW
+
```
okuzan, line 11:
```suggestion
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,42 @@
...
+            case 3:
+                return new Circle(random.nextInt(MIN_SIZE, MAX_SIZE),
+                        new ColorSupplier().getRandomColor());
+            case 4:
+                return new Rectangle(random.nextInt(MIN_SIZE, MAX_SIZE),
+                        random.nextInt(MIN_SIZE, MAX_SIZE),
+                        new ColorSupplier().getRandomColor());
+            default:
+                return new Circle(DEFAULT_RADIUS, Color.WHITE.name());
+        }
```
okuzan, line 40: use existing colorSupplier
better yet, you can generate the color before the switch case and then use this local variable

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -0,0 +1,26 @@
+package core.basesyntax;
+
+public class Rectangle extends Figure {
+    private int size1;
+    private int size2;
```
okuzan, line 5: rename to avoid numbers

src/main/java/core/basesyntax/Square.java
```diff
@@ -0,0 +1,24 @@
...
+
+    @Override
+    public void draw() {
+        System.out.println("Figure: " + getClass().getSimpleName() + ", "
+                + "area: " + getArea() + " sq. units, "
+                + "side: " + size + " units, "
+                + "color: " + getColor());
+    }
+}
+
```
okuzan, line 24:
```suggestion
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,42 @@
...
+    private static final int DEFAULT_RADIUS = 10;
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private Random random = new Random();
+
+    public Figure getDefaultFigure() {
+        return new Circle(DEFAULT_RADIUS, DEFAULT_COLOR);
+    }
+
+    public Figure getRandomFigure(int randomIndex) {
+
```
okuzan, line 18: method should be self sufficient - remove argument, generate index in the method

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,18 @@
...
+        Figure[] figures = new Figure[11];
+        figures[0] = figureSupplier.getDefaultFigure();
+        Random random = new Random();
+        for (int i = 1; i < figures.length; i++) {
+            figures[i] = figureSupplier.getRandomFigure(random.nextInt(5));
+        }
+        for (Figure figure : figures) {
+            figure.draw();
+        }
+    }
```
okuzan, line 17: this needs to be redone
remove random
extract array length (11 in your case) to a private class constant
you need only one for loop here, inside of which - if/else or ternary operator. If index is less then half of array length - then random, else - default

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,42 @@
...
+        switch (randomIndex) {
+            case 0:
+                return new Square(random.nextInt(MIN_SIZE, MAX_SIZE),
```
okuzan, line 21: Let's move this to a separate private method 
random.nextInt(MIN_SIZE, MAX_SIZE)
Also, let's use MAX_SIZE + 1, naming can be confusing since bound is excluded
image

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,18 @@
...
+import java.util.Random;
+
+public class Main {
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] figures = new Figure[11];
+        figures[0] = figureSupplier.getDefaultFigure();
+        Random random = new Random();
+        for (int i = 1; i < figures.length; i++) {
+            figures[i] = figureSupplier.getRandomFigure(random.nextInt(5));
```
okuzan, line 12: you should depend on one constant, don't hardcode its half, if the change will need to be made it will be in two places, instead of one - it's error-prone

okuzan (changes requested):
Good job, some refactoring is due

olekskov (approved):
Good job

PR #1416, opened 2024-01-07
Hm jv oop advanced

vladshender:


src/main/java/core/basesyntax/Color.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public enum Color {
+    blue,
+    white,
+    yellow,
+    black,
+    green,
+    grey,
+    purple
```
okuzan, line 10: uppercase enum values

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public abstract class Figure implements AreaCalculable {
+    protected String color;
+}
```
okuzan, line 5: create a constructor for color, implement Drawable too

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int MAX_VALUES_OF_SIDE = 20;
+    private static final int MAX_RADIUS = 15;
+    private static final int UNIQUE_RADIUS = 10;
+
+    private ColorSupplier color = new ColorSupplier();
+    private Random random = new Random();
+    private Figure circleUnique = new Circle(UNIQUE_RADIUS, Color.white.name());
```
okuzan, line 12: maybe change unique to default

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,35 @@
...
+
+    private Figure[] figures = {
+            new Rectangle(randomNumber(), randomNumber(), color.getRandomColor()),
+            new Circle(random.nextInt(MAX_RADIUS), color.getRandomColor()),
```
okuzan, line 16: The problem with this approach is we will always have the same circle, rectangle etc for one program run, we want to have a different figure each time getRandomFigure method is called

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax;
+
+public class Main {
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] figuresList = new Figure[6];
+
+        for (int i = 0; i < figuresList.length; i++) {
+            if (i < 3) {
```
okuzan, line 9: depend on array length, don't hardcode, if you decide to have a bigger array, you will need to make changes in two places instead of one

okuzan (changes requested):
Good job overall!

okuzan (changes requested):
.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,52 @@
...
+
+public class FigureSupplier {
+    private static final int MAX_VALUES_OF_SIDE = 20;
+    private static final int MAX_RADIUS = 15;
+    private static final int UNIQUE_RADIUS = 10;
+    private static final int VALUES_OF_FIGURES = 5;
+
+    private ColorSupplier color = new ColorSupplier();
+    private Random random = new Random();
+    private Figure circleDefault = new Circle(UNIQUE_RADIUS, Color.WHITE.name());
```
kozhukhovsky, line 13: It’s better not to use it like that, but simply return a new figure with default parameters

kozhukhovsky (approved):

PR #1419, opened 2024-01-10
oop-advanced-done

margosalii:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private Random random = new Random();
+
+    public String getRandomColor() {
+        int index = random.nextInt(Color.values().length);
+        return Color.values()[index].toString();
```
olekskov, line 10: https://github.com/mate-academy/jv-oop-advanced/blob/master/checklist.md#use-name-for-getting-string-representation-of-enum-constants
```suggestion
        return Color.values()[index].name();
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,63 @@
...
+    private IsoscelesTrapezoid getRandomIsoscelesTrapezoid() {
+        return new IsoscelesTrapezoid(randomColor(),
+                getRandomSide(), getRandomSide(), getRandomSide());
+    }
+
+    private RightTriangle getRandomRightTriangle() {
+        return new RightTriangle(randomColor(), getRandomSide(), getRandomSide());
+    }
+
+    private String randomColor() {
```
olekskov, line 55: method names should contain verbs for flawless code reading
```suggestion
    private String getRandomColor() {

```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+public class Main {
+    public static void main(String[] args) {
+        Figure [] figures = new Figure[6];
+        FigureSupplier figureSupplier = new FigureSupplier();
+        figures[0] = figureSupplier.getRandomFigure();
+        figures[1] = figureSupplier.getRandomFigure();
+        figures[2] = figureSupplier.getRandomFigure();
+        figures[3] = figureSupplier.getDefaultFigure();
```
olekskov, line 10: Please use loop to fill array of figures

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,17 @@
...
+public class Main {
+    public static void main(String[] args) {
+        Figure [] figures = new Figure[6];
+        FigureSupplier figureSupplier = new FigureSupplier();
+        figures[0] = figureSupplier.getRandomFigure();
+        figures[1] = figureSupplier.getRandomFigure();
+        figures[2] = figureSupplier.getRandomFigure();
+        figures[3] = figureSupplier.getDefaultFigure();
+        figures[4] = new Rectangle("GREEN", 3,6);
+        figures[5] = new IsoscelesTrapezoid("ORANGE", 9, 5,4);
```
olekskov, line 12: According to task: The other half of the figures should have the same, default parameters.
You shouldn't put some predefined values to constructor, use figureSupplier.getDefaultFigure() for whole second part of the aarray

olekskov (changes requested):

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+public class Main {
+    public static void main(String[] args) {
+        Figure [] figures = new Figure[6];
+        FigureSupplier figureSupplier = new FigureSupplier();
+
+        for (int i = 0; i < figures.length; i++) {
+            if (i < 3) {
```
okuzan, line 9: depend on figures.length, this way, when you need to change the array length and keep the current logic, you will need to make one change, instead of two ( on line 5 and line 9)

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+public class Main {
+    public static void main(String[] args) {
+        Figure [] figures = new Figure[6];
```
okuzan, line 5:
```suggestion
        Figure[] figures = new Figure[6];
```
if you want you can also extract 6 to ARRAY_LENGTH constant

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,63 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+
```
okuzan, line 6: we don't need an empty line after class definition according to mate code style

okuzan (changes requested):

okuzan (approved):
Well done!

PR #1432, opened 2024-01-21
create figures with random properties

Svitlankaa:


src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,23 @@
...
+    public void setColor(String color) {
+        this.color = color;
+    }
+
+    @Override
+    public void drawable() {
+
+    }
+}
+
```
okuzan, line 23:
```suggestion
}
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,36 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    public static final int FIGURE_COUNT = 5;
+    public static final double MAX_VALUE_PROPERTIES = 50;
+    public static final String DEFAULT_COLOR = "white";
```
okuzan, line 8:
```suggestion
    public static final String DEFAULT_COLOR = Сolor.WHITE.name();
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,21 @@
...
+            if (i < FIRST_HALF_ARRAY) {
+                figures[i] = figureSupplier.getRandomFigure();
+            } else {
+                figures[i] = figureSupplier.getDefaultFigure();
+            }
+        }
+        for (Figure figure : figures) {
+            figure.drawable();
+        }
+    }
```
okuzan, line 20: remove second `for`, call `draw()` in the first for loop

src/main/java/core/basesyntax/IsoscelesTrapezoid.java
```diff
@@ -0,0 +1,40 @@
...
+        return ((getShorterSide() + getLongerSide()) / DIVIDER_TRAPEZOID) * getHeight();
+    }
+
+    @Override
+    public void drawable() {
+        System.out.println("Figure: Isosceles trapezoid, shorter side " + getShorterSide()
+                + " units, longer side " + getLongerSide() + " units + area: "
+                + calculateArea() + "sq. units, color " + getColor());
+    }
+
```
okuzan, line 39:
```suggestion
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,21 @@
+package core.basesyntax;
+
+public class Main {
+    public static final int ARRAY_LENGTH = 6;
+    public static final int FIRST_HALF_ARRAY = ARRAY_LENGTH / 2;
```
okuzan, line 5: private
second one doesn't need to be a constant, just use figures.length / 2;

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,36 @@
...
+    public static final double DEFAULT_RADIUS = 10;
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private Random random = new Random();
+    private Figure[] figures = new Figure [] {
+            new Circle(colorSupplier.getRandomColor(), random.nextDouble(MAX_VALUE_PROPERTIES)),
+            new Square(random.nextDouble(MAX_VALUE_PROPERTIES), colorSupplier.getRandomColor()),
+            new RightTriangle(random.nextDouble(MAX_VALUE_PROPERTIES),
```
okuzan, line 15: Interesting idea, but it won't give truly random figures. They are generated once per instance lifetime, we need to generate new figures per each method call, so we should move it there. Better to generate in each case of switch statement separately than keep them in array (all generated figures but one won't be used, it's not efficient)

okuzan (changes requested):
Good job overall

PR #1433, opened 2024-01-22
solved task jv-oop-advanced

DanilKi:


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,46 @@
...
+                return new RightTriangle(triFirstLeg, triSecondLeg, figureColor);
+            case 4:
+                int circleRadius = random.nextInt(10) + 1;
+                return new Circle(circleRadius, figureColor);
+            default:
+                int trapFirstBase = random.nextInt(10) + 11;
+                int trapSecondBase = random.nextInt(10) + 1;
+                int trapSide = random.nextInt(10) + 1;
+                return new IsoscelesTrapezoid(trapFirstBase, trapSecondBase,trapSide, figureColor);
+        }
```
olekskov, line 40: random.nextInt(10) + 1 is used multiple times
extract expression to separate method with proper naming (like getRandomSide() and use it instead of variables.

also extract your 10 to constant with proper naming

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,19 @@
...
+    public static void main(String[] args) {
+        Figure[] figures = new Figure[FIGURE_ARRAY_LENGTH];
+        FigureSupplier figureSupplier = new FigureSupplier();
+        for (int i = 0; i < figures.length; i++) {
+            figures[i] = i < figures.length / 2 ? figureSupplier.getRandomFigure()
+                    : figureSupplier.getDefaultFigure();
+        }
+        for (Figure figure : figures) {
+            figure.draw();
+        }
```
olekskov, line 17: No need in separate loop
```suggestion
        for (int i = 0; i < figures.length; i++) {
            figures[i] = i < figures.length / 2 ? figureSupplier.getRandomFigure()
                    : figureSupplier.getDefaultFigure();
            figures[i].draw();
        }
```

src/main/java/core/basesyntax/model/Circle.java
```diff
@@ -0,0 +1,24 @@
+package core.basesyntax.model;
+
+public class Circle extends Figure {
+    private static final double PI = 3.14159265359;
```
olekskov, line 4: you can use constant from Math class instead of creation new one

src/main/java/core/basesyntax/model/Circle.java
```diff
@@ -0,0 +1,24 @@
...
+        return PI * radius * radius;
+    }
+
+    @Override
+    public String toString() {
+        return "Figure: circle, area: "
+                + calculateArea() + " sq. units, "
+                + "radius: " + radius + " units, "
+                + "color: " + color + ".";
+    }
```
olekskov, line 23: contradicts task
behavior - we can obtain the area of any figure and are able to draw it. To 'draw' means to print out all information about a figure using System.out.println() **(you shouldn't override the toString() method for this)**.

Please implement draw() method instead

src/main/java/core/basesyntax/model/Figure.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax.model;
+
+public abstract class Figure implements Drawable, AreaCalculable {
+    protected String color;
+
+    @Override
+    public void draw() {
+        System.out.println(this);
+    }
```
olekskov, line 9: not required in abstract class, let children classes implement this method in their own way

olekskov (changes requested):

src/main/java/core/basesyntax/model/Circle.java
DanilKi: Sorry, I've read 'shouldn't' as 'should' ((

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,43 @@
...
+            case 4:
+                return new Circle(getRandomLength(), figureColor);
+            default:
+                return new IsoscelesTrapezoid(getRandomLength() + RANDOM_LENGTH_BOUND,
+                        getRandomLength(), getRandomLength(), figureColor);
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(10, Color.WHITE.name().toLowerCase());
```
kozhukhovsky, line 37: move 10 to the constant

kozhukhovsky (approved):

PR #1445, opened 2024-01-27
Homework solution

nalforov:
Added abstract claas Figure and classes that inherit from it. Implemented AreaCalculator and Drawable interfaces. Added FigureSupplier and ColorSupplier for creating random figures.

### DIFF
--- a/src/main/java/core/basesyntax/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/FigureSupplier.java
@@ -1,0 +1,40 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        int figureIndex = random.nextInt(FigureType.values().length);
+        FigureType randomFigure = FigureType.values()[figureIndex];
+        switch (randomFigure) {
+            case CIRCLE -> {
+                return new Circle(colorSupplier.getRandomColor(), random.nextInt(50));
+            }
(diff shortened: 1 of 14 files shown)


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,40 @@
...
+                        random.nextInt(50), random.nextInt(50));
+            }
+            default -> {
+                return getDefaultFigure();
+            }
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(Color.WHITE.name(), 10);
```
Ivan95kos, line 38:
```suggestion
        return new Circle(DEFAULT_COLOR, DEFAULT_RADIUS);
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,40 @@
...
+                return new Rectangle(colorSupplier.getRandomColor(), random.nextInt(50),
+                        random.nextInt(50));
+            }
+            case RIGHT_TRIANGLE -> {
+                return new RightTriangle(colorSupplier.getRandomColor(), random.nextInt(50),
+                        random.nextInt(50));
+            }
+            case ISOSCELES_TRAPEZOID -> {
+                return new IsoscelesTrapezoid(colorSupplier.getRandomColor(), random.nextInt(50),
+                        random.nextInt(50), random.nextInt(50));
```
Ivan95kos, line 29: It is better to use constants instead of numbers

Ivan95kos (approved):
Great job with minor comments

PR #1451, opened 2024-02-02
Home task 6 without statick

VovaZSU:


VovaZSU (commented):
make new branch without static methods

PR #1452, opened 2024-02-03
solve the task

esidarapp:


esidarapp:
I dont know what to do with this error :
StructureTest.init:27 You should not rename base core.basesyntax package and path to project and project name should not contain spaces or words in cyrillic
I didn`t rename anything

olekskov (changes requested):
Test is failing with message
java.lang.AssertionError: Class Circle should have some parent class with common state for all figures - color

In general structure should be following:
You have abstract parent class figure that contains common field `color`.
It also implements 2 interfaces, one for area calculation behaviour, second for ability to be drawed

Regarding your question:
Please check that path to project doesn't contain any RU/UA symbols in naming.
e.g. C:\users\Євген\ideaProjects or anything similar

esidarapp:
I created new pull request for this task.
https://github.com/mate-academy/jv-oop-advanced/pull/1453

okuzan (changes requested):
Only changes to this PR will be tracked and reviewed from now on

PR #1455, opened 2024-02-04
All classes were added, all methods were implemented by the task.

slizko1:


src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax;
```
andrii-hoienko, line 1: It would be great if you will create packages and put figures in 'figure' package, suppliers to 'supplier'

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,22 @@
...
+
+public class Circle extends Figure {
+    private String color;
```
andrii-hoienko, line 4: Field color is common for all figure, it makes sense to move this variable to Figure class, also type of this variable should be Color since you already have this enum

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,22 @@
...
+
+    @Override
+    public void draw() {
+        System.out.println("Circle has " + color + " and radius " + radius
+                + ". Area is " + getArea());
+    }
+
+    @Override
+    public int getArea() {
+        return (int) (Math.PI * Math.pow(radius, 2));
```
andrii-hoienko, line 20: When you cast result of this expression to int value can be lost, so maybe it will be better to make areaCalculator return double

src/main/java/core/basesyntax/Drawable.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax;
+
+public interface Drawable {
+    void draw();
+
```
andrii-hoienko, line 5: Dont make unnecessary line breakes

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public abstract class Figure implements Drawable {
+    private String color;
+
+    public abstract int getArea();
+
+    public abstract void draw();
```
andrii-hoienko, line 8: It is not necessary to declare this method here because it already exists in Drawable interface

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public abstract class Figure implements Drawable {
+    private String color;
+
+    public abstract int getArea();
```
andrii-hoienko, line 6: Please be consistent, if you want to define behavior (methods) in separate interfaces like you did in case of method draw and Drawable interface you should create an interface AreaCalculator with method getArea, or avoid approach to use interfaces and declare everything in abstract class. First approach is more flexible and better from my point of view

src/main/java/core/basesyntax/FigureSuplier.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.util.Random;
+
+public class FigureSuplier {
+    public static final int FIGURE_COUNT = 4;
+    private Random random = new Random();
+    private ColorSupplier colors = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+
```
andrii-hoienko, line 11: Unnecessary line break

src/main/java/core/basesyntax/FigureSuplier.java
```diff
@@ -0,0 +1,43 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSuplier {
+    public static final int FIGURE_COUNT = 4;
+    private Random random = new Random();
+    private ColorSupplier colors = new ColorSupplier();
+
+    public Figure getRandomFigure() {
```
andrii-hoienko, line 10: This constructor exists by default

src/main/java/core/basesyntax/FigureSuplier.java
```diff
@@ -0,0 +1,43 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSuplier {
+    public static final int FIGURE_COUNT = 4;
+    private Random random = new Random();
+    private ColorSupplier colors = new ColorSupplier();
```
andrii-hoienko, line 8: bad variable name colors for a developer means list of objects with type Color, but it is not true in your case

src/main/java/core/basesyntax/FigureSuplier.java
```diff
@@ -0,0 +1,43 @@
...
+
+    public Figure getRandomFigure() {
+
+        int figureNumber = random.nextInt(FIGURE_COUNT);
+        String color = colors.getRandomColor();
+        switch (figureNumber) {
+            case 0:
+                int side = random.nextInt(15) + 1;
+                return new Square(color, side);
```
andrii-hoienko, line 17: To avoid using return in each case you can define variable on method level and return in only once in the end of this metod

src/main/java/core/basesyntax/FigureSuplier.java
```diff
@@ -0,0 +1,43 @@
...
+            case 1:
+                int horizontalSide = random.nextInt(15) + 1;
+                int verticalSide = random.nextInt(15) + 1;
+                return new Rectangle(color,horizontalSide,verticalSide);
+            case 2:
+                int base = random.nextInt(15) + 1;
```
andrii-hoienko, line 23: Code duplication, magic numbers

src/main/java/core/basesyntax/FigureSuplier.java
```diff
@@ -0,0 +1,43 @@
...
+                return new IsoscelesTrapezoid(color,base, height, upperSide);
+            case 3:
+                int firstLeg = random.nextInt(15) + 1;
+                int secondLeg = random.nextInt(15) + 1;
+                return new RightTriangle(color, firstLeg, secondLeg);
+            case 4:
+                int radius = random.nextInt(15) + 1;
+                return new Circle(color,radius);
+            default:
+                break;
```
andrii-hoienko, line 35: In default case it is better to return default figure

src/main/java/core/basesyntax/FigureSuplier.java
```diff
@@ -0,0 +1,43 @@
...
+        }
+        return getDefaultFigure();
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle("white", 10);
```
andrii-hoienko, line 41: Magic numbers

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+public class Main {
+    public static void main(String[] args) {
+        FigureSuplier figureSuplier = new FigureSuplier();
+        Figure[] figures = new Figure[6];
```
andrii-hoienko, line 6: Magic number

src/main/java/core/basesyntax/FigureSuplier.java
```diff
@@ -0,0 +1,43 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSuplier {
+    public static final int FIGURE_COUNT = 4;
```
andrii-hoienko, line 6: If you use this variable in only this class, please make it private

andrii-hoienko (changes requested):

andrii-hoienko (changes requested):
Good job!

andrii-hoienko (approved):
Good job!

PR #1456, opened 2024-02-05
Create simple figures logic with random values

TonyH277:


src/main/java/core/basesyntax/model/Figure.java
```diff
@@ -0,0 +1,23 @@
...
+    @Override
+    public void draw() {
+        System.out.println("Figure: "
+                + "name = " + name
+                + ", color = " + color);
+    }
+
+    @Override
+    public double area() {
+        return 0;
```
andrii-hoienko, line 21: We should not override it here, it could return some illegal value, for example, you create a new Figure and you forgot to override area(), and when area() is called on some instance 0 returned which in not valid value for figure you created

src/main/java/core/basesyntax/supplier/FigureSupplier.java
```diff
@@ -0,0 +1,56 @@
...
+        figures[4] = rectangle;
+        return figures[random.nextInt(figures.length)];
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle("Default Circle", colorSupplier.getWhiteColor(), 10);
+    }
+
+    private int randomValue() {
+        return random.nextInt(100);
```
andrii-hoienko, line 54: Magic number

src/main/java/core/basesyntax/supplier/FigureSupplier.java
```diff
@@ -0,0 +1,56 @@
...
+        figures[0] = square;
+        figures[1] = isoscelesTrapezoid;
+        figures[2] = rightTriangle;
+        figures[3] = circle;
+        figures[4] = rectangle;
+        return figures[random.nextInt(figures.length)];
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle("Default Circle", colorSupplier.getWhiteColor(), 10);
```
andrii-hoienko, line 50: Magic number

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax;
```
andrii-hoienko, line 1: It would be great if you will create packages and put figures in 'figure' package, suppliers to 'supplier' etc

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,22 @@
...
+        Figure[] figures = new Figure[]{
+                figureSupplier.getRandomFigure(),
+                figureSupplier.getRandomFigure(),
+                figureSupplier.getRandomFigure(),
+                figureSupplier.getDefaultFigure(),
+                figureSupplier.getDefaultFigure(),
+                figureSupplier.getDefaultFigure()
+        };
+        for (Figure figure : figures) {
+            figure.draw();
```
andrii-hoienko, line 18: Is it possible to add area to draw(); to void drawing it separately?

andrii-hoienko (changes requested):

andrii-hoienko (approved):
Good job!

PR #1469, opened 2024-02-11
Added the necessary classes and interfaces

AnyaRatusnyak:


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,38 @@
...
+    private static final int MAX_NUMBER = 10;
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        int figureNumber = random.nextInt(FIGURE_COUNT);
+        String figureColor = colorSupplier.getRandomColor();
+
+        switch (figureNumber) {
+            case 0:
+                return new Square(figureColor, random.nextInt(MAX_NUMBER) + 1);
```
andrii-hoienko, line 17: It is not necessary to write return in all cases, it is possible to create a variable on method-level and populate it in cases

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,38 @@
...
+            case 2:
+                return new RightTriangle(figureColor, random.nextInt(MAX_NUMBER) + 1,
+                        random.nextInt(MAX_NUMBER) + 1);
+            case 3:
+                return new Circle(figureColor, random.nextInt(MAX_NUMBER) + 1);
+            case 4:
+                return new IsoscelesTrapezoid(figureColor, random.nextInt(MAX_NUMBER) + 1,
+                        random.nextInt(MAX_NUMBER) + 1, random.nextInt(MAX_NUMBER) + 1);
+            default:
+                return new IsoscelesTrapezoid(figureColor, random.nextInt(MAX_NUMBER) + 1,
```
andrii-hoienko, line 30: Why in default case Isosceles Trapezoid returned?

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,38 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    public static final int FIGURE_COUNT = 5;
```
andrii-hoienko, line 6:
```suggestion
    private static final int FIGURE_COUNT = 5;

```

andrii-hoienko (changes requested):

src/main/java/core/basesyntax/FigureSupplier.java
AnyaRatusnyak: I think I can return any figure? Isn't it? I read it in Common mistakes (jv-oop-advanced). If you use a switch case construction in your solution, you may just put the last possible option in the default case.

andrii-hoienko (approved):
Good job!

PR #1477, opened 2024-02-13
Solution

VlasenkoYevhenii:


src/main/java/core/basesyntax/Application.java
```diff
@@ -0,0 +1,24 @@
...
+        for (int i = 0; i < FIGURES_COUNT; i++) {
+            if (i < FIGURES_COUNT / 2) {
+                figures[i] = figureSupplier.getRandomFigure();
+            } else {
+                figures[i] = figureSupplier.getDefaultFigure();
+            }
+        }
+
+        for (Figure figure : figures) {
+            figure.draw();
```
andrii-hoienko, line 21: No need to iterate twice, you can draw in first cycle

src/main/java/core/basesyntax/AreaCalculator.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax;
+
+public interface AreaCalculator {
+
```
andrii-hoienko, line 4:
```suggestion

```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private final Random random = new Random();
+    private final Color[] colors = Color.values();
```
andrii-hoienko, line 7: No need to make it class level variable

src/main/java/core/basesyntax/Drawable.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax;
+
+public interface Drawable {
+
```
andrii-hoienko, line 4:
```suggestion

```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,50 @@
...
+import core.basesyntax.figures.Square;
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int MAX_FIGURES_TYPES = 5;
+    private static final int SIZE_LIMIT = 5;
+    private static final int DEFAULT_CIRCLE_RADIUS = 10;
+    private static final String DEFAULT_COLOR = "White";
+
+    private final ColorSupplier colorSupplier = new ColorSupplier();
```
andrii-hoienko, line 17: Why this field is finale but random is not?

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,50 @@
...
+    public Figure getDefaultFigure() {
+        return new Circle(DEFAULT_COLOR, DEFAULT_CIRCLE_RADIUS);
+    }
+
+    public Figure getRandomFigure() {
+        String randomColor = colorSupplier.getRandomColor();
+        int sideA = getRandomSize();
+        switch (random.nextInt(MAX_FIGURES_TYPES) + 1) {
+            case 1:
+                return new Circle(randomColor, sideA);
```
andrii-hoienko, line 29: It is not necessary to write return in all cases, it is possible to create a variable on method-level and populate it in cases

src/main/java/core/basesyntax/figures/Square.java
```diff
@@ -0,0 +1,23 @@
...
+        System.out.println("Figure: square, area: " + getArea() + " sq. units, "
+                + "side: " + side + " units, "
+                + "color: " + color);
+    }
+
+    @Override
+    public double getArea() {
+        return Math.pow(side, 2);
+    }
+
```
andrii-hoienko, line 22:
```suggestion

```

andrii-hoienko (changes requested):

src/main/java/core/basesyntax/Drawable.java
VlasenkoYevhenii: Methods declarations must be separated from the previous code with an empty line

src/main/java/core/basesyntax/AreaCalculator.java
VlasenkoYevhenii: Methods declarations must be separated from the previous code with an empty line

src/main/java/core/basesyntax/ColorSupplier.java
VlasenkoYevhenii: We'll get excessive object creation on each call to the getRandomColor() method if these variables move onto the method level

src/main/java/core/basesyntax/FigureSupplier.java
VlasenkoYevhenii: Valid point. But as it's not said that we need to change figures after creation, but just draw them, I made classes immutable and it will cost me to rewrite half of the code to make it. Should I ? =)

src/main/java/core/basesyntax/Application.java
andrii-hoienko: Yes there are such requirement, but I don`t see a conflict. As an example: requirement is to read some filed and make all letters in upper case, in your logic you should do 2 cycles:
1. Reading file
2. Iterating by lines to make them upper-case
But this could be done in one cycle and such an approach is significantly more efficient.

andrii-hoienko: Yes there are such requirement, but I don`t see a conflict. As an example: the requirement is to read some file and make all letters in upper case, in your logic you should do 2 cycles:
1. Reading file line by line
2. Iterating by lines to make them upper-case
But this could be done in one cycle and such an approach is significantly more efficient.
So please remove the second cycle and iterate over elements only once

src/main/java/core/basesyntax/FigureSupplier.java
andrii-hoienko: These statements are not connected, you can declare variable: Figure figure; And than populate it in cases:
figure = new Circle(randomColor, sideA); Where is the problem?

src/main/java/core/basesyntax/Drawable.java
andrii-hoienko: Actually nobody does that, I mean in abstract classes (even in real one) there are no sense to separate method decration from class declaration by an empty line. Here is the examples of one method classes ( https://www.tutorialspoint.com/class-declaration-with-one-method-in-java, https://stackoverflow.com/questions/3670271/design-question-java-class-with-single-method-ok ...) You can google more examples by yourself, I never saw such approach as you did on real code. So please remove line breake

src/main/java/core/basesyntax/figures/Circle.java
```diff
@@ -0,0 +1,22 @@
...
+
+    public Circle(String color, double radius) {
+        super(color);
+        this.radius = radius;
+    }
+
+    @Override
+    public void draw() {
+        System.out.println("Figure: circle, area: " + getArea() + " sq. units, "
+                                                    + "radius: " + radius + " units, "
```
andrii-hoienko, line 14: Please be consistent and make same tabulation in all draw methods

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,50 @@
...
+            case 5:
+                int trapezoidSideB = getRandomSize();
+                int trapezoidHeight = getRandomSize();
+                return new IsoscelesTrapezoid(randomColor, sideA, trapezoidSideB, trapezoidHeight);
+            default:
+                return getDefaultFigure();
+        }
+    }
+
+    private int getRandomSize() {
```
andrii-hoienko, line 47: Bad method name, often word 'Size' means amount of elements in some collection, but this is not true in this case (same about constant)

src/main/java/core/basesyntax/ColorSupplier.java
andrii-hoienko: Then be consistent and make this field constant and put there length of colors, not colors themself

andrii-hoienko (changes requested):

src/main/java/core/basesyntax/FigureSupplier.java
VlasenkoYevhenii: Confuseв it with field setting, my bad

src/main/java/core/basesyntax/Application.java
```diff
@@ -0,0 +1,22 @@
...
+    private static final int FIGURES_COUNT = 10;
+
+    public static void main(String[] args) {
+        final FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] figures = new Figure[FIGURES_COUNT];
+
+        for (int i = 0; i < FIGURES_COUNT; i++) {
+            if (i < FIGURES_COUNT / 2) {
+                figures[i] = figureSupplier.getRandomFigure();
+                figures[i].draw();
```
andrii-hoienko, line 15: No need to write figures[i].draw(); twice

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private static final Random RANDOM = new Random();
```
andrii-hoienko, line 6:
```suggestion
    private final Random random = new Random();

```

andrii-hoienko (changes requested):
Good job, left minor comments

andrii-hoienko (approved):
Good job!

PR #1481, opened 2024-02-14
Jv oop advanced hw

DmytroHadiuchko:


src/main/java/core/basesyntax/AreaCalculator.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
```
andrii-hoienko, line 1: It would be great if you will create packages and put figures in 'figure' package, suppliers to 'supplier'

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+
+    public String getRandomColor() {
+        Color[] colors = Color.values();
+        int index = new Random().nextInt(colors.length);
```
andrii-hoienko, line 9: Make new Random() class level variable

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,31 @@
...
+            case 0 :
+                return new Rectangle(colorSupplier.getRandomColor(), areaValue, areaValue);
+            case 1 :
+                return new Square(colorSupplier.getRandomColor(), areaValue);
+            case 2 :
+                return new RightTriangle(colorSupplier.getRandomColor(), areaValue, areaValue);
+            case 3 :
+                return new IsoscelesTrapezoid(colorSupplier.getRandomColor(), areaValue, areaValue);
+            default:
+                return new Circle(colorSupplier.getRandomColor(), areaValue);
```
andrii-hoienko, line 24: Why random circle returned in default case?

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,31 @@
...
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(Color.WHITE.name(), DEFAULT_RADIUS_FOR_CIRCLE);
```
andrii-hoienko, line 29: Color.WHITE.name() to constant

src/main/java/core/basesyntax/HelloWorld.java
```diff
@@ -4,5 +4,22 @@
  * Feel free to remove this class and create your own.
  */
 public class HelloWorld {
+    private static final int SIZE_OF_FIGURES_ARRAY = 6;
```
andrii-hoienko, line 7: Only one constant is enouth, because FIRST_PART_OF_ARRAY  = SIZE_OF_FIGURES_ARRAY  / 2

andrii-hoienko (changes requested):

src/main/java/core/basesyntax/enums/Color.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax.enums;
```
andrii-hoienko, line 1:
```suggestion
package core.basesyntax.enum;

```

src/main/java/core/basesyntax/supplier/FigureSupplier.java
```diff
@@ -0,0 +1,42 @@
...
+import core.basesyntax.figure.RightTriangle;
+import core.basesyntax.figure.Square;
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int NUMBERS_OF_FIGURES = 5;
+    private static final int DEFAULT_RADIUS_FOR_CIRCLE = 10;
+    private static final String DEFAULT_COLOR = Color.WHITE.name();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
```
andrii-hoienko, line 17:
```suggestion

```

src/main/java/core/basesyntax/supplier/FigureSupplier.java
```diff
@@ -0,0 +1,42 @@
...
+import core.basesyntax.figure.IsoscelesTrapezoid;
+import core.basesyntax.figure.Rectangle;
+import core.basesyntax.figure.RightTriangle;
+import core.basesyntax.figure.Square;
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int NUMBERS_OF_FIGURES = 5;
+    private static final int DEFAULT_RADIUS_FOR_CIRCLE = 10;
+    private static final String DEFAULT_COLOR = Color.WHITE.name();
```
andrii-hoienko, line 15:
```suggestion
    private static final String DEFAULT_COLOR = Color.WHITE.name();
    

```

src/main/java/core/basesyntax/supplier/FigureSupplier.java
```diff
@@ -0,0 +1,42 @@
...
+            case 1 :
+                return new Square(colorSupplier.getRandomColor(), areaValue);
+            case 2 :
+                return new RightTriangle(colorSupplier.getRandomColor(), areaValue, areaValue);
+            case 3 :
+                return new IsoscelesTrapezoid(colorSupplier.getRandomColor(), areaValue, areaValue);
+            case 4:
+                return new Circle(colorSupplier.getRandomColor(), areaValue);
+            default:
+                return null;
```
andrii-hoienko, line 35: Why null returned in default case? It is better to return default figure or throw exception

src/main/java/core/basesyntax/HelloWorld.java
```diff
@@ -1,8 +1,27 @@
 package core.basesyntax;
 
+import core.basesyntax.figure.Figure;
+import core.basesyntax.supplier.FigureSupplier;
+
 /**
  * Feel free to remove this class and create your own.
```
andrii-hoienko, line 7: REmove unnecessary comments

src/main/java/core/basesyntax/HelloWorld.java
```diff
@@ -1,8 +1,27 @@
...
+        Figure[] figures = new Figure[SIZE_OF_FIGURES_ARRAY];
 
+        for (int i = 0; i < SIZE_OF_FIGURES_ARRAY; i++) {
+            if (i < (SIZE_OF_FIGURES_ARRAY / 2)) {
+                figures[i] = figureSupplier.getRandomFigure();
+            } else {
+                figures[i] = figureSupplier.getDefaultFigure();
+            }
+        }
+        for (Figure figure : figures) {
```
andrii-hoienko, line 23: No need to iterate again, draw method can be called in first cycle

andrii-hoienko (changes requested):
Good job, left minor comments

src/main/java/core/basesyntax/supplier/ColorSupplier.java
```diff
@@ -0,0 +1,16 @@
...
+import java.util.Random;
+
+public class ColorSupplier {
+    private final Random random = new Random();
+
+    public String getRandomColor() {
+        Color[] colors = Color.values();
+        int index = random.nextInt(colors.length);
+        Color color = colors[index];
+        return color.name();
```
karma-o, line 13: `color` var use only once, thus can be removed
 ```suggestion
         return colors[index].name();
```

File: src/main/java/core/basesyntax/supplier/ColorSupplier.java [hunk 13]
```diff
@@ -0,0 +1,16 @@
...
+    }
+
```
karma-o, line 15:
```suggestion
```

src/main/java/core/basesyntax/supplier/FigureSupplier.java
```diff
@@ -0,0 +1,41 @@
...
+public class FigureSupplier {
+    private static final int NUMBERS_OF_FIGURES = 5;
+    private static final int DEFAULT_RADIUS_FOR_CIRCLE = 10;
+    private static final String DEFAULT_COLOR = Color.WHITE.name();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final Random random = new Random();
+
+    public Figure getRandomFigure() {
+        int type = random.nextInt(NUMBERS_OF_FIGURES);
+        double areaValue = random.nextDouble();
```
karma-o, line 21: first of all - naming, this is not an area yet, it is just a value for one of the figure parameters,
second of all, with that approach every figure would always have the same parameters that are units measured, for example, a rectangle would have both sides the same, so it would now be also always a square

Also, we can encapsulate it a bit more and add a method like `getRandomDouble()` to call it instead of direct random call to avoid code duplication, that would also be a great example of the `random` re-usage in multiple methods

src/main/java/core/basesyntax/supplier/FigureSupplier.java
```diff
@@ -0,0 +1,41 @@
...
+                return new Square(colorSupplier.getRandomColor(), areaValue);
+            case 2 :
+                return new RightTriangle(colorSupplier.getRandomColor(), areaValue, areaValue);
+            case 3 :
+                return new IsoscelesTrapezoid(colorSupplier.getRandomColor(), areaValue, areaValue);
+            case 4:
+                return new Circle(colorSupplier.getRandomColor(), areaValue);
+            default:
+                return getDefaultFigure();
+        }
```
karma-o, line 35: we can simplify it like this, more about it here: https://www.baeldung.com/java-switch#1-the-new-switch-expression
```suggestion
        return switch (type) {
            case 0 -> new Rectangle(colorSupplier.getRandomColor(), areaValue, areaValue);
            case 1 -> new Square(colorSupplier.getRandomColor(), areaValue);
            case 2 -> new RightTriangle(colorSupplier.getRandomColor(), areaValue, areaValue);
            case 3 -> new IsoscelesTrapezoid(colorSupplier.getRandomColor(), areaValue, areaValue);
            case 4 -> new Circle(colorSupplier.getRandomColor(), areaValue);
            default -> getDefaultFigure();
        }
```

src/main/java/core/basesyntax/supplier/FigureSupplier.java
```diff
@@ -0,0 +1,41 @@
...
+import core.basesyntax.figure.RightTriangle;
+import core.basesyntax.figure.Square;
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int NUMBERS_OF_FIGURES = 5;
+    private static final int DEFAULT_RADIUS_FOR_CIRCLE = 10;
+    private static final String DEFAULT_COLOR = Color.WHITE.name();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final Random random = new Random();
```
karma-o, line 17: static vars are usually separated from the usual with a line break
```suggestion
    private static final int NUMBERS_OF_FIGURES = 5;
    private static final int DEFAULT_RADIUS_FOR_CIRCLE = 10;
    private static final String DEFAULT_COLOR = Color.WHITE.name();
    
    private final ColorSupplier colorSupplier = new ColorSupplier();
    private final Random random = new Random();

```

karma-o (changes requested):

PR #1491, opened 2024-02-19
Main

slavik33:


src/main/java/core/basesyntax/HelloWorld.java
```diff
@@ -1,8 +1,23 @@
...
+            if (i < figures.length / 2) {
+                figures[i] = figureSupplier.getRandomFigure();
+            } else {
+                figures[i] = figureSupplier.getDefaultFigure();
+            }
+        }
+
+        for (Figure figure : figures) {
+            figure.draw();
+        }
```
okuzan, line 20:
```suggestion
            }
            figures[i].draw();
        }
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,46 @@
...
+                return new Rectangle(colorSupplier.getRandomColor(), random.nextDouble(RANGE_VALUE),
+                        random.nextDouble(RANGE_VALUE));
+            }
+            case RIGHT_TRIANGLE -> {
+                return new RightTriangle(colorSupplier.getRandomColor(),
+                        random.nextDouble(RANGE_VALUE),
+                        random.nextDouble(RANGE_VALUE));
+            }
+            default -> {
+                return getDefaultFigure();
```
okuzan, line 38:
```suggestion
                throw new IllegalStateException("Can't recognize this figure type: " 
                    + randomFigure.name());
```
IMO it's an unexpected situation - better make it visible, rather than return the default figure silently

okuzan (changes requested):
Clean code, good job! 🚀

slavik33:
Виправив

okuzan (approved):
Superb 🔥

PR #1503, opened 2024-02-29
Jv oop advanced solution

Oleh-Voitykhivskyi:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private Random random = new Random();
+
+    public String getRandomColor() {
+        int index = random.nextInt(Color.values().length);
+        return String.valueOf(Color.values()[index]);
```
kozhukhovsky, line 10:
```suggestion
        return Color.values()[index].name();
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,42 @@
...
+                return new Square(COLOR, random.nextInt(BOUND));
+            }
+            default -> {
+                return null;
+            }
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle("White", 10);
```
kozhukhovsky, line 40: use Color enum and move 10 to the constant

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,21 @@
+package core.basesyntax;
+
+public class Figure implements Drawable {
```
kozhukhovsky, line 3: where is method for getting area?

kozhukhovsky (changes requested):

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,22 @@
...
+    public String toString() {
+        return super.toString() + "radius: " + radius
+                + ", area: " + getArea();
+    }
+
+    @Override
+    public double getArea() {
+        return Math.PI * radius * radius;
+    }
+
```
okuzan, line 21:
```suggestion
```

src/main/java/core/basesyntax/Figure.java
okuzan: create an interface for this - AreaCalculator, similar to Drawable

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    @Override
+    public void draw(Figure figure) {
+        System.out.println(figure);
+    }
+
+    public double getArea() {
+        return 0;
+    }
```
okuzan, line 24:
```suggestion
```
these methods should be abstract, we let subclasses implement them - we can't know all their details, radius of circle for example, in this abstract superclass

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,44 @@
...
+                return new Square(COLOR, random.nextInt(BOUND));
+            }
+            default -> {
+                return null;
+            }
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(Color.values()[4].name(), DEFAULT_RADIUS);
```
okuzan, line 42:
```suggestion
        return new Circle(DEFAULT_COLOR, DEFAULT_RADIUS);
```
Where DEFAULT_COLOR is Color.WHITE.name()

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,44 @@
...
+            }
+            case RIGHT_TRIANGLE -> {
+                return new RightTriangle(COLOR, random.nextInt(BOUND), random.nextInt(BOUND),
+                        random.nextInt(BOUND));
+            }
+            case SQUARE -> {
+                return new Square(COLOR, random.nextInt(BOUND));
+            }
+            default -> {
+                return null;
```
okuzan, line 36: let's throw IllegalStateExceptoin + figureType -> this shouldn't occur, we need to fail fast. Another option - replace square clause with default

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,44 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int BOUND = 600;
+    private static final int DEFAULT_RADIUS = 10;
+    private static ColorSupplier colorSupplier = new ColorSupplier();
+    private static final String COLOR = colorSupplier.getRandomColor();
```
okuzan, line 9: you don't generate a new color for each call of the method, this constant is set once upon class creation, and doesn't change in runtime

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,44 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int BOUND = 600;
+    private static final int DEFAULT_RADIUS = 10;
+    private static ColorSupplier colorSupplier = new ColorSupplier();
```
okuzan, line 8:
```suggestion
    private ColorSupplier colorSupplier = new ColorSupplier();
```
avoid static where it's not needed, static members are harder to mock and test

src/main/java/core/basesyntax/HelloWorld.java
```diff
@@ -4,5 +4,21 @@
...
+            if (i < 3) {
+                figures[i] = figureSupplier.getRandomFigure();
+            } else {
+                figures[i] = figureSupplier.getDefaultFigure();
+            }
+        }
+
+        for (Figure figure : figures) {
+            System.out.println(figure);
+        }
```
okuzan, line 22: you don't need second if, print in the first one

okuzan (changes requested):
Good job overall

okuzan (approved):

PR #1507, opened 2024-03-05
Created an array with random and non-random figures

ovodolazov:


src/main/java/core/basesyntax/DrawFigure.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface DrawFigure {
```
Rommelua, line 3: Think about why using DrawFigure name for this interface is a mistake. (You can ask in the chat in case of have no ideas)
```suggestion
public interface Drawable {
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,39 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final Random random = new Random();
+
+    public Figure getRandomFigure() {
+        int indexFigure = random.nextInt(5);
```
Rommelua, line 10: image

Rommelua (changes requested):
Before you create PR with your homework, you need to go through the checklist under the task and correct all the points described there.

Vasul-Pankiv (approved):

PR #1509, opened 2024-03-06
MateAcademy/Task/jv-oop-advanced - Bartosz Wójcik solution

banderan:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    public String getRandomColor() {
+        Random random = new Random();
```
nowickimj, line 7: `Random` should be a constant.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,42 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+
+    public Figure getRandomFigure() {
+        Random random = new Random();
```
nowickimj, line 8: Should be a constant.

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+public abstract class Figure implements Drawable {
+    private String color;
```
nowickimj, line 4: You've created an enum that represents colors - why not use it here?

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,42 @@
...
+        //adding bound: 1000 to avoid wrong area and to leave clarity code
+        ColorSupplier colorSupplier = new ColorSupplier();
+        Figure[] figures = {
+                new IsoscelesTrapezoid(
+                        colorSupplier.getRandomColor(),//Color
+                        random.nextInt(1000),//upperSide
```
nowickimj, line 14: Make max value a constant and use it in every case below.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,42 @@
...
+                ),
+                new Square(
+                        colorSupplier.getRandomColor(),//color
+                        random.nextInt(1000)//side
+                )};
+        return figures[random.nextInt(figures.length)];
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(Colors.WHITE.toString(), 10);
```
nowickimj, line 40: Default radius should be a constant.

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax;
+
+public class Main {
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
```
nowickimj, line 5: This could be a constant.

nowickimj (changes requested):

banderan:
code has been fixed

src/main/java/core/basesyntax/GetArea.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface GetArea {
```
nowickimj, line 3: This is not a proper name for an interface - make sure that you've read the guidelines before posting a solution: https://github.com/mate-academy/jv-oop-advanced/blob/master/checklist.md
obraz

src/main/java/core/basesyntax/Square.java
```diff
@@ -0,0 +1,23 @@
...
+    private int side;
+
+    public Square(Colors color, int side) {
+        super(color);
+        this.side = side;
+    }
+
+    @Override
+    public double getArea() {
+
```
nowickimj, line 13: unnecessary new line

nowickimj (changes requested):
2 minor changes required

src/main/java/core/basesyntax/Drawable.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Drawable {
+    Figure draw();
```
```suggestion
    void draw();
```
why make it return something and then return null? if it's not meant to return anything, make it void instead

src/main/java/core/basesyntax/Square.java
```diff
@@ -0,0 +1,22 @@
...
+    @Override
+    public double getArea() {
+        return side * side;
+    }
+
+    @Override
+    public Figure draw() {
+        System.out.println("Figure: Square, area: " + getArea() + " sq. units, side: " + side
+                + " units, color: " + getColor());
+        return null;
```
okuzan, line 20:
```suggestion
```
return null is usually a bad practice

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,16 @@
...
+        final Figure[] figures = new Figure[6];
+
+        for (int i = 0; i < figures.length; i++) {
+            if (i < 3) {
+                figures[i] = figureSupplier.getRandomFigure();
+            } else {
+                figures[i] = figureSupplier.getDefaultFigure();
+            }
+        }
+    }
```
okuzan, line 15: inside of this `for`, print the value that you generate, using `draw()` 
image

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public enum Colors {
```
okuzan, line 3:
```suggestion
public enum Color {
```
enum should be in singular
https://checkstyle.sourceforge.io/styleguides/google-java-style-20180523/javaguide.html#s4.8.1-enum-classes

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,44 @@
...
+                ),
+                new Square(
+                        colorSupplier.getRandomColor(),//color
+                        random.nextInt(MAX_VALUE)//side
+                )};
+        return figures[random.nextInt(figures.length)].draw();
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(Colors.WHITE, DEFAULT_RADIUS).draw();
```
okuzan, line 42:
```suggestion
        return new Circle(DEFAULT_COLOR, DEFAULT_RADIUS);
```
let's make a constant for this too
no need to call draw here

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+public class Main {
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        final Figure[] figures = new Figure[6];
+
+        for (int i = 0; i < figures.length; i++) {
+            if (i < 3) {
```
okuzan, line 9: don't hardcode 3, depend on array's length dynamically

okuzan (changes requested):
Here's how to make your code better 👇

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,45 @@
...
+                new Rectangle(
+                        colorSupplier.getRandomColor(),//color
+                        random.nextInt(MAX_VALUE),//sideA
+                        random.nextInt(MAX_VALUE)//sideB
+                ),
+                new Square(
+                        colorSupplier.getRandomColor(),//color
+                        random.nextInt(MAX_VALUE)//side
+                )};
+        return figures[random.nextInt(figures.length)];
```
nowickimj, line 39: In general, you should avoid the initialization of data that you won't use. In your case here, you're creating the figures ahead but you're only using one. Please replace this with a proper `switch` statement that will only create an object when needed. The rest looks good :)

nowickimj (changes requested):
small remark, besides that looks great ;)

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,21 @@
...
+
+    @Override
+    public double getArea() {
+        return Math.PI * radius * radius;
+    }
+
+    @Override
+    public void draw(Figure figure) {
+        System.out.println("Figure: Circle, area: " + getArea() + " sq. units, radius: "
+                + radius + " units, color: " + getColor());
```
nowickimj, line 19: `color` is a protected field so you can call it immediately, without using the getter

nowickimj (approved):

PR #1529, opened 2024-03-22
Jv oop advanced

citruscaviar:
applied suggested changes for magic numbers

### DIFF
--- a/src/main/java/core/basesyntax/AreaCalculator.java
+++ b/src/main/java/core/basesyntax/AreaCalculator.java
@@ -1,0 +1,5 @@
+package core.basesyntax;
+
+public interface AreaCalculator {
+    double getArea();
+}
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private final Color[] colors = Color.values();
+    private final Random random = new Random();
+
+    public String getRandomColor() {
+        int randomColorIndex = random.nextInt(colors.length);
+        return colors[randomColorIndex].name();
+    }
+}
--- a/src/main/java/core/basesyntax/Drawable.java
+++ b/src/main/java/core/basesyntax/Drawable.java
@@ -1,0 +1,6 @@
+package core.basesyntax;
+
+public interface Drawable {
+    void draw();
+
+}
--- a/src/main/java/core/basesyntax/Figure.java
+++ b/src/main/java/core/basesyntax/Figure.java
@@ -8,7 +8,6 @@ public class Figure implements Drawable, AreaCalculator {
     }
 
     public Figure() {
-
     }
 
     public String getColor() {
@@ -17,7 +16,6 @@ public class Figure implements Drawable, AreaCalculator {
 
     @Override
     public void draw() {
-
     }
 
     @Override
--- a/src/main/java/core/basesyntax/Figure.java
+++ b/src/main/java/core/basesyntax/Figure.java
@@ -1,6 +1,8 @@
 package core.basesyntax;
 
 public class Figure implements Drawable, AreaCalculator {
+    protected static final double ROUNDING_PRECISION = 10.0;
+    protected static final int DEFAULT_LENGTH = 10;
     private String color;
 
     public Figure(String color) {
--- a/src/main/java/core/basesyntax/Figure.java
+++ b/src/main/java/core/basesyntax/Figure.java
@@ -1,0 +1,27 @@
+package core.basesyntax;
+
+public class Figure implements Drawable, AreaCalculator {
+    private String color;
+
+    public Figure(String color) {
+        this.color = color;
+    }
+
+    public Figure() {
+
+    }
+
+    public String getColor() {
+        return color;
+    }
+
+    @Override
+    public void draw() {
+
+    }
+
+    @Override
+    public double getArea() {
+        return 0;
+    }
+}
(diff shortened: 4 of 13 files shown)


src/main/java/core/basesyntax/Drawable.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax;
+
+public interface Drawable {
+    void draw();
+
```
MAMentorFX, line 5:
```suggestion
```

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,27 @@
+package core.basesyntax;
+
+public class Figure implements Drawable, AreaCalculator {
+    protected static final double ROUNDING_PRECISION = 10.0;
```
MAMentorFX, line 4: Why do we need this rounding precision? If so, it feels like it's better to move it AreaCalculator

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,27 @@
+package core.basesyntax;
+
+public class Figure implements Drawable, AreaCalculator {
```
MAMentorFX, line 3: This task doesn't require creating a new Figure. It challenges your knowledge of working with abstraction.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,46 @@
...
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        int figureType = random.nextInt(FIGURE_COUNT);
+        String color = colorSupplier.getRandomColor();
+        int randomNumber = random.nextInt(DEFAULT_LENGTH) + 1;
+        switch (figureType) {
+            case 0 -> {
+                return new Circle(color,randomNumber);
+            }
```
MAMentorFX, line 18:
```suggestion
        return switch (figureType) {
            case 0 -> new Circle(color,randomNumber);
```
Make it beautiful :)

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,46 @@
...
+public class FigureSupplier extends Figure {
+    public static final int FIGURE_COUNT = 5;
+    private static final Random random = new Random();
+    private static final Color WHITE = Color.WHITE;
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        int figureType = random.nextInt(FIGURE_COUNT);
+        String color = colorSupplier.getRandomColor();
+        int randomNumber = random.nextInt(DEFAULT_LENGTH) + 1;
```
MAMentorFX, line 14: It's confusing why we pass Default Length here. Create a separate constant (also you won't need + 1 then)

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,15 @@
...
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] arrayOfFigures = new Figure[6];
+        for (int i = 0; i < arrayOfFigures.length; i++) {
+            arrayOfFigures[i] = (i < 3) ? figureSupplier.getRandomFigure()
+                    : figureSupplier.getDefaultFigure();
+        }
+        for (Figure figure : arrayOfFigures) {
+            figure.draw();
+        }
```
MAMentorFX, line 13: You can do it with one for-loop. Imagine 100.000 figures. And now instead of 100.000 iterations, you have 200.000

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,46 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier extends Figure {
```
MAMentorFX, line 5: FigureSuplier is just a class for business logic purposes, we use it to get some figures. Why does it extend Figure? FigureSupplier doesn't want to have Figure state and functionality. Now FigureSupplier inherited color and a couple of methods

MAMentorFX (changes requested):
good job, just needs a little polish

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private final Color[] colors = Color.values();
```
MAMentorFX, line 6: Usually, class-level variables are some services or fields that are part of a bigger entity, like engine is a part of a car, or an email is a detail of some user, but an array like this is more of a part of a method and is not worth to be stored in a class, even if it is a constant value like in this case

src/main/java/core/basesyntax/AreaCalculator.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax;
+
+public interface AreaCalculator {
+    int DEFAULT_LENGTH = 10;
```
MAMentorFX, line 4: This constant is used as a value of created instances. So it's just a constant of FigureSupplier

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,17 @@
...
+        this.color = color;
+    }
+
+    public Figure() {
+    }
+
+    public String getColor() {
+        return color;
+    }
+
```
MAMentorFX, line 16:
```suggestion
```

src/main/java/core/basesyntax/FigureSupplier.java
MAMentorFX: Still not fixed. Is unclear wehy exactly DEfault length is in nextint, You better clarify by creating new constant with proper meaning

src/main/java/core/basesyntax/Square.java
```diff
@@ -0,0 +1,25 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Square extends Figure {
+    private final int side;
+    private final Random random = new Random();
```
MAMentorFX, line 7: redundant field.

MAMentorFX (approved):
Good job. Everytime you create a new class think its purpose

PR #1531, opened 2024-03-25
implement figures

artyomwhite:


src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,25 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+    private static final double DEFAULT_DECIMAL_RANK = 10.0;
```
MentorMA, line 4: Why do you need this?

src/main/java/core/basesyntax/IsoscelesTrapezoid.java
```diff
@@ -0,0 +1,33 @@
+package core.basesyntax;
+
+public class IsoscelesTrapezoid extends RightTriangle {
+    private static final double DEFAULT_DECIMAL_RANK = 10.0;
```
MentorMA, line 4: You don`t need this

MentorMA (changes requested):

PR #1534, opened 2024-03-25
Initial solution

MarcusPy:


src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,21 @@
+package core.basesyntax;
+
+public class Circle extends Figure {
+    private final double radius;
+    private final String color;
```
nowickimj, line 5: Please note that every figure has a color - why not put it in the base class then?

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax;
+
+public abstract class Figure implements AreaCalculator {
+
+    abstract void draw();
```
nowickimj, line 5: This should be provided by an interface

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax;
+
+public abstract class Figure implements AreaCalculator {
+
```
nowickimj, line 4: Unnecessary new line - please keep the code consistent :)

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,21 @@
...
+        this.color = color;
+    }
+
+    @Override
+    public double calculateArea() {
+        return Math.PI * radius * radius;
+    }
+
+    public void draw() {
+        System.out.printf("Figure: %s, Area: %f sq. units, Radius: %f units, Color: %s\n",
```
nowickimj, line 18: `println()` will make the output much clearer as each entry will be printed in separate line

nowickimj (changes requested):
Few remarks

src/main/java/core/basesyntax/Circle.java
MarcusPy: I stuck with printf, but used a multiline string to achieve the same effect and improve readability of both code and output

src/main/java/core/basesyntax/DrawPrinter.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface DrawPrinter {
```
nowickimj, line 3: Hint for naming interfaces in Java: https://www.oreilly.com/library/view/java-pocket-guide/9781491938683/ch01.html

"Interface names should be adjectives. They should end with “able” or “ible” whenever the interface provides a capability; otherwise, they should be nouns."

How about `Drawable` then?

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,29 @@
...
+            case 3 -> new Circle(random.nextDouble(MAX_SIZE), colorSupplier.getRandomColor());
+            case 4 -> new IsoscelesTrapezoid(
+                    random.nextDouble(MAX_SIZE), random.nextDouble(MAX_SIZE),
+                    random.nextDouble(MAX_SIZE), colorSupplier.getRandomColor());
+            default -> getDefaultFigure();
+        };
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(10, Colors.WHITE.name());
```
nowickimj, line 27: You can extract default radius to constant.
If you make the `color` field immutable (make it `final`), you might as well extract the whole default `Circle` object to constant.

nowickimj (changes requested):
Small changes

nowickimj (approved):

PR #1546, opened 2024-04-03
solution of task JV-OOP-advansed

Julianevinska:


src/main/java/core/basesyntax/DrawFigure.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax;
+
+public interface DrawFigure {
```
nowickimj, line 3: According to Java naming convention, interface name should be an adjective, may also be a noun, but not verb.

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,27 @@
+package core.basesyntax;
+
+public class Figure implements DrawFigure, AreaCalculator {
```
nowickimj, line 3: `Figure` should be an abstract class:
1. We shouldn't allow to create a `Figure` instance - we don't know which figure it is.
2. If you use an abstract class, you can delegate the implementation of interface methods to subclasses.

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    @Override
+    public double area() {
+        return 0;
+    }
+
+    @Override
+    public void draw() {
+    }
```
nowickimj, line 26: Shouldn't be here.

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,27 @@
+package core.basesyntax;
+
+public class Figure implements DrawFigure, AreaCalculator {
+
+    private String color;
```
nowickimj, line 5: Could be an `enum`.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,24 @@
...
+        Figure[] figures = new Figure[5];
+        figures[0] = new Circle(colorSupplier.getRandomColor(),Math.random() * 100);
+        figures[1] = new Rectangle(colorSupplier.getRandomColor(),Math.random() * 100,
+                Math.random() * 100);
+        figures[2] = new IsoscelesTrapezoid(colorSupplier.getRandomColor(),Math.random() * 100,
+                Math.random() * 100, Math.random() * 100);
+        figures[3] = new RightTriangle(colorSupplier.getRandomColor(), Math.random() * 100,
+                Math.random() * 100);
+        figures[4] = new Square(colorSupplier.getRandomColor(), Math.random() * 100);
+        int count = (int)(Math.random() * 4);
```
nowickimj, line 17: Notice that you create multiple object even though you return only one - please use a proper instructions to build only the requested figure. How about assigning each figure to a number 0-4 and use switch?

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,36 @@
...
+    private double radius;
+
+    public Circle(String color, double radius) {
+        super(color);
+        this.radius = radius;
+    }
+
+    public String getNameFigure() {
+        return nameFigure;
+    }
```
nowickimj, line 15: Notice that you added `figureName` to each figure - why not moving it to `Figure` then?
Besides that, it's not necessary so you can remove it as well and just write the name in `draw()` method.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+public class ColorSupplier {
+    private String[] colors = new String[]{"white","black","yellow","red","green","blue","gray"};
```
nowickimj, line 4: Enum usage would be perfect here.

nowickimj (changes requested):

src/main/java/core/basesyntax/AreaCalculator.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax;
+
+public interface AreaCalculator {
+
+    double area();
```
nowickimj, line 5: Method names should be a verb, for example getArea()

nowickimj (changes requested):
Few changes requested

PR #1547, opened 2024-04-03
Create Square`, `Rectangle`, `RightTriangle`, `Circle`, `IsoscelesTra…

SviatoslavTataryn:
…pezoid classes. abstract class Figure and two interface class ColorSupplier, FigureSupplier. All classes have state and method. In main() return array with 6 random Objects and random parameters.

### DIFF
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private static final Random random = new Random();
+    private static final String[] colors = {"Red", "Green", "Blue", "Yellow", "Pink"};
+
+    public String getRandomColor() {
+        return colors[random.nextInt(colors.length)];
+    }
+}
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -3,10 +3,10 @@ package core.basesyntax;
 import java.util.Random;
 
 public class ColorSupplier {
-    private static final Random random = new Random();
-    private static final String[] colors = {"Red", "Green", "Blue", "Yellow", "Pink"};
+    private final Random random = new Random();
+
 
     public String getRandomColor() {
-        return colors[random.nextInt(colors.length)];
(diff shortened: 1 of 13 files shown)


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private static final Random random = new Random();
```
Rommelua, line 6: It's not a constant
```suggestion
    private final Random random = new Random();
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
...
+    private static final String[] colors = {"Red", "Green", "Blue", "Yellow", "Pink"};
```
Rommelua, line 7: Lets create `enum Color` in a separate file

src/main/java/core/basesyntax/DrawFigure.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface DrawFigure {
```
Rommelua, line 3: Think about why we split abstract class and interface.
```suggestion
public interface Drawable {
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,37 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    public static final int FIGURE_COUNT = 5;
+    private static final Random random = new Random();
+    private static final ColorSupplier colorSupplier = new ColorSupplier();
+
+    private double generateRandomValue() {
```
Rommelua, line 10: Private ones (methods, classes) should be after public ones.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,37 @@
...
+        return (random.nextInt(10000) + 1) / 100.0;
```
Rommelua, line 11: image

Rommelua (changes requested):

PR #1559, opened 2024-04-11
jv-oop-advanced - first attempt

dedis34:


src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+public abstract class Figure implements AreaCalculator, ToDraw {
+    private String color;
```
nowickimj, line 4: Why not use enum to store the color?

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,17 @@
...
+
+    public Figure(Color color) {
+        this.color = String.valueOf(color);
```
nowickimj, line 7: 1. Keep the enum type as a field
2. Parse to string only if necessary - you can use `name()` method.

src/main/java/core/basesyntax/Circle.java
```diff
@@ -0,0 +1,31 @@
...
+                + "Circle"
+                + ","
+                + " area: "
+                + getArea()
+                + " sq. units,"
+                + " radius: "
+                + radius
+                + " units,"
+                + " color: "
+                + getColor());
```
nowickimj, line 28: To many new lines - at some point it makes code less readable. How about a line per property? you can use a line separator (`\n`) to print output in multiple lines:
```java
System.out.println("Figure: circle,\n"
  "area: " + getArea() + ",\n"
  ...
```

nowickimj: This comment applies to every `draw()` implementation.

src/main/java/core/basesyntax/ToDraw.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface ToDraw {
```
nowickimj, line 3: Interface names should be adjectives (most preferable) or nouns (as alternative).
How about `Drawable`?

nowickimj (changes requested):
Great job, few remarks but I see improvement :)

nowickimj (approved):

PR #1562, opened 2024-04-12
added solution

xotekijke:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import core.basesyntax.model.Color;
+import java.util.Random;
+
+public class ColorSupplier {
+    private static final Random random = new Random();
+
+    public static Color getRandomColor() {
```
kerrrusha, line 9: generally using statics is a bad practice
```suggestion
    private final Random random = new Random();

    public Color getRandomColor() {
```

src/main/java/core/basesyntax/Draw.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Draw {
```
kerrrusha, line 3: bad interface name
```suggestion
public interface Drawable {
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,49 @@
...
+    private final Random random = new Random();
+
+    public Figure getDefaultFigure() {
+        return new Circle(Color.WHITE, 10);
+    }
+
+    public Figure getRandomFigure() {
+        int randomIndex = random.nextInt(FIGURE_AMOUNT);
+        return new Figure[]{getRandomCircle(), getRandomRectangle(), getRandomIsoscelesTrapezoid(),
+                getRandomRightRectangle()}[randomIndex];
```
kerrrusha, line 23: better replace with switch-case block , it will make your code more optimized and faster

src/main/java/core/basesyntax/FindArea.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface FindArea {
```
kerrrusha, line 3: better name
```suggestion
public interface AreaCalculator {
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,23 @@
...
+            if (i == 1) {
+                figures[i] = figureSupplier.getDefaultFigure();
+            } else {
+                figures[i] = figureSupplier.getRandomFigure();
+            }
+        }
+
+        for (Figure figure : figures) {
+            figure.draw();
+        }
```
kerrrusha, line 21: try to use single loop to make code faster

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,23 @@
...
+import core.basesyntax.model.Figure;
+
+public class Main {
+    private static final int amountOfCreatedFigures = 5;
+
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] figures = new Figure[amountOfCreatedFigures];
+        for (int i = 0; i < amountOfCreatedFigures; i++) {
+            if (i == 1) {
```
kerrrusha, line 12: The first half of figures in this array should be generated with random parameters.

kerrrusha (changes requested):

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int FIGURE_AMOUNT = 4;
+    private static final int MAX_VALUE_FOR_RANDOM_FIGURE = 50;
+    private final Random random = new Random();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getDefaultFigure() {
+        return new Circle(Color.WHITE, 10);
```
Rommelua, line 18: image

src/main/java/core/basesyntax/Main.java
Rommelua: not fixed

Rommelua (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import core.basesyntax.model.Color;
+import java.util.Random;
+
+public class ColorSupplier {
+    private static final Random random = new Random();
```
kerrrusha, line 7: reference-type fields should not be static
```suggestion
    private final Random random = new Random();
```

src/main/java/core/basesyntax/FigureSupplier.java
kerrrusha, line 18: not fixed. 10 is magic number

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,53 @@
...
+                random.nextInt(MAX_VALUE_FOR_RANDOM_FIGURE));
+    }
+
+    private Figure getRandomRectangle() {
+        return new Rectangle(colorSupplier.getRandomColor(),
+                random.nextInt(MAX_VALUE_FOR_RANDOM_FIGURE),
+                random.nextInt(MAX_VALUE_FOR_RANDOM_FIGURE));
+    }
+
+    private Figure getRandomRightRectangle() {
```
kerrrusha, line 41:
```suggestion
    private Figure getRandomRightTriangle() {
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,53 @@
...
+    public Figure getDefaultFigure() {
+        return new Circle(Color.WHITE, 10);
+    }
+
+    public Figure getRandomFigure() {
+        return switch (random.nextInt(FIGURE_AMOUNT)) {
+            case 0 -> getRandomIsoscelesTrapezoid();
+            case 1 -> getRandomRectangle();
+            case 2 -> getRandomRightRectangle();
+            default -> getRandomCircle();
```
kerrrusha, line 26: how about creating Square?

kerrrusha (changes requested):
almost perfect

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import core.basesyntax.model.Color;
+import java.util.Random;
+
+public class ColorSupplier {
+    private final Random random = new Random();
+
+    public Color getRandomColor() {
+        return Color.values()[random.nextInt(Color.values().length)];
```
Elena-Bruyako, line 10: it's better to separate the logic: first find `int randomIndex ` and then other logic

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax;
+
+import core.basesyntax.model.Figure;
+
+public class Main {
+    private static final int amountOfCreatedFigures = 4;
```
Elena-Bruyako, line 6: it looks like a constant, so use the rule of naming constants. 
Probably FIGURES_COUNT will be better

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,13 @@
...
+import core.basesyntax.model.Color;
+import java.util.Random;
+
+public class ColorSupplier {
+    private final Random random = new Random();
+
+    public Color getRandomColor() {
+        int randomIndex = random.nextInt(Color.values().length);
+        return Color.values()[randomIndex];
+    }
```
MAMentorFX, line 12: return String please :)

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax;
+
+import core.basesyntax.model.Figure;
+
+public class Main {
+    private static final int FIGURES_COUNT = 4;
+
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] figures = new Figure[FIGURES_COUNT];
```
MAMentorFX, line 10: You dont really need Figure[] for this task

src/main/java/core/basesyntax/model/Figure.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax.model;
+
+import core.basesyntax.AreaCalculator;
+import core.basesyntax.Drawable;
+
+public abstract class Figure implements Drawable, AreaCalculator {
+    private final Color color;
```
MAMentorFX, line 7:
```suggestion
    private final String color;
```

src/main/java/core/basesyntax/model/Figure.java
```diff
@@ -0,0 +1,16 @@
...
+
+    public Figure(Color color) {
+        this.color = color;
+    }
+
+    public Color getColor() {
```
MAMentorFX, line 13:
```suggestion
    public String getColor() {
```

MAMentorFX (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
xotekijke: I don't see any sense of refactoring that. We are getting randomColor, which returns us Color, which is supposed to be enum as per task description. We can cast it to String if required or reuse in any other places in future if required

PR #1568, opened 2024-04-17
created 2 interfaces and abstract method "Figure" that implements the…

Zartones:
…m, created classes for 5 shapes that extend abstract class, enum with shape`s colors, created "ColorSupplier" to randomize and supply colors to shapes, "FigureSupplier" is similar but instead of colors it supplies random or default shapes, "Main" have an array of Figures and loop that prints result

### DIFF
--- a/src/main/java/core/basesyntax/Area.java
+++ b/src/main/java/core/basesyntax/Area.java
@@ -1,0 +1,5 @@
+package core.basesyntax;
+
+public interface Area {
+    double getArea();
+}
--- a/src/main/java/core/basesyntax/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/FigureSupplier.java
@@ -1,0 +1,35 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int BOUND = 100;
+    private static final int DEFAULT_RADIUS = 10;
+    private static final String DEFAULT_COLOR = "WHITE";
+    private Random random = new Random();
+    private ColorSupplier color = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        switch (random.nextInt(5)) {
+            case 0:
+                return new Circle(color.getRandomColor(), random.nextInt(BOUND) + 1);
+            case 1:
+                return new IsoscelesTrapezoid(color.getRandomColor(),
+                        random.nextInt(BOUND) + 1, random.nextInt(BOUND) + 1,
+                        random.nextInt(BOUND) + 1);
+            case 2:
+                return new Rectangle(color.getRandomColor(),
+                        random.nextInt(BOUND) + 1, random.nextInt(BOUND) + 1);
+            case 3:
+                return new RightTriangle(color.getRandomColor(),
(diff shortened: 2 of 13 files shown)


src/main/java/core/basesyntax/Area.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Area {
```
Elena-Bruyako, line 3:
```suggestion
public interface AreaCalculator {
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public class Main {
+    public static void main(String[] args) {
+        FigureSupplier figure = new FigureSupplier();
+        Figure[] array = {figure.getRandomFigure(), figure.getRandomFigure(),
```
Elena-Bruyako, line 6: It's hardcoding, think about how to resolve task description in other way: The first half of figures in this array should be generated with random parameters.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,35 @@
...
+
+public class FigureSupplier {
+    private static final int BOUND = 100;
+    private static final int DEFAULT_RADIUS = 10;
+    private static final String DEFAULT_COLOR = "WHITE";
+    private Random random = new Random();
+    private ColorSupplier color = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        switch (random.nextInt(5)) {
```
Elena-Bruyako, line 13: 5 is a magic number

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,35 @@
...
+            case 2:
+                return new Rectangle(color.getRandomColor(),
+                        random.nextInt(BOUND) + 1, random.nextInt(BOUND) + 1);
+            case 3:
+                return new RightTriangle(color.getRandomColor(),
+                        random.nextInt(BOUND) + 1, random.nextInt(BOUND) + 1);
+            default:
+                return new Square(color.getRandomColor(), random.nextInt(BOUND) + 1);
+        }
+
```
Elena-Bruyako, line 29:
```suggestion
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,35 @@
...
+    private static final int BOUND = 100;
+    private static final int DEFAULT_RADIUS = 10;
+    private static final String DEFAULT_COLOR = "WHITE";
+    private Random random = new Random();
+    private ColorSupplier color = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        switch (random.nextInt(5)) {
+            case 0:
+                return new Circle(color.getRandomColor(), random.nextInt(BOUND) + 1);
```
Elena-Bruyako, line 15: a lot of duplication, create local variable for color.getRandomColor() and method for calculating sides of shapes

Elena-Bruyako (changes requested):

kozhukhovsky (approved):

PR #1569, opened 2024-04-17
Create random figures generator

RomanChygryn:


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Feel free to remove this class and create your own.
+ */
+public class Main {
+    public static final int FIGURES_AMOUNT = 7;
+
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] figures = new Figure[FIGURES_AMOUNT];
+        figures[0] = figureSupplier.getRectangle();
```
Elena-Bruyako, line 12: This solution is hardcoding, think about how you can implement task description in other way: The first half of figures in this array should be generated with random parameters.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,60 @@
...
+            case 3:
+                return getRightTriangle();
+            case 4:
+                return getIsoscelesTrapezoid();
+            default:
+                return getDefaultFigure();
+        }
+    }
+
+    public Figure getCircle() {
```
Elena-Bruyako, line 30:
```suggestion
    private Figure getCircle() {
```

Check other mehtods

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/FigureSupplier.java
RomanChygryn: Changed

PR #1570, opened 2024-04-18
advanced oop (first draft)

alisahoys:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private static final String[] COLORS = {"beige", "magenta", "blue", "yellow", "pink", "purple"};
```
kerrrusha, line 6: better extract it to the enum
```suggestion
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
...
+    private static final Random RANDOM = new Random();
```
kerrrusha, line 7: reference-type variables should be fields, not constants
```suggestion
    private final Random random = new Random();
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
...
+
+    public static String getRandomColor() {
```
kerrrusha, line 9: generally using static is a bad practice
```suggestion
    public String getRandomColor() {
```

src/main/java/core/basesyntax/FigureProperties.java
kerrrusha, line 1: makes no sense in that case to extract getters into interfaces

src/main/java/core/basesyntax/FigureBehavior.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax;
+
+public interface FigureBehavior {
+    double getArea();
+
+    void draw();
```
kerrrusha, line 6: separate them into `AreaCalculator` and `Drawable` interfaces to implement Interface Segregation SOLID principle

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,32 @@
...
+                        RANDOM.nextInt(10) + 1, ColorSupplier.getRandomColor());
+            case 2:
+                return new RightTriangle(RANDOM.nextInt(10) + 1,
+                        RANDOM.nextInt(10) + 1, ColorSupplier.getRandomColor());
+            case 3:
+                return new Circle(RANDOM.nextInt(10) + 1, ColorSupplier.getRandomColor());
+            case 4:
+                return new IsoscelesTrapezoid(RANDOM.nextInt(10) + 1,
+                        RANDOM.nextInt(10) + 1,
+                        RANDOM.nextInt(10) + 1, ColorSupplier.getRandomColor());
```
kerrrusha, line 23: 5 and 10 is a magic number

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,32 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final Random RANDOM = new Random();
```
kerrrusha, line 6: same

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,21 @@
...
+        }
+
+        for (int i = arraySize / 2; i < arraySize; i++) {
+            figures[i] = FigureSupplier.getDefaultFigure();
+        }
+
+        // Display all figures
+        for (Figure figure : figures) {
+            figure.draw();
+        }
```
kerrrusha, line 19: think about how to combine them into single loop

kerrrusha (changes requested):

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,37 @@
...
+            case 2:
+                return new RightTriangle(random.nextDouble() * maxSize,
+                        random.nextDouble() * maxSize, color);
+            case 3:
+                return new Circle(random.nextDouble() * maxSize, color);
+            case 4:
+                return new IsoscelesTrapezoid(random.nextDouble() * maxSize,
+                        random.nextDouble() * maxSize, random.nextDouble() * maxSize, color);
+            default:
+                return null;
```
okuzan, line 30: remember that it's better to throw an exception that return null silently, nulls in java are a pain in the backside :)

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,37 @@
...
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(maxSize, Color.WHITE);
```
okuzan, line 35: extract to constants: DEFAULT_RADIUS, DEFAULT_COLOR

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,37 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private final int maxSize = 10;
```
okuzan, line 6: should be constant, it's a primitive :) add static, use constant case for naming

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,37 @@
...
+    private final Random random = new Random();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        Color color = colorSupplier.getRandomColor();
+        int numFigureTypes = 5;
```
okuzan, line 12: could be class-level constant too

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,37 @@
...
+                return new Square(random.nextDouble() * maxSize,color);
+            case 1:
+                return new Rectangle(random.nextDouble() * maxSize,
+                        random.nextDouble() * maxSize, color);
+            case 2:
+                return new RightTriangle(random.nextDouble() * maxSize,
+                        random.nextDouble() * maxSize, color);
+            case 3:
+                return new Circle(random.nextDouble() * maxSize, color);
+            case 4:
```
okuzan, line 26:
```suggestion
            case 4:
            default:
```
to avoid last case (null or throwing), you can do this

okuzan (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
...
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private final Color[] colors = Color.values();
+    private final Random random = new Random();
+
+    public Color getRandomColor() {
+        return colors[random.nextInt(colors.length)];
+    }
```
MentorMA, line 11: Return color in String type

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,44 @@
...
+
+public class FigureSupplier {
+    private static final Color DEFAULT_COLOR = Color.WHITE;
+    private static final int DEFAULT_RADIUS = 10;
+    private static final int MAX_SIZE = 10;
+    private final ColorSupplier colorSupplier;
+
+    {
+        colorSupplier = new ColorSupplier();
+    }
```
MentorMA, line 13: If you want to use init block, place it after fields but better init this in constructor

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,44 @@
...
+
+    private final Random random = new Random();
+
+    public Figure getRandomFigure() {
+        Color color = colorSupplier.getRandomColor();
+        int numFigureTypes = 5;
```
MentorMA, line 19: It should be constant

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,44 @@
...
+            case 0:
+                return new Square(random.nextDouble() * MAX_SIZE,color);
+            case 1:
+                return new Rectangle(random.nextDouble() * MAX_SIZE,
+                        random.nextDouble() * MAX_SIZE, color);
+            case 2:
+                return new RightTriangle(random.nextDouble() * MAX_SIZE,
+                        random.nextDouble() * MAX_SIZE, color);
+            case 3:
+                return new Circle(random.nextDouble() * MAX_SIZE, color);
```
MentorMA, line 32: Better create own methods for creating figures

MentorMA (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
alisahoys: why?

src/main/java/core/basesyntax/FigureSupplier.java
alisahoys: You mean: createRandomSquare(), createRandomRectangle(), createRandomRightTriangle(), createRandomCircle(), and createRandomIsoscelesTrapezoid()? That's not what is written in a README file. "FigureSupplier with the public Figure getRandomFigure() method - for generating figures with random properties."

okuzan: It doesn't contradict Readme, public method will return figures, relying on private methods internally. However, it's up to you whether to extract this logic to a private method. The rule of thumb is readability, which option is easier to grasp for someone who reads it for the first time. I think logic here is rather trivial, and by creating many more private methods we can even lose readability, but that's debatable

src/main/java/core/basesyntax/ColorSupplier.java
okuzan: Returning Color leaves more options to users of this class, if a consumer wants string representation, he can just call `.name()`, but we should force someone to parse String into a constant if there ever will be a need for it. Flexibility and 'future-proofness' are important considerations, so better leave as is

okuzan (approved):
LGTM 🤝

src/main/java/core/basesyntax/ColorSupplier.java
alisahoys: thank you very much for the explanation!

PR #1571, opened 2024-04-19
Tv1803

Tatian1801:
Creating a Drawable interface and a Shape class for drawing shapes


Tatian1801:
Creating a Drawable interface and a Shape class for drawing shapes. Look here please.

PR #1612, opened 2024-06-19
First solution

Rzarcik97:


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int LOWWER_BOUND = 1;
+    private static final int UPPER_BOUND = 101;
+    private Random number = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Circle getDefaultFigure() {
```
MateuszRostek, line 11:
```suggestion
    public Figure getDefaultFigure() {
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,34 @@
...
+        return new Circle(Color.WHITE, 10);
```
MateuszRostek, line 12: color and radius should also be constants:)

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,34 @@
...
+                                    number.nextInt(bounds)),
+                            new IsoscelesTrapezoid(colorSupplier.getRandomColor(),
+                                    number.nextInt(bounds),
+                                    number.nextInt(bounds),
+                                    number.nextInt(bounds)),
+                            new RightTriangle(colorSupplier.getRandomColor(),
+                                    number.nextInt(bounds),
+                                    number.nextInt(bounds))};
+        return figures[number.nextInt(figures.length)];
+    }
```
MateuszRostek, line 32: You are creating 5 figures, but you are using only one at the end, so these 4 are redundant each time.
Better to create a `switch` with cases here, and in each case create only one Figure and return it.

MateuszRostek (changes requested):
Really good job! There are few things that you should consider, according to my comments, and implement them.

MateuszRostek (approved):
Good job!

PR #1637, opened 2024-07-11
First solution of the task

B4rt1omiej:


src/main/java/core/basesyntax/IsoscelesTrapezoid.java
```diff
@@ -0,0 +1,45 @@
...
+public class IsoscelesTrapezoid extends Figure {
+    private int base1;
+    private int base2;
+    private int height;
+
+    public IsoscelesTrapezoid(int base1, int base2, int height, String color) {
+        this.base1 = base1;
+        this.base2 = base2;
+        this.height = height;
+        this.setColor(color);
```
beatarjn, line 12: If constructor that sets color is added in Figure class you could user instead of this.setColor(color) super(color).
super() refers to parent class object - in this case constructor from Figure that sets the color

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -0,0 +1,39 @@
+package core.basesyntax;
+
+public class Rectangle extends Figure {
+    private int side;
+    private int base;
+
+    public Rectangle(int side, int base, String color) {
+        this.side = side;
+        this.base = base;
+        this.setColor(color);
```
beatarjn, line 10: The same comment for color if Figure has a constructor

src/main/java/core/basesyntax/RightTriangle.java
```diff
@@ -0,0 +1,39 @@
+package core.basesyntax;
+
+public class RightTriangle extends Figure {
+    private int side;
+    private int base;
+
+    public RightTriangle(int side, int base, String color) {
+        this.side = side;
+        this.base = base;
+        this.setColor(color);
```
beatarjn, line 10: The same for color if the Figure has a constructor

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -5,41 +5,41 @@
...
-            "circle",
-            "isoscelestrapezoid"};
+    private final FigureType[] figureTypes = FigureType.values();
     private Random random = new Random();
 
     public Figure getRandomFigure() {
-        int randomFigureIndex = random.nextInt(figures.length);
-        String figure = figures[randomFigureIndex];
+        int randomFigureIndex = random.nextInt(figureTypes.length);
+        String figure = figureTypes[randomFigureIndex].toString();
```
beatarjn, line 13: Don't use toString() or String.valueOf()(it will call toString() under the hood) for getting the String representation of enum constants. toString() is common for all enum constants. It's better to use the standard method of enum name() that will be returning always String representation of the concrete enum constant.
You also don't need to have a String value here, you can operate on enums: figure will by of type enum and later in if statement you can use enum in comparison, e.g
if (FigureType.SQUARE.equals(figure))

beatarjn (changes requested):

beatarjn (approved):

PR #1657, opened 2024-08-02
Hw 5 is done, please check

Hero284:


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,20 @@
...
+            figures[i] = figureSupplier.getRandomFigure();
+        }
+
+        for (int i = figures.length / 2; i < figures.length; i++) {
+            figures[i] = figureSupplier.getDefaultFigure();
+        }
+
+        for (Figure figure : figures) {
+            System.out.println(figure);
+        }
```
okuzan, line 18: combine into one for loop

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,38 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private final Random random = new Random();
+    private final double defaultRadius = 10;
+    private final String defaultColor = "White";
```
okuzan, line 8: those should be constants, use constant case (uppercase with _ )
white - use Color.WHITE.name()

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,38 @@
...
+    private final double defaultRadius = 10;
+    private final String defaultColor = "White";
+
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Figure getRandomFigure() {
+        switch (random.nextInt(5)) {
+            case 0:
+                return new Circle(colorSupplier.getRandomColor(),
+                        random.nextInt(10));
```
okuzan, line 16: 10 - repeated magic number, should be a constant

okuzan (changes requested):

okuzan (approved):

PR #1665, opened 2024-08-19
Solution

vsupruniuk:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private final String[] colors = new String[]{"Red", "Yellow", "Blue"};
```
liliia-ponomarenko, line 6: Let's create an enum class for colors.

src/main/java/core/basesyntax/IFigure.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface IFigure {
```
liliia-ponomarenko, line 3: Let's give more informative interface name. For example Drawable

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,37 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private final Random random = new Random();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final int figuresCount = 5;
```
liliia-ponomarenko, line 8: Make this variable constant 
`private static final int FIGURE_COUNT = 5;`

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,37 @@
...
+            case 3:
+                return new Circle(color, this.random.nextDouble());
+            case 4:
+                return new IsoscelesTrapezoid(
+                        color, this.random.nextDouble(),
+                        this.random.nextDouble(),
+                        this.random.nextDouble()
+                );
+            default:
+                return null;
```
liliia-ponomarenko, line 30: Returning null from a method is a bad practice. If you use a switch case construction in your solution, you may just put the last possible option in the default case.

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,37 @@
...
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle("White", 10.0);
```
liliia-ponomarenko, line 35: use constant with informative name for these values (`White`  - should be enum value )

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,20 @@
...
+        FigureSupplier figureSupplier = new FigureSupplier();
+
+        Figure[] figures = new Figure[]{
+                figureSupplier.getRandomFigure(),
+                figureSupplier.getRandomFigure(),
+                figureSupplier.getRandomFigure(),
+                figureSupplier.getDefaultFigure(),
+                figureSupplier.getDefaultFigure(),
+                figureSupplier.getDefaultFigure(),
+        };
```
liliia-ponomarenko, line 14: Let's use a loop to create an array of Figure

liliia-ponomarenko (changes requested):
Good job! Let's improve your solution ;)

liliia-ponomarenko (approved):
Great 🔥

PR #1760, opened 2024-12-11
Implement tasks

euulu:


euulu:
> 3. **Enum String Representation**: Ensure that you use `name()` instead of `toString()` for enum constants to align with best practices.

Can you please explain in detailsthis point from your review. Where exactly should i use `name()` method instead of `toString()`?

PR #1771, opened 2024-12-21
created a program that generate the array with random/default figures

yaroslav-pryshchepa:
it also prints information about figures

### DIFF
--- a/src/main/java/core/basesyntax/Figure.java
+++ b/src/main/java/core/basesyntax/Figure.java
@@ -1,0 +1,17 @@
+package core.basesyntax;
+
+public abstract class Figure implements FigureSquare, FigureInformation {
+    private String color;
+
+    public String getColor() {
+        return color;
+    }
+
+    public void setColor(String color) {
+        this.color = color;
+    }
+
+    public abstract double getArea();
+
+    public abstract String printInformation();
+}
--- a/src/main/java/core/basesyntax/FigureSupplier.java
+++ b/src/main/java/core/basesyntax/FigureSupplier.java
@@ -1,0 +1,37 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int FIRST_MAX_NUMBER = 5;
+    private static final int SECOND_MAX_NUMBER = 100;
+    private static final int RADIUS = 10;
+    private final Random random;
+    private final ColorSupplier colorSupplier;
+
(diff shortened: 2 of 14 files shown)


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -0,0 +1,23 @@
...
+        FigureSupplier figureSupplier = new FigureSupplier(random, colorSupplier);
+        Figure[] figures = new Figure[6];
+
+        for (int i = 0; i < figures.length; i++) {
+            if (i < figures.length / 2) {
+                figures[i] = figureSupplier.getRandomFigure();
+                System.out.println(figures[i].printInformation());
+            } else {
+                figures[i] = figureSupplier.getDefaultFigure();
+                System.out.println(figures[i].printInformation());
```
okuzan, line 19: extract common part from if-else

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,37 @@
...
+    public Figure getRandomFigure() {
+        int max1 = random.nextInt(SECOND_MAX_NUMBER) + 1;
+        int max2 = random.nextInt(SECOND_MAX_NUMBER) + 1;
+        Figure figure = switch (random.nextInt(FIRST_MAX_NUMBER) + 1) {
+            case 1 -> new Rectangle(max1, max2);
+            case 2 -> new RightTriangle(max1, max2);
+            case 3 -> new Ellipse(max1, max2);
+            case 4 -> new Square(max1);
+            case 5 -> new IsoscelesTrapezoid(max1, max2, max2);
+            default -> throw new IllegalArgumentException("Unexpected value");
```
okuzan, line 26: IllegalState is more applicable here, cuz this method doesn't have arguments, problem can only occur via developer's mistake, misconfiguration of lengths. You can also make case 5 default

src/main/java/core/basesyntax/Figure.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+public abstract class Figure implements FigureSquare, FigureInformation {
+    private String color;
+
```
okuzan, line 5: consider creating a constructor for color and calling super() in subclasses

okuzan (changes requested):
Good job overall!

sarakhmen (approved):
lgtm!

PR #1790, opened 2025-01-17
first solution

anpiet:


src/main/java/core/basesyntax/figures/ColorSupplier.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax.figures;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    public String getRandomColor() {
+        Random random = new Random();
```
slade13, line 7: The Random object is instantiated every time the getRandomColor method is called. Consider making it a class-level field to avoid unnecessary object creation.

src/main/java/core/basesyntax/figures/Square.java
```diff
@@ -0,0 +1,24 @@
...
+
+    @Override
+    public double getArea() {
+        return (double) Math.round(side * side * 100) / 100;
+    }
+
+    @Override
+    public void draw() {
+
+        System.out.println("Figure: square, area: " + getArea()
```
slade13, line 19: The `draw` method should not have an empty line at the beginning. Remove the empty line for better readability and consistency.

src/main/java/core/basesyntax/figures/Circle.java
```diff
@@ -0,0 +1,22 @@
...
+    private final double radius;
+
+    public Circle(double radius, String color) {
+        super(color);
+        this.radius = radius;
+    }
+
+    @Override
+    public double getArea() {
+        return (double) Math.round(Math.PI * radius * 100) / 100;
```
slade13, line 13: The calculation of the area is incorrect. The formula should be Math.PI * radius * radius to calculate the area of a circle. Additionally, rounding should be done after the entire calculation, not just multiplying by 100.

src/main/java/core/basesyntax/Application.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import core.basesyntax.figures.FigureSupplier;
+
+public class Application {
+    public static void main(String[] args) {
+        for (int i = 0; i < 6; i++) {
+            if (i < 3) {
+                new FigureSupplier().getRandomFigure().draw();
```
slade13, line 9: Creating a new instance of FigureSupplier in each iteration is inefficient. Consider creating a single instance of FigureSupplier outside the loop and reusing it.

slade13 (approved):
Good job, please review my comments.

PR #1802, opened 2025-02-05
Jv oop advanced solution

stanko-dev:


stanko-dev (commented):
constant number of figures added

PR #1807, opened 2025-02-07
Having obtained an array of figures. What is formed from the figures …

GennadyX7:
…of randomly generated and default figures.

### DIFF
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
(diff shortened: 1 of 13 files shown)


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    public String getRandomColor() {
+        int index = new Random().nextInt(Color.values().length);
```
Elena-Bruyako, line 7: Random should be class-level variable

Elena-Bruyako (changes requested):

Elena-Bruyako:
please, fix build

PR #1883, opened 2025-05-15
Update ci.yml

liudmylabiletska:


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,54 @@
...
+
+import java.util.Random;
+
+public class FigureSupplier {
+    private static final int DEFAULT_RADIUS = 10;
+    private static final int FIGURE_COUNT = 5;
+    private final Random random = new Random();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    private int generateRandomSize() {
```
Elena-Bruyako: private method should be placed after all public methods

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,20 @@
...
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] figures = new Figure[FIGURE_COUNT];
+
+        for (int i = 0; i < figures.length; i++) {
+            figures[i] = (i < figures.length / 2)
+                    ? figureSupplier.getRandomFigure()
+                    : figureSupplier.getDefaultFigure();
+        }
+
+        for (Figure figure : figures) {
```
Elena-Bruyako: you can use one for loop

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #1923, opened 2025-07-15
feat: implement Figure hierarchy and suppliers

anurmetov:
- Create abstract Figure class with getArea() and draw() methods
- Implement concrete figures: Circle, Square, Rectangle, etc.
- Create FigureSupplier for random/default figure generation
- Add ColorSupplier that generates random Color from enum
- Add Main class to test and print generated figures

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -8,14 +8,11 @@ public class Application {
         for (int i = 0; i < figures.length; i++) {
             if (i < figures.length / 2) {
                 figures[i] = figureSupplier.getRandomFigure();
+                figures[i].draw();
             } else {
                 figures[i] = figureSupplier.getDefaultFigure();
+                figures[i].draw();
             }
-        }
-
-        for (Drawable figure : figures) {
-            figure.draw();
-            System.out.println("\n");
         }
     }
 
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -1,11 +1,12 @@
 package core.basesyntax;
 
 public class Application {
-    private final static int NUMBER_OF_FIGURES = 6;
+    static final int FIGURES_NUMBER = 6;
 
     public static void main(String[] args) {
+
         FigureSupplier figureSupplier = new FigureSupplier();
-        Figure[] figures = new Figure[NUMBER_OF_FIGURES];
+        Figure[] figures = new Figure[FIGURES_NUMBER];
 
         for (int i = 0; i < figures.length; i++) {
             if (i < figures.length / 2) {
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -13,11 +13,10 @@ public class Application {
             }
         }
 
-        for (Figure figure : figures) {
+        for (Drawable figure : figures) {
             figure.draw();
             System.out.println("\n");
         }
-
     }
 
 }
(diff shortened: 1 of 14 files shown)


anurmetov:
@mateacademy-ai-mentor 
I have reviewed your comments, and i would like to clarify a few points that were misunderstood:

- Due to excercise it was correct to add abstract Figure class which implemets Drawable and Calculatable interfaces which methods are overwritten in each class, cause all the figures have the same state (Figure abstract class) and behaviour (Interfaces Drawable and Calculatable)

- Then, when you need to generate a random figure in the FigureSupplier class, which is not possible if we are not extending Figure class, because each figure is a separate object.

src/main/java/core/basesyntax/Application.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax;
+
+public class Application {
+    public static void main(String[] args) {
+        FigureSupplier figureSupplier = new FigureSupplier();
+        Figure[] figures = new Figure[6];
```
Elena-Bruyako: 6 is a magic number, create constant

src/main/java/core/basesyntax/Application.java
```diff
@@ -0,0 +1,20 @@
...
+
+        for (int i = 0; i < figures.length; i++) {
+            if (i < figures.length / 2) {
+                figures[i] = figureSupplier.getRandomFigure();
+                figures[i].draw();
+            } else {
+                figures[i] = figureSupplier.getDefaultFigure();
+                figures[i].draw();
+            }
```
Elena-Bruyako:
```suggestion
            if (i < figures.length / 2) {
                figures[i] = figureSupplier.getRandomFigure();
            } else {
                figures[i] = figureSupplier.getDefaultFigure();
            }
            figures[i].draw();
```

src/main/java/core/basesyntax/Calculatable.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public interface Calculatable {
```
Elena-Bruyako:
```suggestion
public interface AreaCalculator {
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private Random random = new Random();
```
Elena-Bruyako:
```suggestion
    private final Random random = new Random();
```

src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,47 @@
...
+                        random.nextInt(1, MAX_LENGTH_BASE), random.nextInt(1, MAX_LENGTH_HEIGHT));
+            }
+            default -> {
+                return null;
+            }
+        }
+    }
+
+    public Figure getDefaultFigure() {
+        return new Circle(Color.WHITE.name(), 10);
```
Elena-Bruyako: 10 is a magic number, create constant

src/main/java/core/basesyntax/IsoscelesTrapezoid.java
```diff
@@ -0,0 +1,28 @@
+package core.basesyntax;
+
+public class IsoscelesTrapezoid extends Figure {
+    private double base1;
```
Elena-Bruyako: don't use numbers for naming

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #1970, opened 2025-09-16
Feature/task completion

DianaKiyan:


src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public enum Colors {
```
Elena-Bruyako:
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/IsoscelesTrapezoid.java
```diff
@@ -0,0 +1,28 @@
+package core.basesyntax;
+
+public class IsoscelesTrapezoid extends Figure {
+    private final double base1;
```
Elena-Bruyako: don't use numbers for naming
same comment for other places

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,21 @@
+package core.basesyntax;
+
+public class Main {
+    private static final int FIGURE_COUNT = 6;
+
+    public static void main(String[] args) {
+        Figure[] figures = new Figure[FIGURE_COUNT];
+        FigureSupplier figureSupplier = new FigureSupplier();
+
+        for (int i = 0; i < FIGURE_COUNT / 2; i++) {
```
Elena-Bruyako: use one for loop to implement main method

Elena-Bruyako (changes requested):
see comments

src/main/java/core/basesyntax/Colors.java
Elena-Bruyako: enum name should be singular according to the naming convention

src/main/java/core/basesyntax/Main.java
Elena-Bruyako: fix comment

Elena-Bruyako (changes requested):
fix previous comments

Elena-Bruyako (approved):

PR #2018, opened 2025-12-15
solve Jv-oop-advanced task

b1gsal:


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+public class Main {
+    public static void main(String[] args) {
+        Figure[] figures = new Figure[6];
```
Elena-Bruyako: create constant for magic number 6

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,17 @@
...
+        FigureSupplier figureSupplier = new FigureSupplier();
+        for (int i = 0; i < 3; i++) {
```
Elena-Bruyako: use only one for loop to implement main method

Elena-Bruyako (changes requested):
Good job! See comments

sarakhmen (approved):
lgtm!

PR #2049, opened 2026-02-09
Solution implementation

tylerj231:


src/main/java/core/basesyntax/RightTriangle.java
```diff
@@ -0,0 +1,44 @@
...
+        this.firstLeg = firstLeg;
+        this.secondLeg = secondLeg;
+    }
+
+    @Override
+    public double getArea() {
+        return (double) firstLeg * secondLeg / 2;
+    }
+
+    public static RightTriangle getRandomRightRectangle() {
```
fetis-off, line 20: Static methods are in general a bad practice. Let's better create an instance of a class which method you want to call.

src/main/java/core/basesyntax/Square.java
```diff
@@ -0,0 +1,46 @@
...
+        return (double) width * height;
+    }
+
+    public static Square getRandomSquare() {
+        Random random = new Random();
+        Colors [] colors = Colors.values();
+        int randomColorIndex = random.nextInt(colors.length);
+        String randomColor = colors[randomColorIndex].name();
+        int randomWidth = random.nextInt(VALUES_BOUND);
+        int randomHeight = random.nextInt(VALUES_BOUND);
```
fetis-off, line 26: As I know from math square has equals all 4 sides :)

fetis-off (changes requested):
Good one! See comments

PR #2080, opened 2026-03-23
1 commit

Daniul2:


Elena-Bruyako (changes requested):
There are no FigureSupplier and Main class, please provide them

